PR #194, opened 2022-12-01
[WE-2566] Fix Multi-Plane Issues

TaylorBeeston:
# Overview

#### 🎟 Relevant Jira Issues

[WE-2566] Fix Issues with Multiple Planes

There are a couple of issues currently when using a wallet with multiple plugins that implement read, store, index, or cache

#### 🥴 TL; RL:
Fixes the above issues

- Fixes a race condition in conglomerate `get` methods
    - This was caused by using `Promise.any` on all valid plugins, assuming that all valid plugins would return _something_.
What can _actually_ happen is that one plugin can return `undefined` and cause the plugin that _can_ return something to get skipped.
This effectively means that what is returned is whatever runs _fastest_, and _not_ whatever actually resolves to something.
This has been changed to use `Promise.allSettled` and then a loop through the results to find the first plugin
that found _something_.
- Fixes an issue with deduping on the conglomerate `index.get` method
    - I was accidentally deduping based on object reference, rather than by ids 🤦
- Adds tests/test plugins to catch above regressions

None! These are just quick fixes

#### 🔍 Types of Changes

- [x] No
- [ ] Yes

# Testing

`pnpm test`

You can also pop open the CLI and add the test plugins manually, following the Control Planes tests.
When the SQLite Plugin PR is up, that would be a solid test of this as well, but I suppose that can't
really go up until this is done 🤔

#### 🧪 Code Coverage

Added tests to catch these guys

# Documentation

#### 📊 Storybook

# ✅ PR Checklist

### DIFF
--- a/.changeset/eight-keys-hammer.md
+++ b/.changeset/eight-keys-hammer.md
@@ -1,0 +1,9 @@
+---
+'@learncard/core': minor
+---
+
+Fix Multi-Plane issues:
+
+-   Fix race condition in `get`
+-   Dedupe `index.all` results by `id` rather than by object reference
+-   Fix race condition in cache `get` methods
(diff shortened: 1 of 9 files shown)


goblincore (approved):
Awesome LGTM 👏

PR #301, opened 2023-05-25
Return all Boost Recipients

TaylorBeeston:
- :bug: Loosen BoostRecipientInfo type
- :bug: Return boost recipients who haven't accepted yet

# Overview

Right now, when calling `getBoostRecipients`, only recipients who have accepted the boost are returned.

#### 🥴 TL; RL:
Adds an `includeUnacceptedBoosts` flag (defaults to true) that allows asking for all recipients, not
just those who have accepted

- `BoostRecipientInfo` type has been loosened to allow the received date to be optional (in case it hasn't been received yet!)
- `getBoostRecipients` now takes in an `includeUnacceptedBoosts` flag

The Neo4j query was kind of tough to write, because the `BoostRecipientInfo` type wants a full user 
Profile and not just the ID (which is all you have if they haven't accepted the boost!), so, it actually
does two queries: One to get all the recipient/sender IDs (and received date), and another to go and
grab the full recipient profile for all recipients. I _think_ this could be optimized into one query
using the Neo4j `UNION` operator, but I'm not 100% sure, and it would heavily complicate this call,
so I left it like this for now.

#### 🔍 Types of Changes

- [x] No
- [ ] Yes

# Testing

Run the tests! Also, pop open the CLI, make a boost, send it to someone (but don't accept it), and 
call `getBoostRecipients` and make sure they show up!

#### 🧪 Code Coverage

I wrote tests to make sure both cases of the flag are covered!

# Documentation

#### 📊 Storybook

# ✅ PR Checklist

### DIFF
--- a/packages/learn-card-types/src/lcn.ts
+++ b/packages/learn-card-types/src/lcn.ts
@@ -43,7 +43,7 @@ export type Boost = z.infer<typeof BoostValidator>;
 export const BoostRecipientValidator = z.object({
     to: LCNProfileValidator,
     from: z.string(),
-    received: z.string(),
+    received: z.string().optional(),
 });
 
 export type BoostRecipientInfo = z.infer<typeof BoostRecipientValidator>;
--- a/packages/plugins/learn-card-network/src/plugin.ts
+++ b/packages/plugins/learn-card-network/src/plugin.ts
@@ -336,10 +336,21 @@ export const getLearnCardNetworkPlugin = async (
 
                 return client.boost.getBoosts.query();
             },
-            getBoostRecipients: async (_learnCard, uri, limit = 25, skip = undefined) => {
+            getBoostRecipients: async (
+                _learnCard,
+                uri,
+                limit = 25,
+                skip = undefined,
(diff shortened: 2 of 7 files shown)


goblincore:
Screenshot 2023-05-25 at 11 30 53 AM
@TaylorBeeston Probably some error on my part, but I was trying to test this in the CLI but I keep getting this error:
Screenshot 2023-05-25 at 11 30 53 AM

For context, I checked this branch out, I ran pnpm exec nx build cli, then in the cli folder i ran pnpm start. I had just checked out the OBJ update PR previously (where I ran the same commands and that worked fine).

TaylorBeeston:
@goblincore hmm.. Looks more like an issue on my part! That error happens if there's a mismatch between the wasm binary itself and the wrapping code around it. Usually it happens when I forget to update the hosted binary and something uses that. But in this case I think the CLI is explicitly pointed at the wasm somewhere, and that's likely out of date!

TaylorBeeston:
Ah wait hold on I mixed up which PR this comment is on! This is likely from nx caching the didkit plugin!

PR #351, opened 2023-08-03
Fix Didkit Concurrency Bug

TaylorBeeston:
# Overview

Right now, if you instantiate two LearnCards at the same time via Promise.all, it will not work and
at least one of them will totally fail!

#### 🥴 TL; RL:
Say goodbye to this error! 
image

Prior to this change, I was preventing costly re-inits of the webassembly by keeping a variable around
to check if the init function had already been ran. This results in ~9x speedup when instantiating multiple
LearnCards. However, this check was pretty crude, and if a second LearnCard wanted to come in and initialize
while the first one was currently initializing, it would just _not_ init, and then didkit would have no access
to wasm. This PR beats this by adding a more robust `generating` variable. Using this, if a second LearnCard
comes in wanting to instantiate while the init function is running from another LC, it will now poll
every 250ms until that `generating` variable becomes false, at which point it will make sure init
ran successfully, then return as expected.

I just kind of arbitrarily picked 250ms as a poll time, because I think it is sane. I didn't do any kind
of benchmarking or anything, but I think it's fine

#### 🔍 Types of Changes

- [x] No
- [ ] Yes

# Testing

In the CLI code (`packages/learn-card/cli/src/index.ts`), add a `await Promise.all([initLearnCard({ seed: 'a' }), initLearnCard({ seed: 'b' })])`.
On `main`, that will error, on this branch it will not!

#### 🧪 Code Coverage

# Documentation

#### 📊 Storybook

# ✅ PR Checklist
- [ ] I have **commented** my code, particularly where ambiguous
- [x] I have made corresponding changes to **gitbook documentation**

### DIFF
--- a/.changeset/soft-oranges-report.md
+++ b/.changeset/soft-oranges-report.md
@@ -1,0 +1,5 @@
+---
+'@learncard/didkit-plugin': minor
+---
+
+Fix didkit failing to properly init when instantiating multiple LearnCards concurrently
--- a/packages/plugins/didkit/src/didkit/index.ts
+++ b/packages/plugins/didkit/src/didkit/index.ts
@@ -3,16 +3,22 @@ import _init, { InitInput } from './pkg/didkit_wasm';
 export * from './pkg/didkit_wasm';
 
 let initialized = false;
+let generating = false;
 
 export const init = async (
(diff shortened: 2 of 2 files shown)


Custard7 (approved):
Tested and worked for me - code looks good! This will get rid of lots of pesky errors in Sentry as well 💪 🎉

PR #359, opened 2023-08-07
[WE-2902] LearnCloud Index Plane Caching

TaylorBeeston:
- :white_check_mark: Flush cache in user/index plane tests
- :label: Update return type for cache.keys
- :label: Add doc comment for user cache ttl
- :arrow_up: Add json-stringify-deterministic dep
- :zap: Add caching to users via access layer methods
- :zap: Cache index plane
- :bug: Allow adding a did if its user is only itself

# Overview

#### 🎟 Relevant Jira Issues

[WE-2902] Add Caching for LearnCloud Index

Right now, the `index.get` call can sometimes take a _very_ long time (~3-4s)! We can mitigate this
by adding Redis caching in front of it, so long as we generously flush the cache when performing write
operations

#### 🥴 TL; RL:
Adds caching to index plane and users

- Cache the `index.get` method
- Flush the cache for all of a users dids on any of the other `index` methods, as well as when changing the dids of a user
- Add caching in front of users to allow the flush method to be more performant
- Flush cache in all tests to ensure that they work correctly
- Allow adding a did if it is the only did for a user
    - This is because the new caching will now force a user to be created when adding to the index.
      Before, you could add to your index without technically becoming a user, meaning your did could 
      still be added to a different user after doing that! This is intentional behavior, and to preserve
      that, I am allowing users with only one did to be added as if they weren't a user

There are extra hits to redis to get all the dids for a user. This could be defeated with request 
batching a la dataloaders, but I didn't want to go crazy and throw that in, so I just decided to 
cache users in Redis as well to keep Mongo hits down.

#### 🔍 Types of Changes

- [x] No
- [ ] Yes

# Testing

Run the test suite! It will use node ioredis and test it all out!

#### 🧪 Code Coverage

# Documentation

#### 📊 Storybook

# ✅ PR Checklist

### DIFF
--- a/services/learn-card-network/learn-cloud-service/src/routes/user.ts
+++ b/services/learn-card-network/learn-cloud-service/src/routes/user.ts
@@ -9,6 +9,7 @@ import { getEmptyLearnCard } from '@helpers/learnCard.helpers';
 import { DidAuthVP } from 'types/vp';
 import { addDidToUser, removeDidFromUser, setDidAsPrimary } from '@accesslayer/user/update';
 import { ensureUserForDid } from '@accesslayer/user/create';
+import { deleteUserByDid } from '@accesslayer/user/delete';
 
 export const userRouter = t.router({
     getDids: didAndChallengeRoute
@@ -62,10 +63,14 @@ export const userRouter = t.router({
                 const existing = await getUserForDid(did);
 
                 if (existing) {
-                    throw new TRPCError({
-                        code: 'CONFLICT',
-                        message: 'Did is already associated with a user',
-                    });
+                    if (existing.did !== did || existing.associatedDids.length > 0) {
+                        throw new TRPCError({
+                            code: 'CONFLICT',
+                            message: 'Did is already associated with a user',
+                        });
+                    }
+
+                    await deleteUserByDid(did);
                 }
 
                 const user = await ensureUserForDid(ctx.user.did);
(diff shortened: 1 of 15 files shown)


services/learn-card-network/learn-cloud-service/src/routes/user.ts
```diff
@@ -62,10 +63,14 @@ export const userRouter = t.router({
...
-                    });
+                    if (existing.did !== did || existing.associatedDids.length > 0) {
+                        throw new TRPCError({
+                            code: 'CONFLICT',
+                            message: 'Did is already associated with a user',
+                        });
+                    }
+
+                    await deleteUserByDid(did);
                 }
```
Custard7, line 74: I'm slightly confused by what's going on here? Is it supposed to be existing.did !== did?

Custard7 (approved):
This looks good to me! Code is good, and tests pass locally ✅ 

I get one TS build error that looks like it could be fixed easily, not sure if you are getting locally:

Screen Shot 2023-08-07 at 5 22 33 PM

services/learn-card-network/learn-cloud-service/src/routes/user.ts
TaylorBeeston: So this guy is pretty confusing, so I put a lengthy comment there helping to explain what's going on.

Basically, if you add some stuff to your cache using did:key (i.e. before you sign up for LCN), then switch over to did:web and want to associate those two accounts, that's what this method is for. However, because of the caching stuff, we are now actually making a user for that first did:key, and this would cause this route to error. This added check simply allows deleting that first simple user which was probably made by accident!

Custard7: Awesome! Yes, thanks for adding an explanatory comment 😎

Custard7 (approved):

PR #360, opened 2023-08-08
[WE-2895] SQS for Notifications

TaylorBeeston:
# Overview

#### 🎟 Relevant Jira Issues

[WE-2895] Add SQS Queue and Don't await Notifications

We are wasting a lot of time when sending notifications waiting for the whole notification request to
finish. We can't not await it though because if the lambda ends, the request might not go through

#### 🥴 TL; RL:
Adds an SQS queue that we can use to send those notifications instead

- Uses serverless lift to create a notifications Queue
- Adds a handler for the above queue that just calls `sendNotification`
- Created a drop-in replacement function for `sendNotification` called `addNotificationToQueue` that instead sends a request to SQS
- Turned off sending notifications/using the queue in test and local environments because SQS doesn't work there and is a bit involved to set up

Right now there's no way to test the queue, because we'd have to get something like ElasticMQ running
in a docker container next to the test suite to do so. This is fine locally, but I think in CI that
would probably be _super_ annoying

#### 🔍 Types of Changes

- [x] No
- [ ] Yes

# Testing

So what I did is to just like stub out the notification handler with logs, then push that up to a lambda
on my personal AWS account, hit it, and make sure that things were going through. If you wanted to
test this yourself, you could probably just do the same thing with like the dev stage for LCN or whatever.
So basically comment out all the domain/ssl stuff in `serverless.yml`, change the `sendNotification`
function to just have like console logs so you know it's working, then add an `await addNotificationToQueue({})`
in the health-check route, deploy the lambda, then go to `api/health-check` and check cloudwatch to
see if it worked!

#### 🧪 Code Coverage

I couldn't write any tests for the reasons mentioned in tradeoffs =(

# Documentation

#### 📊 Storybook

# ✅ PR Checklist


Custard7 (approved):
@TaylorBeeston - I haven't done the full test locally like you suggested with a personal lambda, but the code matches the documentation in the serverless-lift docs ✅ 

Additionally, tests are passing locally & running serverless locally is working. 

We should definitely test and monitor post-deploy 👍

PR #388, opened 2023-11-19
[LC-113] Allow Adding Additional Boost Admins

TaylorBeeston:
- :hammer: Emit declarationMaps
- :sparkles: Add basic CRUD for Boost Admins
- :sparkles: Update permissions to allow admins to actually do things!
- :sparkles: Return boosts you are admin of in getBoosts
- :bookmark: Changeset
- :sparkles: Add new methods to plugin

# Overview

#### 🎟 Relevant Jira Issues

[LC-113] Assign additional admins to Boosts

Right now, only the creator of a boost can send or edit it. That is sad =(

#### 🥴 TL; RL:
Allows adding additional Boost admins who may then send and edit boosts!

- I added `declarationMap: true` to all tsconfigs to allow jumping straight to ts files in your editor!
- Add three new routes: `getBoostAdmins`, `addBoostAdmin`, `removeBoostAdmin`, they just do what they say on the tin
- There is a new relationship between profiles and Boosts called `ADMIN_OF`
- All relevant boost queries have been updated to check for both `CREATED_BY` and `ADMIN_OF` relationships
- Admins are allowed to add more admins
- Admins are allowed to remove themselves as admins
- The Boost creator is _not_ allowed to be removed as an admin
- Anyone is allowed to view a Boost's admins list

I'm not 100% sure that we want to allow everyone to view a Boost's admin list, but it seems right to me.
Also, to address potential Boost orphans and to limit malicious behavior, I have set it so that the
Boost creator is kind of like a super admin in that they cannot be removed as an admin. In the future,
we may want to flesh this out a bit better and instead be able to allow more fine-grained permissions
to boost admins so that potentially untrusted admins can be allowed to send a boost but _not_ be able
to just remove other admins at will.

#### 🔍 Types of Changes

- [x] No
- [ ] Yes

# Testing

Run the test suite!

#### 🧪 Code Coverage

I wrote lots of new tests to cover all the new behavior =)

# Documentation

#### 📊 Storybook

# ✅ PR Checklist

### DIFF
--- a/services/learn-card-network/brain-service/src/routes/boosts.ts
+++ b/services/learn-card-network/brain-service/src/routes/boosts.ts
@@ -315,7 +315,7 @@ export const boostsRouter = t.router({
             })
         )
         .output(PaginatedLCNProfilesValidator)
-        .mutation(async ({ input, ctx }) => {
+        .query(async ({ input, ctx }) => {
             const { uri, limit, cursor, includeSelf } = input;
 
             const boost = await getBoostByUri(uri);
--- a/services/learn-card-network/brain-service/src/routes/boosts.ts
+++ b/services/learn-card-network/brain-service/src/routes/boosts.ts
@@ -86,7 +86,7 @@ export const boostsRouter = t.router({
 
             if (!boost) throw new TRPCError({ code: 'NOT_FOUND', message: 'Could not find boost' });
 
-            if (!(await isProfileBoostOwner(profile, boost))) {
+            if (!(await isProfileBoostAdmin(profile, boost))) {
                 throw new TRPCError({
                     code: 'UNAUTHORIZED',
                     message: 'Profile does not own boost',
@@ -163,7 +163,7 @@ export const boostsRouter = t.router({
 
             if (!boost) throw new TRPCError({ code: 'NOT_FOUND', message: 'Could not find boost' });
 
-            if (!(await isProfileBoostOwner(profile, boost))) {
+            if (!(await isProfileBoostAdmin(profile, boost))) {
                 throw new TRPCError({
                     code: 'UNAUTHORIZED',
                     message: 'Profile does not own boost',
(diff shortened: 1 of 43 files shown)


services/learn-card-network/brain-service/test/boosts.spec.ts
```diff
@@ -546,6 +623,340 @@ describe('Boosts', () => {
...
+            ).resolves.not.toThrow();
+
+            expect(
+                (await userA.clients.fullAuth.boost.getBoostAdmins({ uri })).records
+            ).toHaveLength(0);
+
+            await userA.clients.fullAuth.boost.addBoostAdmin({ uri, profileId: 'userb' });
+
+            expect(
+                (await userA.clients.fullAuth.boost.getBoostAdmins({ uri })).records
```
Custard7, line 722: Is this 1 because of 'include self' param? So the count is just for userb?

services/learn-card-network/brain-service/src/routes/boosts.ts
```diff
@@ -287,6 +296,166 @@ export const boostsRouter = t.router({
...
+                path: '/boost/admins/{uri}',
+                tags: ['Boosts'],
+                summary: 'Get boost admins',
+                description: 'This route returns the admins for a boost',
+            },
+        })
+        .input(
+            PaginationOptionsValidator.extend({
+                limit: PaginationOptionsValidator.shape.limit.default(25),
+                includeSelf: z.boolean().default(false),
```
Custard7, line 313: Does it make sense for this to be default false? As in, would it be confusing to not get yourself if you are specifically querying for boost admins? 

But maybe it does - thoughts?

Custard7 (approved):
This is great! Test suite looks great and is passing ✅ 

Also yes - I agree with you that admins should _not_ be able to remove the creator of the boost for now. 👍

services/learn-card-network/brain-service/test/boosts.spec.ts
TaylorBeeston: Yup! I wasn't sure if `includeSelf` makes more sense as true or false as a default, so I decided to go with excluding self by default

services/learn-card-network/brain-service/src/routes/boosts.ts
TaylorBeeston: I think when I first thought about this I'd thought it would make sense to default to false, but actually now that you're bringing it up, I think it would make more sense to be true by default. I'll go ahead and update!

PR #399, opened 2023-12-04
[LC-133] Fix Connections Caching Bug

TaylorBeeston:
# Overview

#### 🎟 Relevant Jira Issues

[LC-133] Connections are incorrectly cached

If a user updates their profile and someone who was connected to them had already warmed up their
connections cache, then the cache would not return the updated profile if the other profile requested
their connections again.

#### 🥴 TL; RL:
Fixes the above

Right now it's just kind of a simple fix where, when updating a profile, we grab all connections and
then evict their connections cache. In the future, when this all becomes more complicated, we'll want
to move that into the Queue so that we don't slow down the initial update request doing cache
management

Current solution works for low numbers of connections, but will not scale! This isn't a huge issue
right now though as connections are still capped at 50 users anyways. When pagination comes into play
though, we'll want to move the cache management logic to a queue and paginate our way through the
user's connections, ideally with a larger page size and only requesting the profileIds of the connections,
whilst quickly giving the update request a response.

#### 🔍 Types of Changes

- [x] No
- [ ] Yes

# Testing

Run the test suite!

#### 🧪 Code Coverage

I added a test that covers this exact case =)

# Documentation

#### 📊 Storybook

# ✅ PR Checklist

### DIFF
--- a/.changeset/fair-peaches-divide.md
+++ b/.changeset/fair-peaches-divide.md
@@ -1,0 +1,5 @@
+---
+'@learncard/network-brain-service': patch
+---
+
+Fix cache error with connections


goblincore (approved):
LGTM

PR #463, opened 2024-06-21
TS 5.5 Update and Send Boost Fix

TaylorBeeston:
- **:arrow_up: Bump TS to 5.5.2**
- **:sparkles: Allow changing boost VCs before sending them**
- **:bookmark: Changesets**

# Overview

We've been stuck on TS 5.0.4 for a while, and I finally got that fixed today!

Also, right now, you can not actually change a boost before sending it (such as to add a `credentialStatus` field),
nor can you easily get the Brain Client object from the Network plugin

#### 🥴 TL; RL:
Upgrades TypeScript across the whole monorepo and allows for better flexibility when sending boosts

- Upgrade to TS 5.5.2
    - In order to make this work, I had to do this weird change where I can't use `[T] extends [1 & T]` directly to determine if `T` is any, but I _can_ put that into a helper type `type IsAnyOrNever = [T] extends [1 & T] ? true : false` and use that: `IsAnyOrNever<T> ? true : false`. Don't ask me why, but that was the breaking change from the TS upgrade =(
- Update `sendBoost` type to allow overriding the boost
    - Very simply just allows passing in an override function that gives you the unsigned VC, and then expects you to return an unsigned VC that it will use to send the boost
    - In order to keep this backwords compatible, I made the last parameter _either_ a boolean or an object
- Add `getLCNClient` method to network plugin
    - This just allows for full flexibility when interacting with the LCN. Let's consumers do whatever they want with the TRPC client!

Just the annoying backwards compat boolean or object thing

#### 🔍 Types of Changes

- [x] No
- [ ] Yes

# Testing

Build _everything_ (`pnpm exec nx build cli --skip-nx-cache`) and then open the CLI and try different
things like issuing creds. Also, try grabbing the client in the CLI via `learnCard.invoke.getLCNClient`
and hitting the API directly (e.g. `client.utilities.healthCheck.query()`)

#### 🧪 Code Coverage

# Documentation

#### 📊 Storybook

# ✅ PR Checklist

### DIFF
--- a/.changeset/bright-shrimps-learn.md
+++ b/.changeset/bright-shrimps-learn.md
@@ -1,0 +1,5 @@
+---
+'@learncard/network-brain-client': patch
+---
+
+Expose Client type
--- a/.changeset/healthy-rivers-kiss.md
+++ b/.changeset/healthy-rivers-kiss.md
@@ -1,0 +1,5 @@
+---
+'@learncard/network-plugin': patch
+---
+
+Allow overriding a boost VC before sending it
--- a/.changeset/odd-otters-watch.md
+++ b/.changeset/odd-otters-watch.md
@@ -1,0 +1,5 @@
+---
+'@learncard/core': minor
+---
+
+Update to TS 5.5.2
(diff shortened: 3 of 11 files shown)


Custard7 (approved):
LGTM 👍✅

PR #577, opened 2024-12-05
Improve Initialization Speed

TaylorBeeston:
# Overview

While I was working on the family stuff, I realized that we are unnecessarily `await`ing multiple
different tRPC requests to LearnCloud and LCN when calling `initLearnCard`

#### 🥴 TL; RL:
Removes the `await` from the relevant calls

For almost everything I was able to move the `await` to be in front of any method that actually needed
these things, with the only exception being the call to `learnCard.id.did()` after you've gotten your LCN account
If it's totally necessary, you can forcibly make sure that is properly waited for by calling `await learnCard.invoke.getProfile`
prior to calling it, but I think that in every current case, there is a reasonable delay between when
you would initialize the wallet vs getting the did.

This situation _does_ bother me though, and there's a couple of more robust fixes:
- Making `id.did()` Return a promise so that we can `await` the first network call in the LCN plugin
- Adding the concept of `signal`s to `@learncard/core` so that we can register that first call as like an init signal and then await it in `@learncard/init`
- Adding a convention method like `__init__` that a user could `await` to make sure all plugins are ready

I think all of those options have a lot of various tradeoffs, with `id.did` becoming async being my favorite, though it's a _huge_ breaking change, so I don't think it's feasible.
Signals seems second best, but if I do that, I want to have like a really solid implementation for it, so I'm not gonna add those in this PR!

#### 🔍 Types of Changes

- [x] No
- [ ] Yes

# Testing

`pnpm exec nx start cli --skip-nx-cache` And then feel the speed when you call `initLearnCard`, especially with `network: true`!

#### 🧪 Code Coverage

I added an E2E test demoing the did problem mentioned above. In practical scenarios, you usually wouldn't need
to think about it because the request is so quick, but in script-type scenarios such as a test like that,
awaiting `getProfile` is the solution!

# Documentation

#### 📊 Storybook

# ✅ PR Checklist

### DIFF
--- a/packages/plugins/learn-card-network/src/types.ts
+++ b/packages/plugins/learn-card-network/src/types.ts
@@ -37,6 +37,7 @@ import { VerifyExtension } from '@learncard/vc-plugin';
 
 /** @group LearnCardNetwork Plugin */
 export type LearnCardNetworkPluginDependentMethods = {
+    __init__?: () => Promise<void>;
     getDIDObject: () => DID;
     getDidAuthVp: (options?: ProofOptions) => Promise<VP | string>;
     issueCredential: (
--- a/packages/plugins/learn-card-network/src/types.ts
+++ b/packages/plugins/learn-card-network/src/types.ts
@@ -37,7 +37,6 @@ import { VerifyExtension } from '@learncard/vc-plugin';
 
 /** @group LearnCardNetwork Plugin */
 export type LearnCardNetworkPluginDependentMethods = {
-    __init__?: () => Promise<void>;
     getDIDObject: () => DID;
     getDidAuthVp: (options?: ProofOptions) => Promise<VP | string>;
     issueCredential: (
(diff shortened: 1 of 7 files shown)


packages/plugins/learn-card-network/src/types.ts
```diff
@@ -37,6 +37,7 @@ import { VerifyExtension } from '@learncard/vc-plugin';
 
 /** @group LearnCardNetwork Plugin */
 export type LearnCardNetworkPluginDependentMethods = {
+    __init__?: () => Promise<void>;
```
Custard7, line 40: Should we add some documentation/comment around using this in scripts/tests

TaylorBeeston: I actually didn't end up taking this approach! LearnCloud plugin was able to do it all behind the scenes, and the LCN plugin can just have you await `getProfile` (which is what the test does!)

I'll add some docs for that though =)

PR #578, opened 2024-12-11
Add Display Field to LCN Profiles

TaylorBeeston:
# Overview

To make the personal LearnCard screen work, we need to include more display metadata for profiles

#### 🥴 TL; RL:
Adds a `display` object with these fields as requested:
```ts
export type profileDisplay = {
    // container styles
    backgroundColor?: string;
    backgroundImage?: string;
    fadeBackgroundImage?: boolean;
    repeatBackgroundImage?: boolean;

    // id styles
    fontColor?: string;
    accentColor?: string;
    accentFontColor?: string;
    idBackgroundImage?: string;
    fadeIdBackgroundImage?: boolean;
    idBackgroundColor?: string;
    repeatIdBackgroundImage?: boolean;
};
```

Because this is a nested object, I had to restructure the `Profile` neo4j object to use flatten/inflate
during storage. This time, I decided that the access layer should just totally hide that from everything
else (as opposed to boosts, where application code is more responsible), which kind of resulted in a
pretty big refactor, but I think this was actually a net positive, and I want to in the future move
towards instances never being exposed and things always being done via `QueryBuilder`, because I think
it's more maintainable.

Confusingly, you now should pretty much _never_ use the `Profile` model directly! It will strip out
the display field =(

#### 🔍 Types of Changes

- [ ] No
- [x] Yes
`Profile` can not be used directly, and I think that there is like a major restructuring that should
happen to make this more obvious. For now though, it's fine to just use access layer methods as-is
and it will cause no harm

# Testing

Run the test suite! There are tests for the display fields =)

#### 🧪 Code Coverage

Added tests for creating and updating display fields

# Documentation

#### 📊 Storybook

# ✅ PR Checklist

### DIFF
--- a/.changeset/clean-gifts-film.md
+++ b/.changeset/clean-gifts-film.md
@@ -1,0 +1,6 @@
+---
+'@learncard/network-brain-service': patch
+'@learncard/types': patch
+---
+
+Add display field to profiles
--- a/packages/learn-card-types/src/lcn.ts
+++ b/packages/learn-card-types/src/lcn.ts
@@ -3,6 +3,21 @@ import { z } from 'zod';
 import { PaginationResponseValidator } from './mongo';
 import { StringQuery } from './queries';
 
+export const LCNProfileDisplayValidator = z.object({
+    backgroundColor: z.string().optional(),
+    backgroundImage: z.string().optional(),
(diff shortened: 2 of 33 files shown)


gerardopar (approved):
LGTM! 🚀

PR #1198, opened 2026-04-29
chore: LC-1829 Update jira tagging to associate jira with release

Custard7:
## Summary

Extends the `tag-jira-release` job in [.github/workflows/deploy.yml](/.github/workflows/deploy.yml:0:0-0:0) so that, in addition to creating a Jira version on each production/scouts release, it now **associates every Jira ticket touched since the previous release with that version's `fixVersions` field**. The release page in Jira will populate automatically with the relevant tickets.

## Motivation

Previously, the workflow only created an empty Jira "version" container — tickets had to be added to the release manually in the Jira UI, which never actually happened. Releases existed in Jira but had no tickets attached, making release notes useless.

### New: `Checkout Repo` step (`fetch-depth: 0`)

The `tag-jira-release` job had no checkout, so it had no git history to scan. Added a full-history checkout as the first step.

### New: previous-release SHA lookup (before version creation)

Before creating the new Jira version, we now query:
`GET /rest/api/3/project/{id}/version?orderBy=-releaseDate&status=released&maxResults=10`

 …and walk the resulting versions in release-date-desc order, extracting `commit <SHA>` from each `description`. The first SHA that resolves locally (`git cat-file -e`) becomes `PREV_SHA` — the commit lower bound for issue scanning.

This works because every version we create already embeds the source commit SHA in its description (`Deployed from commit ${GITHUB_SHA} via GitHub Actions`). The lookup walks up to 10 versions back, so a single manually-edited description in Jira UI doesn't break the chain.

### New: `fixVersions` linking (after version creation)

After the version is created, we:
1. Run `git log <PREV_SHA>..HEAD` (or `HEAD~100..HEAD` on first-ever release) to extract issue keys.
2. Match `<PROJECT_KEY>-<NUMBER>` **case-insensitively** (`LC-123`, `lc-123`, `Lc-123` all match) and normalize to uppercase before calling Jira.
3. For each unique key, `PUT /rest/api/3/issue/{key}` with `{"update":{"fixVersions":[{"add":{"name":"<release>"}}]}}`. Using `add` (not `set`) appends rather than overwrites any existing fixVersions on the ticket.
4. Each PATCH is logged with HTTP code; per-issue failures become `::warning::` annotations rather than failing the job.

## How the chain bootstraps

| Run | Lookup result | Range used |
|-----|--------------|------------|
| First-ever | No previous version | `HEAD~100..HEAD` (fallback) |
| Run N | Previous version's description yields SHA from run N-1 | `<run N-1 SHA>..HEAD` — exact commits between release PR merges |

Because each successful run leaves a SHA breadcrumb in its own version description, every subsequent run finds the precise commit boundary regardless of how many intermediate non-release pushes landed on main.

## Robustness

- **Multi-version walk** — looks at the last 10 released versions, not just 1. Stale/edited descriptions don't break the lookup.
- **SHA validation** — `git cat-file -e` ensures we never use a force-pushed-away SHA.
- **Permissive hex pattern** — `[0-9a-f]{7,40}` matches both abbreviated and full SHAs.
- **Non-fatal per issue** — one ticket failing (e.g. someone deleted it) doesn't stop the rest from being linked.
- **Graceful skip** — if the version creation itself fails, we `exit 0` rather than try to link to a non-existent release.

## Sample log output
```
Previous Jira release commit: a1b2c3d4e5f6... Scanning commits in range: a1b2c3d4e5f6...HEAD Linking issues to release 2026-04-29-067a8d3: LC-1673 LC-1779 LC-1780 ok LC-1673 (HTTP 204) ok LC-1779 (HTTP 204) ok LC-1780 (HTTP 204) Linked 3 issue(s) to 2026-04-29-067a8d3; 0 failure(s).
```

## Operational notes

- **Don't strip `commit <SHA>` from version descriptions in Jira UI** — that's the breadcrumb the next run uses to find its commit range. Editing other parts of the description is fine.
- **Commit hygiene matters** — only commits/PR titles that mention `LC-###` will be linked. Squash-merged PRs with the ticket key in the PR title work great.
- **Jira API token** must have `Browse Projects` + `Edit Issues` permissions on the LC project. (Already validated: the existing token on the most recent successful run.)

## Testing

- Workflow only runs on push to main with `is_release == true`, or on manual dispatch with `target-environment` of `production` or `scouts`. Will be exercised by the next release PR merge.
- Job is `continue-on-error: true`, so even if the linking has issues in production, the deploy itself isn't impacted.

---

### ✨ PR Description
Purpose: Enhance Jira release automation to link issues to releases by tracking commit ranges between versions and associating relevant issue keys via fixVersions.

Main changes:
- Add repository checkout step with full history to enable git commit range analysis and version tracking
- Implement previous release SHA extraction from Jira version descriptions to establish commit range boundaries
- Extract issue keys from commit messages and link them to release via Jira fixVersions API with fallback commit depth


Custard7:
@claude please review

Custard7:
@claude please re-review:

Addressed
#1 HTTP validation on versions query (deploy.yml:825-852) — now captures HTTP code via curl -w, only parses .values[] on 2xx, and emits a ::warning:: with body on failure so the fallback is explicit and auditable, not silent.
#4 Configurable fallback depth (deploy.yml:887) — reads JIRA_FALLBACK_DEPTH env var, defaults to 100. If you want to bump it for a huge first release, set it once in the step's env: block.
#6 Range validation (deploy.yml:897-903) — git rev-list --count "$RANGE" check before scanning. Catches shallow-clone edges and unreachable SHAs.
#7 Tighter key regex (deploy.yml:912-913) — replaced \b with explicit [^A-Za-z0-9_] boundaries (plus start/end anchors), then a second grep strips the captured boundary chars so only the key is emitted. Prevents XLC-123 from matching when key is LC.
Not addressed (and why)
#2 Format parsing robustness — the "commit <SHA>" format is set by us 10 lines above (at deploy.yml:864). The multi-version walk already handles one-off manual edits. Adding alternative patterns invites more false positives than it prevents failures.
#3 Rate limiting — Jira Cloud uses cost-based per-tenant limits (~100+ req/sec headroom for most tenants); sequential curls for a typical release (5–30 tickets) won't come close. Adding sleep slows every normal run for a risk that likely never materializes. If you ever do hit a 429, the per-issue warning already surfaces it — we can add backoff reactively.

PR #1219, opened 2026-05-07
feat: [LC-1837] Add tracing to brain service

Custard7:
#### 🎟 Relevant Jira Issues
LC-1837

The brain-service has a tracing system (`@tracing`) capable of producing per-request span trees, but coverage today is essentially **one happy path**: only `boosts.ts` (the `sendBoost` mutation) is instrumented end-to-end. The other 22 of 23 route files, all ~35 access-layer files, and all of the helper layer are untraced — meaning we currently cannot answer the question "which of our endpoints is slow, and where is the time going?" for production traffic.
Rather than manually wrap ~600 individual call sites, this PR adds tracing at **two infrastructure choke points** to get near-complete coverage with a tiny diff.

#### 🥴 TL; RL:
Two surgical changes that together produce a span for every tRPC procedure and every Cypher query in the brain-service, with no per-call-site work.

#### 💡 Feature Breakdown
**1. tRPC tracing middleware** (`src/routes/index.ts`)
Adds a single middleware as the first `.use()` on `openRoute`, the base procedure all other routes chain off of. Every tRPC query/mutation now produces a top-level span named `route:<type>:<path>` (e.g. `route:mutation:boost.send`, `route:query:appStore.browseListedApps`).
**2. Auto-traced Neogma queryRunner** (`src/instance.ts`)
Wraps `neogma.queryRunner.run` once at the instance layer. Because `QueryBuilder.queryRunner = neogma.queryRunner`, this single wrap covers both raw `neogma.queryRunner.run(...)` calls AND `new QueryBuilder(...).run()` chains — i.e. effectively all DB operations in the access layer and helpers. Span names are derived from the first verb + first label in the Cypher (e.g. `db:MATCH AppStoreListing`, `db:CREATE Profile`); the full Cypher (truncated to 500 chars) is attached as span metadata for the JSON / Sentry providers.
Combined effect on the trace tree:
TRACE:abc123 ✓ route:mutation:boost.send 234ms
TRACE:abc123   ✓ db:getBoostByUri 12ms        ← existing manual span
TRACE:abc123     ✓ db:MATCH Boost 11ms        ← NEW auto-span (child)
TRACE:abc123   ✓ db:MATCH Profile 8ms         ← NEW auto-span (no manual parent)
Coverage delta:
- Routes traced: **1/23 → 23/23** (100%)
- DB calls traced: **~10 manual sites → ~95%+ via the wrap** (every queryRunner.run + every QueryBuilder.run)

- Console output is now opt-in (default off). Local devs and tests get a clean console; enable verbose tracing with TRACE_CONSOLE=true pnpm dev when chasing performance issues.
- **`createContext` is outside the route span.** Auth resolution (JWT verify, profile lookup, challenge validation) runs before middlewares, so its DB calls produce top-level spans rather than nesting under a route. Structurally correct, but worth knowing when reading traces.
- **Redundant span left in `boosts.ts`.** `sendBoost` still has its own `trace('route', 'send', ...)` wrap inside the handler, which now becomes a child of the auto `route:mutation:boost.send` span. Functionally fine, just one extra layer in the tree. Trivial cleanup deferred to a follow-up PR.
- **Tier 1 (external HTTP wrapping) deliberately deferred.** Postmark, SQS, webhook fan-out, OIDC JWKS, federation registry, etc. are still untraced. Decision: ship the structural changes alone, then use the resulting trace data to *prioritize* which external call sites actually warrant `traceHttp` wraps. Avoids guessing which 9 of 9 sites matter when probably only 3-4 do.
- **No new dependencies, no env-var changes.** All providers self-enable based on existing env (`TRACE_CONSOLE`, `TRACE_JSON`, `SENTRY_DSN`).

#### 🔍 Types of Changes
-   [ ] Bug fix
-   [ ] New feature
-   [ ] Breaking change
-   [x] Chore (observability infrastructure)

-   [x] Yes (minor, acknowledged):
    - Redundant `trace('route', 'send', ...)` wrap inside `boosts.ts:sendBoost` is now superseded by the middleware. To be removed in a follow-up.
    - Tier 1 (HTTP/crypto/cache wrapping) is the natural follow-up tracked separately.

# Testing

1. Pull the branch and start the brain-service locally (`pnpm dev` from `services/learn-card-network/brain-service/`).
2. Make any tRPC call against it (e.g. via the LearnCard app, or directly via curl/Postman).
3. Verify the console output now shows a `route:<type>:<path>` span at the top of each request, with auto-generated `db:MATCH X` / `db:CREATE Y` child spans for every Cypher query the request triggers.
4. Sanity check: `sendBoost` should still produce its existing manual `traceDb` spans (e.g. `db:getBoostByUri`), but those are now nested under the new `route:mutation:boost.send` parent rather than under a manual `route:send` parent.
5. Confirm error paths still work — trigger a 4xx/5xx response and verify the span is marked with status `error`.
6. Confirm `TRACE_CONSOLE=false` cleanly disables console output without breaking JSON or Sentry providers.

N/A — backend-only change. Verify via local brain-service or staging.

#### 🧪 Code Coverage
No new unit tests added. Rationale:
- The wrappers are pass-through by construction (preserve types, preserve return values, preserve thrown errors via the existing `trace()` helper which is already covered).
- The change is best validated by running the existing brain-service test suite end-to-end (`pnpm test -- run`) and by observing real traces during QA.
# Documentation

**Internal/AI Docs**
-   [x] **Code comments/JSDoc** — Added a JSDoc block on the queryRunner wrap explaining the parent/child span behavior with manual `traceDb` calls and the metadata visibility contract (which providers see the truncated cypher).

#### 💭 Documentation Notes
The existing `services/learn-card-network/brain-service/AGENTS.md` "Tracing & Performance Monitoring" section already documents the `trace*` API and provider config, and it remains accurate after this change. No user-facing docs apply (purely internal observability infra).

# ✅ PR Checklist
-   [x] Related to a Jira issue (LC-1837)
-   [x] My code follows style guidelines
-   [x] I have manually tested common end-to-end cases
-   [x] I have reviewed my code
-   [x] I have commented my code, particularly where ambiguous
-   [x] New and existing unit tests pass locally with my changes
-   [x] I have completed the Documentation Checklist above
-   [x] N/A — no user-facing analytics implications

-   [x] I have thoughtfully considered the security implications of this change. Truncated Cypher in span metadata could in theory contain sensitive identifiers (DIDs, profileIds, credential URIs); these are not secrets but are not anonymous either. Visible only to operators with access to the JSON log stream or Sentry — same trust boundary as Sentry breadcrumbs / existing logs.
-   [x] This change does not expose new public-facing endpoints

---


Custard7:
@claude please review

PR #1220, opened 2026-05-07
[LC-2065] Automate DIDKit WASM updates

TaylorBeeston:
# Overview

Jira: [LC-2065 — Automate DIDKit and SSI WASM updates](https://welibrary.atlassian.net/browse/LC-2065)

I was manually bumping DIDKit/SSI and rebuilding the WASM, which was annoying and easy to get wrong. This adds a reproducible, event-driven update path instead.

## What changed

- PRs build and smoke-test the exact checked-in DIDKit/SSI submodules without touching AWS or changing the default URL.
- Repository dispatches and manual runs build upstream updates with a tracked Cargo lock.
- Binaryen 130 is checksum-pinned and applies `wasm-opt -Oz`; the resulting WASM must initialize and derive an Ed25519 `did:key` before upload.
- Content-addressed artifacts upload to the existing S3/CloudFront setup, verify Brotli delivery, then open a patch changeset PR.
- `.gitmodules` now points at the `learningeconomy/didkit` and `learningeconomy/ssi` repositories.

## Upstream triggers

There is no nightly polling. These companion PRs dispatch the exact triggering SHA on every push to their respective `main` branch:

- [learningeconomy/didkit#7](https://github.com/learningeconomy/didkit/pull/7)
- [learningeconomy/ssi#4](https://github.com/learningeconomy/ssi/pull/4)

Merge this receiver first. Before merging either sender, add `LEARNCARD_DISPATCH_TOKEN` to that repository; it needs permission to call `POST /repos/learningeconomy/LearnCard/dispatches`. The receiver remains the publication gate: it builds, tests, runtime-smokes, and verifies the artifact before changing any consumed URL.

## Important tradeoff

Ubuntu's Binaryen 108 produced a WASM table that failed during wasm-bindgen initialization. Binaryen 130 works, so the workflow pins that release and verifies its upstream SHA-256. The optimized artifact is 9,001,604 bytes, below CloudFront's 10,000,000-byte compression ceiling; a live cache-miss request returned a 2,810,095-byte Brotli response.

Publish runs are serialized and never cancelled mid-upload or mid-PR creation. Only redundant PR validation runs can cancel each other.

## Verification

OMP verified:

- Locked and lock-refresh WASM builds against current DIDKit/SSI.
- Binaryen 130 output initializes and derives an Ed25519 `did:key`.
- `bunx nx build didkit-plugin` and `bunx nx test didkit-plugin`.
- Native DIDKit regression suite.
- Workflow syntax with `actionlint`; build script syntax with `bash -n`.
- [Forced publish run](https://github.com/learningeconomy/LearnCard/actions/runs/30659726503): build, tests, S3 upload, CloudFront Brotli verification, and update PR creation all passed.

No user-facing docs or changeset are needed here; generated update PRs include their own patch changeset.

- OMP

### ✨ PR Description
Purpose: Automate DIDKit WASM build and deployment pipeline with reproducible builds, version-locked artifacts, and CI/CD integration for consistent binary publishing.

Main changes:
- Extracted DEFAULT_DIDKIT_WASM_URL constant to enable automated URL updates via set-default-wasm-url.mjs script
- Added build-wasm-from-submodules.sh for reproducible WASM compilation with wasm-pack and wasm-opt optimization
- Created comprehensive GitHub Actions workflow with validation, S3 upload, CloudFront verification, and automated PR generation
- Updated submodule URLs to learningeconomy forks and added smoke test script for WASM integrity validation


TaylorBeeston:
@claude pls review

TaylorBeeston:
@wendell-mossberry Addressed this in `b581b68e2` — thanks for the thorough pass!

The new smoke test immediately earned its keep: CI caught Ubuntu’s Binaryen 108 producing a WASM table that couldn’t grow at runtime. I removed the extra `wasm-opt` pass, kept the Rust release optimizations, and the generated artifact now initializes successfully in CI.

Also fixed publish cancellation, added the inferred Nx test target, moved shell values through `env`, made the wasm-pack pin cache-safe, gated AWS/upload behind real changes, hardened temporary lock cleanup, documented the deliberate skip behavior/trust boundary, moved the lock outside Nx’s `scripts` input, and tightened the PR scope/changeset notes. The fork transfer and a durable team reviewer remain separate ownership follow-ups since there isn’t an established team handle in this repo yet.

All checks are green now.

— OMP

TaylorBeeston:
Follow-up after the WASM size question: I restored `wasm-opt -Oz`, but pinned Binaryen 130 with its upstream SHA-256 instead of using Ubuntu’s broken Binaryen 108 package.

I also ran the whole publish path for real. The generated artifact initialized correctly, uploaded to S3, and came back through CloudFront with `Content-Type: application/wasm`, `Content-Encoding: br`, and the immutable one-year cache policy. The live transfer was 2,810,095 bytes versus 9,001,604 bytes uncompressed. The workflow now fails before opening its update PR if that Brotli response is missing.

All PR checks and the forced publish run are green.

- OMP

Custard7:
@TaylorBeeston Cool, okay, so:

 1. set-default-wasm-url.mjs only works once
 • Its regex matches the `arg: InitInput | Promise<InitInput> = '...'` signature on one line. The new content-addressed URL makes that line ~170 chars, and `.husky/pre-commit` → lint-staged → prettier (printWidth 100) wraps the union across 3 lines when create-pull-request commits.
 • That's exactly what your forced run produced in #1446 — index.ts came out as the multi-line `arg:\n | InitInput\n | Promise<InitInput> = '…'`.
 • So on the next dispatch the regex misses and the script exits 1, after the full WASM build. And since 1444 already carries the wrapped format, it'd actually fail on the first real run, not the second.
 • Suggested fix: pull the URL into its own `export const DEFAULT_DIDKIT_WASM_URL = '...'` and rewrite that instead — a lone string constant can't get re-wrapped, so it's prettier-proof. Maybe also assert the URL actually changed after the rewrite so a silent no-op can't reach S3. Does that match how you were thinking about it?

 Minor:
 • CloudFront verify is --retry 10 --retry-delay 3 ≈ 30s for a first cache-miss on a 9MB object — maybe add --retry-max-time 300.
 • The URL rewrite happens before the `git diff --quiet -- …/src/didkit` check, so `has_changes=false` is basically unreachable. Harmless, just misleading.
 • Wouldn't mark `Validate DIDKit WASM build` as required since it's paths-gated — unrelated PRs would hang.
 • FYI gitStream's "missing daily schedule" comment is wrong, your description explicitly says dispatch-only. Ignoring it.

Custard7 (approved):

PR #1221, opened 2026-05-08
fix: [LC-1841] Eliminate cold-start flash sequence

Custard7:
On native cold start (and to a lesser degree on web), authenticated users see up to three visible loader/login screens flash by before landing on their home page. This PR closes the race condition that causes the LoginPage to mount during boot, and hands the entire bootstrap window over to the native Capacitor splash so authenticated users see a single fade from splash → app — zero JS-side loaders.

## Problem
Reproducing on Android (kill app, wait 1+ hours, reopen):

0ms   ─ Native splash (green #00BA88)
~600ms ─ Splash hides
~700ms ─ 🟣 LoadingPageDumb           ← FLASH #1 (Suspense fallback for FullApp chunk)
1500ms ─ 🟣 LoginLoadingPage         ← FLASH #2 (cycling-color typewriter)
2000ms ─ 🟣 LoginPage                ← FLASH #3 (full login UI with social buttons)
~2200ms ─ Home page

On web first-load, only the LoadingPageDumb flash is visible because chunks load fast enough that the LoginPage race window stays sub-frame.

## Root cause
There are two competing "is logged in?" sources of truth, and one of them was being read inside the boot race window:
| Signal | Source | When it becomes true |
|---|---|---|
| `useAppAuth().walletReady` | `!!wallet` (BespokeLearnCard instance) | After `initLearnCard()` resolves in a separate `useEffect` |
| `useIsLoggedIn()` | `currentUserStore.currentUserIsLoggedIn` | After wallet-init writes back to `currentUserStore` |
`AppRouter.tsx` was gating the loader on `coordinatorLoading`, which only covers `authenticating | checking_key_status | deriving_key`. The moment the coordinator hit `state.status === 'ready'` (private key derived), `coordinatorLoading` flipped to false — but `walletReady` was still false because the wallet object hadn't been constructed yet. That ~100–300ms gap unmounted the loader, mounted `<Routes>`, and `<Routes>` saw `useIsLoggedIn() === false` (currentUserStore not yet populated) and redirected to `/login`. The wallet then finished a moment later, `useIsLoggedIn()` flipped true, and the LoginPage's own redirect-on-mount effect pushed the user to `/launchpad`.
The Suspense fallback flash on top of that was caused by `SplashScreen.hide()` being called in `index.tsx` *before React even mounted*, removing the only thing that could bridge the chunk-load gap.

### `src/AppRouter.tsx` — Fix A (race) + Fix C native-side
- Replaced the `initLoading` gate to keep the loader up during **every** transitional auth state, not just the three coordinator-loading states. The new gate stays up unless we're in one of four decidable states: `walletReady`, `idle`, `needs_recovery`, or `error`. This closes the `ready`-without-wallet race window.
- Added a `useEffect` that hides the native splash with a 200ms fade once the gate flips false (i.e., once we know what to render).
- Added an 8-second safety-net `useEffect` that hides the splash even if auth init hangs (e.g., slow Keystore read on Android, SSS server unreachable).

### `src/index.tsx` — Fix C web-side
- Removed the eager `SplashScreen.hide()` that was firing before React mounted.
- Added a 10-second React-mount-failure safety net: if React never mounts at all (LaunchDarkly init throws, bootstrap rejects, chunk download fails), the splash dismisses anyway. This timer lives outside React on purpose.

### `capacitor.config.ts` + tenant overrides — Fix C config
- `launchAutoHide: true` → `false` in three places:
  - `apps/learn-card-app/capacitor.config.ts` (default)
  - `apps/learn-card-app/environments/learncard/assets/config/capacitor.config.json`
  - `apps/learn-card-app/environments/vetpass/assets/config/capacitor.config.json`
- `launchShowDuration: 6000` is left untouched for documentation, but is unused with `launchAutoHide: false` — JS now owns the dismiss lifecycle end-to-end.

### Android/iOS cold start (authenticated user)
**Before:** green splash → 🟣 LoadingPageDumb → 🟣 LoginLoadingPage → 🟣 LoginPage flash → home (≈2.2s, three visible flashes)
**After:** green splash holds through entire bootstrap → 200ms fade → home (≈1.5–2s, zero JS-side loaders)

### Android cold start (logged-out user)
**Before:** green splash → 🟣 LoadingPageDumb → 🟣 LoginLoadingPage → LoginPage
**After:** green splash holds until `state.status === 'idle'` → 200ms fade → LoginPage

### Web first-load
- Fix A still applies — no LoginPage flash, even on web.
- The `LoadingPageDumb` Suspense fallback for the lazy `FullApp` chunk still shows briefly. Fully eliminating that requires Fix B (un-lazy `FullApp`) or Fix D (paint splash from `index.html`), which are intentionally out of scope for this PR.

## Deployment steps
After merging, anyone working with native builds needs to re-run the tenant config prep so the updated `capacitor.config.json` propagates:
```bash
cd apps/learn-card-app
pnpm prepare-config learncard   # or vetpass, as needed
npx cap sync
```
CI deploy paths already run prepare-native-config.ts, so no infra changes are required.

Testing

I verified:
- LSP diagnostics clean on all three TS/TSX files (AppRouter.tsx, index.tsx, capacitor.config.ts).
- No new as any, @ts-ignore, or type-error suppressions.

Recommended manual testing:
1. Android cold-start, authenticated: kill app from recents, wait 1+ hours, reopen. Should see green splash → fade → home, no purple flashes.
2. Android cold-start, logged-out: sign out, kill app, reopen. Should see green splash → fade → login.
3. iOS cold-start, authenticated: same as #1 but iOS-side. Verify Keychain read doesn't extend boot beyond the 8s safety net.
4. Web first-load (LearnCard tenant, then VetPass tenant): confirm no LoginPage flash on either tenant. Brief LoadingPageDumb is expected (Suspense fallback for lazy FullApp chunk).
5. Slow-network simulation: throttle to 3G in DevTools / Android emulator; confirm splash never sticks past 10 seconds (the React-mount safety net).
6. Recovery flow: trigger needs_recovery (e.g., new device, no local share). Splash should dismiss and RecoveryFlowModal should overlay normally.
7. Tenant override sanity check: run pnpm prepare-config vetpass and confirm the synced native config reflects launchAutoHide: false.

I have not run the Playwright suite for this PR (no behavior changes that affect existing test selectors), but it should pass clean — the only test-visible difference is that the splash takes a bit longer to dismiss in non-headless runs.

Risks / edge cases considered
- Splash never hides (auth init hangs unrecoverably): handled by the 8s timer in AppRouter and the 10s React-mount-failure timer in index.tsx.
- Multiple SplashScreen.hide() calls: Capacitor treats subsequent calls as no-ops; the controlled hide + safety nets don't conflict.
- Logout during boot: state.status returns to 'idle', gate flips false, splash dismisses, LoginPage renders. Behaves correctly.
- Web platform: all Capacitor.isNativePlatform() guards short-circuit, so the splash code is dead on web. Fix A still applies and prevents the LoginPage race.
- Tenant config drift: changes were applied to both existing tenants (LearnCard, VetPass). Any new tenant created via pnpm create-tenant will inherit launchAutoHide: false from the default capacitor.config.ts. Add a doc note to environments/README.md if we want to make this explicit (not done in this PR).

---

### ✨ PR Description
Purpose: Eliminate cold-start UI flash by deferring native splash screen hide until authentication state stabilizes and wallet initialization completes.

Main changes:
- Disabled automatic splash screen hide (launchAutoHide: false) across Capacitor configs to keep native splash visible during JS bootstrap
- Implemented AppRouter logic to conditionally render loader or routes based on wallet readiness and coordinator state, hiding splash only when auth settles
- Added dual safety timeouts: AppRouter hides splash at 8s, index.tsx hides at 10s as fallback if React mount or auth init hangs


Custard7:
@claude please review

gerardopar (approved):
LGTM! 👍🏽

PR #1222, opened 2026-05-08
fix: [LC-1842] Eliminate avatar flicker and '#' fallback

Custard7:
The user-profile avatar in the top-right QRCodeScannerButton (and any other consumer of UserProfilePicture) had three visible bugs: it flickered on page switch, frequently displayed '#' as a "broken" fallback, and showed the browser's broken-image glyph for a beat while the avatar URL was fetching. This PR closes all six root causes with surgical changes to four files.

### Problems
1. Local useState mirroring React Query data in useGetCurrentLCNUser — every component mount started with lcnProfile = null for one frame, even when React Query had warm cached data. That's the page-switch flicker.
2. No staleTime on useGetProfile — React Query default of 0 means every mount triggers a background refetch. That's why every navigation refetched.
3. Sequential double-fetch in useGetCurrentLCNUser — two useGetProfile() calls hitting the same endpoint. Doubled the time-to-first-render of the avatar.
4. updateCurrentUserNameAndImage(displayName ?? '', image ?? '') — when the LCN profile had no displayName / image (sparse / brand-new accounts), this wiped the auth-provider-supplied values in currentUserStore, causing the entire fallback chain in UserProfilePicture to degrade to '#'.
5. UserProfilePicture final fallback was '#' — a hashtag character reads to users as "broken". Should be a person glyph.
6.  had no load/error handling — when src flipped from undefined → URL, the browser showed its native broken-image icon for ~50–500ms during the network fetch.
What changed

#### Before
- App open (cold cache): ~200ms broken-image icon → letter fallback → image swaps in (visible 3-state transition)
- App open (sparse LCN profile): letter fallback → '#' after sync wipes Firebase data → image (if it eventually loads)
- Page switch (warm cache): 1-frame letter fallback flicker → image
- Image load failure (404, CORS): browser broken-image icon, forever

#### After
- App open (cold cache): colored placeholder with letter glyph paints immediately → image fades in over 150ms (single smooth transition)
- App open (sparse LCN profile): auth-provider values preserved → letter fallback uses Firebase name / email (no '#')
- Page switch (warm cache): cached profile flows through synchronously, no flicker
- Image load failure: gracefully falls back to letter or person glyph, never broken-image icon

Cache contract change (worth noting)

useGetProfile now caches profile data for 5 minutes (staleTime: 5 * 60 * 1000). This is fine for read-only consumers, but mutations that update the user's profile must invalidate the query:
`queryClient.invalidateQueries({ queryKey: ['getProfile'] });`

I checked the existing profile-edit mutations and most already do this correctly. Worth a quick review during PR — if any mutation updates displayName or image and doesn't invalidate, the user would see stale data for up to 5 minutes after a self-edit.

If preferred, set staleTime: 30 * 1000 (30 seconds) to be more conservative until invalidation coverage is verified.

### Recommend manual testing:
1. Page-switch flicker: open /wallet, switch to /connect, switch back. Avatar should not flicker.
2. Cold-start avatar load: kill app, reopen with throttled network (3G in DevTools). Should see colored letter circle → fade to image, no broken-image glyph.
3. Sparse LCN profile: test with a user whose LCN profile has empty displayName/image (or brand-new signup). Avatar should show letter from Firebase name or email, never '#'.
4. Image fetch failure: in DevTools, block the avatar image URL pattern. Avatar should fall back to letter/glyph gracefully, not broken-image.
5. Self-edit profile: change displayName/image in profile settings. Avatar should update once mutation invalidates the query.
6. Other UserProfilePicture consumers: grep UserProfilePicture across the codebase — there are several call sites in side-menus, modals, etc. Verify their containers still size correctly (the new  has relative to layer over the absolute placeholder).

### Edge cases considered
- Container has no explicit height/width: the absolute placeholder needs a sized parent. All checked call sites pass explicit h-[Npx] w-[Npx] in customContainerClass. If a future consumer passes a container without sizing, the placeholder collapses and the image alone would still render in flow. Acceptable graceful degradation.
- overflow-hidden on parent: preserved by leaving customContainerClass untouched. Both the placeholder and the  clip cleanly inside the rounded container.
- Tailwind purge: the new classes (absolute, inset-0, transition-opacity, duration-150, opacity-0, opacity-100, w-1/2, h-1/2, opacity-90) are all standard utility classes — no purge changes needed.
- type import in useGetCurrentLCNUser: removed unused LCNProfile and useState imports. The hook now returns the React Query data type (QueriedProfile | null) directly. Downstream consumers that read .displayName, .image, .profileId, .did are unaffected because both types share those fields.
Out of scope (deferred)
- Duplicate UserProfilePicture in react-learn-card: there's a near-identical copy at packages/react-learn-card/src/components/UserProfilePicture/UserProfilePicture.tsx (the one you originally pointed at). If anything still imports from @learncard/react, it'll keep showing '#'. Worth either deleting the duplicate or re-exporting from learn-card-base. Not in this PR.
- useGetCurrentUser's secure-storage round-trip: adds another ~100ms of latency on cold start. Tied to the deeper auth-flash work from the earlier PR. Not in this PR.
- Avatar prefetch: a <link rel="preload" as="image"> injected during boot would put the avatar in browser cache before any component renders it. Worth adding to the tenant template pipeline. Not in this PR.


Custard7:
@claude please review

PR #1225, opened 2026-05-11
[LC-1644][LC-1811] Optimize sendCredential performance: backend + frontend + telemetry

goblincore:
# Overview

#### 🎟 Relevant Jira Issues

Fixes: [LC-1644](https://welibrary.atlassian.net/browse/LC-1644). Related to: [LC-1811](https://welibrary.atlassian.net/browse/LC-1811) (latency premise disproven — see [LC-1644 performance summary](https://github.com/learningeconomy/LearnCard/blob/feat/lc-1644-frontend-claim-perf/docs/internal/lc-1644-summary.md)).

**Supersedes [#1189](https://github.com/learningeconomy/LearnCard/pull/1189).** This branch now contains both the backend optimizations (originally PR #1189) and the frontend optimizations, so close #1189 in favor of this one as the canonical PR.

Brief loom just showing the testing harness used to collect the the data:

https://www.loom.com/share/56f03ebf83724583a92b1460dbee34e1

LC-1644 was opened to investigate a reported "5–10 second" sendCredential latency. Measurement showed three root causes:

1. **Backend** — every signing call paid full TCP+TLS handshake to a fresh `lca-api` Lambda container, blowing the p95 tail to ~4.5s.
2. **Frontend** — the claim phase fired four tRPC calls sequentially, so any single slow call stalled the whole flow (we observed one event at 31644ms).
3. **UX** — the credential preview flashed placeholder text ("Credential", "by [?]") during a brief enrichment window.

This PR fixes all three, adds permanent telemetry (gated behind a LaunchDarkly kill switch), and ships an admin bench panel for repeatable measurement.

**Scope note:** the backend perf work targets the LCA-managed Signing Authority only (`lca-api`). Partner-hosted SAs run on partner infra and are explicitly out of scope.

#### 🥴 TL; RL:

Backend I ran 200 iterations for each branch (200 unoptimized  runs vs 200 optimized runs) and compiled the numbers. Frontend is manual testing as it triggers the full modal ui flow more or less so I only ran around 10/10 for each of those but the numbers were consistent.

Tokyo client → US-east staging (US users will see ~half these absolute latencies):

| Metric | Before | After | Δ |
|---|---:|---:|---:|
| Backend p50 | 570ms | 292ms | −49% |
| **Backend p95 (slow tail)** | **4547ms** | **366ms** | **−92%** |
| Modal mount → credential resolved | ~500ms | **0ms** | −100% |
| Claim phase median | 4745ms | 1810ms | −62% |
| **Claim phase worst-case** | **31644ms** ⚠️ | 2046ms | **15.5×** |
| **Time-to-modal-interactive** | ~6000ms | **~1563ms** | **−74%** |

US-projected time-to-modal-interactive: **~500ms**. (staging)

#### 💡 Feature Breakdown

**Backend (`services/learn-card-network/brain-service`)**

- `undici.Agent` with `keepAlive: true` + HTTP/2 to the LCA SA — collapses the p95 tail (does 90% of the backend work in one commit)
- Parallelized Neo4j reads (listing/boost/recipient + owner/SA chain)
- Fire-and-forget `logActivity` + `sendBoost` via pre-generated activity UUID
- LRU cache + retry + tightened timeout on SA calls
- **Backend pre-resolves credential** before the fast-path response so the modal renders enriched on first paint (replaces an earlier frontend shimmer fallback)
- `bench.appevent.iteration` PostHog telemetry from the bench-panel harness — admin-tools manual trigger only; does **not** fire in production traffic
- Admin Tools → "AppEvent Perf Bench" panel for repeatable measurement

**Frontend (`apps/learn-card-app`)**
- `Promise.allSettled` parallelization of the 3 claim-phase tRPC calls (`addVCtoWallet` + `acceptCredential` + `queryNotifications`)
- Fire-and-forget `updateNotificationMeta` no longer blocks claim-success UI
- Threaded `credential` prop from response → modal skips the `wallet.read.get(uri)` round-trip
- `frontend.sendcredential.iteration` PostHog event with per-phase timings; joined to backend events via shared `run_id`
- LaunchDarkly flag **`enableSendCredentialPosthogTelemetry`** (default **off**) gates production emission. Bench-panel flows (`triggered_by_bench: true`) bypass for admin measurement. The backend mirrors this via env var **`ENABLE_SEND_CREDENTIAL_TELEMETRY=true`** on the brain-service Lambda (default off; brain-service has no LD client today).
- Bench panel "Trigger Single sendCredential" button for end-to-end flow timing

- **Backend pre-resolve adds ~100ms** to `request_to_response_ms` (1461 → 1561 median) in exchange for a strictly cleaner first-paint UX (no shimmer flash, no state swap). Well within iteration noise (prior n=7 spanned 1031–1642). 0ms `modal_mount_to_credential_resolved_ms` win preserved.
- **Telemetry off by default** — measurement campaigns flip the LD flag. Trade for zero PostHog cost from natural production flows.
- **LCA-only scope on backend perf** — partner-hosted SAs are unaffected and unmeasured. Tracked as a separate follow-up.

#### 🔍 Types of Changes

- [x] Bug fix (claim-phase sequential stall, first-paint placeholder flash)
- [x] New feature (bench panel, joinable backend+frontend telemetry, LD kill switch)
- [x] Chore (perf instrumentation, parallelization refactor)
- [ ] Breaking change

- [x] No — backend telemetry uses existing `posthog-node` lazy init pattern; frontend uses existing analytics-provider wiring; LD flag uses existing `useFlags()` pattern.

# Testing

You could test this locally though the data wouldn't be very useful imo. Before running anything, you need to set the POSTHOG api key in both the frontend and backend brain service .env configs (find in infisical) and also set the analytics provider env variable to posthog (and would need access to posthog to be able to view the analytics....). Then after that you can open up the testing panel as seen in the loom which is under the profile admin tools sidebar.
Another way is just testing the app as usual but like trying the flow via the app flow and seeing if it works.

1. **Visual smoke test (any tenant with a partner integration):** Trigger a send-credential from the partner app. Confirm the modal renders the correct achievement title + issuer on the first frame — no "Credential" / "by [?]" placeholder flash. Click Accept → confirm success state.
2. **Admin bench panel (staging):** Admin Tools → AppEvent Perf Bench. Keep "Auto-cleanup recipient" checked. Click **"Trigger Single sendCredential"** repeatedly. Validate each PostHog event has `fast_path: true`, `modal_mount_to_credential_resolved_ms = 0`, `outcome: claimed`.
3. **LD flag gating:** With `enableSendCredentialPosthogTelemetry = false` (default), confirm no `frontend.sendcredential.iteration` events fire from natural partner-driven flows. Flip the flag → events flow.

Chromium + Safari (desktop + mobile web). Bench panel is desktop-admin-only.

#### 🧪 Code Coverage

Unit tests on `posthog.helpers` (no-op when key unset, error swallowing) + the perf-helper module shape. 

# Documentation

**Internal/AI Docs**
- [x] **Code comments/JSDoc** — `sendCredentialFlow.helpers.ts` carries a lifecycle docstring for the phase markers
- [ ] AGENTS.md — no new patterns introduced

**Visual Documentation**
- [x] **Mermaid diagram** — instrumented flow + phase labels in the leadership writeup

#### 💭 Documentation Notes

Internal perf/refactor PR — no public API changes, no user-facing doc updates needed. Leadership-facing summary with the full instrumented flow, phase tables, results, and LC-1811 recommendation: [docs/internal/lc-1644-summary.md](https://github.com/learningeconomy/LearnCard/blob/feat/lc-1644-frontend-claim-perf/docs/internal/lc-1644-summary.md).

# ✅ PR Checklist

- [x] Related to a Jira issue ([LC-1644](https://welibrary.atlassian.net/browse/LC-1644))
- [x] Code follows style guidelines (eslint / prettier)
- [x] Manually tested end-to-end (staging, n=11 + visual QA)
- [x] Reviewed my code
- [x] Commented complex logic
- [x] Unit tests pass locally
- [x] Documentation Checklist completed
- [x] Product analytics considered — telemetry gated on LD flag

- [x] No "Do Not Merge" label
- [x] Security implications considered (LD flag gates analytics; no new endpoints; no auth changes)
- [x] No new public-facing unauthenticated endpoints

[LC-1644]: https://welibrary.atlassian.net/browse/LC-1644?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LC-1811]: https://welibrary.atlassian.net/browse/LC-1811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LC-1644]: https://welibrary.atlassian.net/browse/LC-1644?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### ✨ PR Description
Purpose: Optimize sendCredential flow performance across backend and frontend with comprehensive telemetry instrumentation to reduce end-to-end latency from 4–12 seconds to ~1.5 seconds.

Main changes:
- Backend optimizations: HTTP connection pooling for signing authority calls, parallel Neo4j queries with 30s LRU caching, and retry logic reduce p95 latency by 92%
- Frontend parallelization: concurrent credential claim operations and pre-resolved credential response eliminate sequential bottlenecks and 15× worst-case stalls
- PostHog telemetry integration: phase-level timing events joinable by run_id across backend and frontend, gated by LaunchDarkly flag with admin bench panel for measurement campaigns


goblincore:
@claude please review

services/learn-card-network/brain-service/src/helpers/signingAuthority.helpers.ts
```diff
@@ -46,143 +146,226 @@ export async function issueCredentialWithSigningAuthority(
...
+                    const result = await attempt(i);
+                    perf.mark(i === 0 ? 'http' : `http_retry_${i}`);
+                    perf.done({
+                        saEndpoint: signingAuthorityForUser.signingAuthority.endpoint,
+                        attempts: i + 1,
+                    });
+                    const benchCtx = benchContextStorage.getStore();
+                    if (benchCtx) {
+                        const captured = perf.capture();
+                        benchCtx.sa_http_ms = captured.phases.http ?? 0;
```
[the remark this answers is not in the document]
goblincore: resolved

goblincore:
@claude please review

apps/learn-card-app/netlify/edge-functions/tenant-config.ts
```diff
@@ -23,6 +23,21 @@ export default async (request: Request, _context: Context) => {
     const url = new URL(request.url);
     const hostname = url.hostname;
 
+    // Local development: return empty overlay so the baked /tenant-config.json
+    // wins the merge. This lets devs edit public/tenant-config.json (e.g. to
+    // flip analyticsProvider to "posthog") without the edge function clobbering
+    // their changes with the production-default config.
+    if (hostname === 'localhost' || hostname === '127.0.0.1') {
```
Custard7: Is this supposed to be here still?

packages/plugins/learn-card-network/src/plugin.ts
```diff
@@ -2432,6 +2432,17 @@ export async function getLearnCardNetworkPlugin(
...
+                await ensureUser();
+                // tRPC output is z.record(), cast to known shape
+                return client.appStore.benchAppEvent.mutate(input) as any;
+            },
+
+            cleanupBenchAppEventData: async (_learnCard, input) => {
+                await ensureUser();
+                return client.appStore.cleanupBenchAppEventData.mutate(input);
+            },
+
```
Custard7: Should these be added to this plugin?

services/learn-card-network/brain-service/src/routes/app-store.ts
```diff
@@ -528,13 +534,93 @@ const getListingOrThrow = async (listingId: string) => {
...
+};
+
+const getPrimarySigningAuthorityForIntegrationCached = async (integrationId: string) => {
+    const hit = integrationSaCache.get(integrationId);
+    if (hit !== undefined) return hit;
+    const value = await getPrimarySigningAuthorityForIntegration(integrationId);
+    integrationSaCache.add(integrationId, value, SA_CACHE_TTL_MS);
+    return value;
+};
+
```
Custard7: these should be moved to a separate cache layer file/folder

services/learn-card-network/brain-service/src/routes/app-store.ts
```diff
@@ -2747,6 +2891,154 @@ export const appStoreRouter = t.router({
             throw new TRPCError({ code: 'BAD_REQUEST', message: 'Unknown event type' });
         }),
 
+    // ==================== Bench Routes (gated client-side via LaunchDarkly) ====================
```
Custard7: These should be gated like how our test router works - right now, even with client side gating these would be always exposing these routes to the public. We should add an env var that conditionally mounts these routes (and maybe in a different folder?)

Custard7: ⚠️ cleanupBenchAppEventData allows any authenticated user to wipe any user's data! tRPC endpoints are public APIs—any caller with an app-store:write token can hit /app-store/cleanup-bench-data directly with any recipientProfileId and DETACH DELETE that user's credentials, notifications, and credential-activity entries. Same risk class on benchAppEvent (no recipientProfileId destruction, but a 100-iteration credential-issuance loop is a DoS amplifier).

services/learn-card-network/brain-service/vitest.unit.config.ts
```diff
@@ -6,6 +6,10 @@ export default defineConfig({
     test: {
         environment: 'node',
         globals: true,
-        include: ['test/uri-helpers.spec.ts', 'test/oidc-jwt.spec.ts'],
+        include: [
+            'test/uri-helpers.spec.ts',
+            'test/oidc-jwt.spec.ts',
+            'src/helpers/posthog.helpers.test.ts',
+        ],
```
Custard7, line 15:
```suggestion
            'src/helpers/posthog.helpers.test.ts',
             'percentile.helpers.test.ts',
             'perf.test.ts'
        ],
```

apps/learn-card-app/src/pages/launchPad/CredentialClaimModal.tsx
```diff
@@ -29,69 +39,171 @@ export const CredentialClaimModal: React.FC<CredentialClaimModalProps> = ({
...
+
         const resolveCredential = async () => {
             try {
                 setIsLoading(true);
                 setError(null);
+
+                if (preResolvedCredential) {
+                    // Fast path — render immediately with what we have
+                    if (!cancelled) {
+                        setCredential(preResolvedCredential);
```
Custard7, line 85: Claude noted this, we should verify which claim is true: 

Fast-path race: walletRef.current may be null when user clicks Accept

In the fast path, setCredential(preResolvedCredential) runs before await initWallet() resolves. The intent is "modal interactive at 0ms after mount" — but isLoading only flips false in the outer finally block after both initWallet() and the background enrichment read.get() complete. So either:
- (a) The fast-path UX doesn't actually deliver on the "0ms interactive" claim — the loading state still blocks the view until enrichment finishes (~200–700ms). The time_to_modal_interactive_ms telemetry over-reports the win because it's measured at setCredential time, not at the loading→interactive transition.
- (b) If you set setIsLoading(false) immediately in the fast path to actually deliver "0ms interactive", the Accept button can fire handleClaim() before walletRef.current is set. The Promise.allSettled then silently degrades: addVCtoWallet runs (via the hook, separate code path), but acceptCredential and queryNotifications become Promise.resolve() no-ops. The user sees "Successfully claimed Credential!" but the server-side accept never happened.

Fix: Either gate the Accept button on a walletReady boolean (not just isClaiming/claimed), or have handleClaim await initWallet() defensively when the ref is null.

services/learn-card-network/brain-service/src/helpers/signingAuthority.helpers.ts
```diff
@@ -10,13 +11,110 @@ import {
...
+            // Prefer HTTP/2 when the server supports it (e.g. staging.api.learncard.app
+            // fronted by API Gateway). H/2 multiplexes streams over a single TCP
+            // connection, further reducing handshake overhead across concurrent SA calls.
+            allowH2: true,
+        });
+        // Install as global dispatcher so native fetch() picks it up. Node's global
+        // fetch only honors a per-call `dispatcher` option when it's on undici's
+        // internal options object; setting the global dispatcher is the reliable path
+        // and also benefits any other SA-adjacent calls that use plain fetch.
+        setGlobalDispatcher(_saAgent);
```
Custard7: setGlobalDispatcher mutates Node process state

This affects every fetch() call in the lambda, not just SA calls. The _saAgent is tuned for SA endpoints (connections: 10, keepAliveTimeout: 30s, HTTP/2). Any other outbound HTTP from brain-service now shares this pool and these timeouts. We should document this, or use the per-call dispatcher option on SA fetches only.

services/learn-card-network/brain-service/src/cache/in-memory-lru.ts
```diff
@@ -0,0 +1,71 @@
...
+
+        const entry = items[index]!;
+        if (entry.expiresAt <= now) {
+            // Expired — drop and miss.
+            items.splice(index, 1);
+            return undefined;
+        }
+
+        entry.timestamp = now;
+        items.sort((a, b) => b.timestamp - a.timestamp);
```
Custard7: Another claude recommendation:

LRU cache: O(n log n) sort on every hit

Called inside get(). At limit=200 on a hot Boost/AppStoreListing cache this is ~1500 comparisons per hit. The whole point of the cache is to win microseconds. Use a Map + linked list, or just move the touched entry to index 0 manually (O(n) with a splice — already done for add, do the same for get).

services/learn-card-network/brain-service/src/helpers/signingAuthority.helpers.ts
```diff
@@ -46,143 +146,231 @@ export async function issueCredentialWithSigningAuthority(
...
+                    });
+                    await sleep(backoff);
+                }
             }
-        },
-        {
-            issuer: getIssuerProfileId(issuer),
-            saEndpoint: signingAuthorityForUser.signingAuthority.endpoint,
+            // Exhausted retries. Re-throw the last error with full context preserved.
+            throw lastError;
```
Custard7: Logic is currently sound (success returns early, failure assigns lastError before re-loop), but defensive: 

```suggestion
            throw lastError ?? new Error('SA request failed without recorded error');
```

packages/plugins/learn-card-network/src/plugin.ts
```diff
@@ -2432,6 +2432,17 @@ export async function getLearnCardNetworkPlugin(
                 return client.appStore.appEvent.mutate({ listingId, event });
             },
 
+            benchAppEvent: async (_learnCard, input) => {
+                await ensureUser();
+                // tRPC output is z.record(), cast to known shape
+                return client.appStore.benchAppEvent.mutate(input) as any;
```
Custard7: Replace `as any` with typed output schema

internal-docs/lc-1644-summary.md
Custard7, line 1: @goblincore - you know, so we should discuss as a team where we want to put these internal working docs, because I actually do think they are quite helpful - when reviewing PRs, and for continuing to build on each other's work. But! *anything* in the `docs` folder gets published to gitbook, so we should standardize where we put these in the repo. Maybe a high-level folder, `internal-docs`? Or a `.codex`?

Custard7 (commented):
@goblincore this is excellent work! 

The major item to address it the benchmarking routes - these need to be gated so they don't mount in prod because they are fairly sensitive/destructive. 

And e2e tests are passing for me: ✅

Image

internal-docs/lc-1644-summary.md
Custard7: Also, I've noticed our agent friends have started writing quite a few inline comments. They are helpful for review, but I wonder if we should think about whether we want to trim them down a bit. Doesn't need to happen in this PR in particular, but curious what your thoughts are

goblincore: Cool! thanks for taking a closer look, will take a pass at resolving the above comments and observations. Yeah we can trim down the inline comments, probably just need a nudge in AGENTS or CLAUDE.md to tell agents not to leave very verbose comments. As for internal documentation - thats a good question - i think a high level separate folder like internal docs is good. @Custard7

goblincore:
@Custard7 thanks for the review — pass complete, one commit per theme, summary below. Ready for re-review whenever you have time.

## Addressed

### Bench-route gating (security) — [`4d5055899`](https://github.com/learningeconomy/LearnCard/pull/1225/commits/4d50558990519cae0eaa3a36c08b8b0c2a6db86a)
> **>** _These should be gated like how our test router works..._
> **>** _⚠️ cleanupBenchAppEventData allows any authenticated user to wipe any user's data!_

- Moved `benchAppEvent` + `cleanupBenchAppEventData` out of `app-store.ts` into a dedicated `src/routes/bench.ts` router.
- Mounted via `bench: process.env.ENABLE_BENCH_ROUTES ? benchRouter : undefined` in `app.ts` — mirrors the `IS_E2E_TEST` test-router pattern. Production never exposes these routes.
- Renamed paths to `bench.appEvent` / `bench.cleanupAppEventData`.
- **Defense in depth on cleanup**: even with `ENABLE_BENCH_ROUTES=true`, `cleanupAppEventData` now requires `recipientProfileId === ctx.user.profile.profileId` and throws `FORBIDDEN` otherwise — no more "wipe arbitrary user's data" footgun.

### Bench methods removed from SDK plugin — same commit
> **>** _Should these be added to this plugin?_
> **>** _Replace `as any` with typed output schema_

- Dropped `benchAppEvent` / `cleanupBenchAppEventData` from `packages/plugins/learn-card-network/src/{plugin.ts,types.ts}` — the `as any` cast goes with them.
- Bench panel now talks to tRPC directly via `wallet.invoke.getLCNClient().bench?.appEvent.mutate(...)`. When the server-side env var is off, the SDK type is `undefined` and the panel surfaces "Bench routes not enabled on this environment. Set `ENABLE_BENCH_ROUTES=true` on the server."

### Cache layer extracted — [`5ec45de2b`](https://github.com/learningeconomy/LearnCard/pull/1225/commits/5ec45de2b)
> **>** _these should be moved to a separate cache layer file/folder_

- Six LRU caches + cached-getter helpers (`getBoostForListingCached`, `readAppStoreListingByIdCached`, `getIntegrationForListingCached`, `getOwnerProfileForIntegrationCached`, `getPrimarySigningAuthorityFor{Listing,Integration}Cached`) lifted out of `app-store.ts` into `src/cache/app-store.caches.ts`. No behavior change.

### LRU `get()` complexity — [`a32fda861`](https://github.com/learningeconomy/LearnCard/pull/1225/commits/a32fda861)
> **>** _LRU cache: O(n log n) sort on every hit ... do the same for get._

- Replaced the `items.sort(...)` on every `get()` hit with splice+unshift (the same pattern `add()` already used). At `limit=200` on a hot SA path that's ~1500 comparisons saved per cache hit.

### SA `setGlobalDispatcher` + retry throw — [`3b5a4f24e`](https://github.com/learningeconomy/LearnCard/pull/1225/commits/3b5a4f24e)
> **>** _setGlobalDispatcher mutates Node process state ... use the per-call dispatcher option on SA fetches only._
> **>** _Logic is currently sound (success returns early...), but defensive: `throw lastError ?? new Error('SA request failed without recorded error');`_

- Dropped `setGlobalDispatcher(_saAgent)` — the SA fetch sites already pass `dispatcher: getSaAgent()` per-call, which is sufficient. Federation / notifications / oidc-jwt etc. now keep Node's default fetch pool.
- Updated the surrounding comment to reflect per-call scoping.
- Applied the defensive `throw lastError ?? new Error(...)` suggestion verbatim.

### CredentialClaimModal fast-path race — [`6a9aeb843`](https://github.com/learningeconomy/LearnCard/pull/1225/commits/6a9aeb843)
> **>** _Fast-path race: walletRef.current may be null when user clicks Accept ... Fix: Either gate the Accept button on a walletReady boolean ... or have handleClaim await initWallet() defensively when the ref is null._

Both, belt-and-braces:
- New `walletReady` state flips true only after `initWallet()` resolves (fast and slow paths).
- Accept button is `disabled={isClaiming || claimed || !walletReady}` and shows a "Preparing…" spinner while `walletReady` is false.
- `handleClaim` also re-awaits `initWallet()` defensively if `walletRef.current` is somehow still null.

### Internal docs moved out of `docs/` — [`642fea8c3`](https://github.com/learningeconomy/LearnCard/pull/1225/commits/642fea8c3)
> **>** _anything in the `docs` folder gets published to gitbook..._

- `docs/internal/lc-1644-summary.md` → `internal-docs/lc-1644-summary.md` (new top-level folder).
- This PR establishes the convention; happy to move other internal working docs the same way in a separate sweep — agreed `internal-docs/` is a reasonable home, but flag if you'd prefer `.codex/` or something else.

### `tenant-config.ts` reverted — [`cc157cc49`](https://github.com/learningeconomy/LearnCard/pull/1225/commits/cc157cc49)
> **>** _Is this supposed to be here still?_

- Reverted the localhost short-circuit on this branch — `apps/learn-card-app/netlify/edge-functions/tenant-config.ts` now has no net change from `main`. The local-dev concern that originally motivated the change (baked `/tenant-config.json` getting clobbered by the edge function overlay) will be handled separately if it re-surfaces.

### vitest config nit — [`c7e42159f`](https://github.com/learningeconomy/LearnCard/pull/1225/commits/c7e42159f)
- Applied your `vitest.unit.config.ts` suggestion — `percentile.helpers.test.ts` and `perf.test.ts` are now included in the unit suite.

### Branch updated against `main` — [`e7d38e63a`](https://github.com/learningeconomy/LearnCard/pull/1225/commits/e7d38e63a)
- Merged latest `origin/main` (bitstring status list work + misc). One conflict in `signingAuthority.helpers.ts` — resolved to keep both main's `appendBitstringStatusListEntries` wrap (credential → `credentialToIssue`) and this branch's perf tracking / retry loop / per-call dispatcher. `credentialToIssue` now flows through the request body, subjectId computation, and test mock.

## Discussion items, no code change

- **Inline comment verbosity from agents** — agreed, will add a `keep inline comments terse` nudge to `AGENTS.md` / `CLAUDE.md` in a follow-up so it applies across future PRs.

## Verification
- `nx build network-brain-service` ✓
- `nx build network-plugin` ✓
- `nx build learn-card-app` ✓
- Brain-service unit suite: 76/76 passing (now includes percentile + perf tests).

services/learn-card-network/brain-service/.env.example
```diff
@@ -2,3 +2,9 @@ SEED=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
 NEO4J_URI=neo4j+s://example.com
 NEO4J_USERNAME=neo4j
 NEO4J_PASSWORD=asdfasdfasdfasdfasdfasdfasdfasdfasdf
+
+# LC-1644 perf bench (admin-only). Optional — if unset, the bench still runs
+# but emits no PostHog events.
+POSTHOG_API_KEY=
+POSTHOG_HOST=https://us.i.posthog.com
+GIT_SHA=
```
Custard7, line 10:
```suggestion
GIT_SHA=

# Set to 'true' to mount /bench/* perf routes (dev/staging only)
ENABLE_BENCH_ROUTES=
```

Custard7 (approved):
@goblincore excellent, thank you for tackling all review comments. This is really solid, nice work! 👍 ✅  

E2E still passing for me, too:

Image

We should verify critical routes on prod when this goes out (i.e. claiming a cred in apps) since this touches high value areas. Excited for the perf gains here!

Custard7:
Although, @goblincore these tests need to be addressed, minor thing based on the change in this PR to the logic for the IS_OFFLINE:

Screenshot 2026-05-13 at 9 11 32 AM

goblincore:
> Although, @goblincore these tests need to be addressed, minor thing based on the change in this PR to the logic for the IS_OFFLINE:

@Custard7 oh cool ill take a look shortly at the failing tests and address it

goblincore:
@Custard7 everything green now! gonna merge

PR #1234, opened 2026-05-13
ci: Packages Smoketest on publish + nightly

Custard7:
## Why

`@learncard/didkit-plugin-node@0.2.14` shipped to npm with literal `"workspace:*"` strings in its `dependencies` field. The moment that version hit the registry, every `npx @learncard/cli` invocation in the wild started failing with `EUNSUPPORTEDPROTOCOL`. **Internal CI didn't catch it** because everything we test runs inside the monorepo, where pnpm resolves `workspace:*` correctly; the bug only appears when you install the published artifact from outside the workspace.

This PR adds a smoketest canary for published packages.

## What this does

Adds `.github/workflows/smoketest-npm-packages.yml`, a reusable workflow that can be triggered via workflow, nightly cron, and workflow_dispatch

For each matrix cell, it:

1. Sets up Node and the package manager
2. Creates a clean directory **outside the monorepo** (`mktemp -d`) with a minimal `package.json`
3. Installs `@learncard/cli`, `@learncard/init`, `@learncard/partner-connect` from npm at the requested versions (defaults to `latest`)
4. Verifies `@learncard/init` loads via both ESM (`import { initLearnCard }`) and CJS (`require('@learncard/init')`) and exports a function — `@learncard/init` ships both formats and we want to catch dual-package hazards
5. Same dual-format check for `@learncard/partner-connect` (verifies `createPartnerConnect` is exported)
6. Verifies `./node_modules/.bin/cli` exists (sanity-checks `@learncard/cli`'s `bin` wiring without actually invoking the interactive REPL)
7. Reports the installed versions for log breadcrumbs

## Matrix

| package-manager | node-version | rationale |
|---|---|---|
| `npm` | `18` | min currently-supported LTS; **this is the package manager that broke us** |
| `npm` | `20` | active LTS |
| `npm` | `22` | latest LTS; catches forward-compat issues |
| `pnpm` | `20` | many LearnCard partners use pnpm; different resolution algorithm catches different bug classes |
| `yarn` | `20` | other major package manager |

## Wiring into `deploy.yml`

New job between `publish-npm` and the existing per-environment `smoketest-*` jobs:

```yaml
smoketest-published-packages:
    name: Smoketest Published NPM Packages
    needs: publish-npm
    if: |
        always() &&
        needs.publish-npm.result == 'success'
    continue-on-error: true
    uses: ./.github/workflows/smoketest-npm-packages.yml
```

**`continue-on-error: true`** makes it advisory.


Custard7:
@claude review

TaylorBeeston (approved):
This is awesome! =)

PR #1239, opened 2026-05-14
feat: Use Claim Link UX

Custard7:
Lets users on devices without a camera (desktop browsers, PWA mobile web, broken-camera native) claim credentials by **pasting the link** or **uploading a QR code image** instead of scanning. All input sources — camera scan, paste, image upload, clipboard auto-detect — share a single routing engine, so they accept the same URL formats with the same precedence rules.

## Why
Right now the "Claim Credential" menu item is gated by `Capacitor.isNativePlatform()` (`AddToLearnCardMenu.tsx#L179-L187`). Desktop and PWA users see no path to claim an offer. Specific scenarios this blocks today:
- User gets an `openid-credential-offer://…` link in an email on their desktop. Custom schemes don't resolve in desktop browsers → dead end.
- User sees a QR on a website they're already viewing on desktop. Can't scan their own screen → dead end.
- User on PWA mobile web with no camera permission, or a phone with a broken camera → dead end.
- Issuer sends a link via SMS that the user copies but doesn't tap → dead end.
A wallet that only accepts credentials through a camera is a wallet that excludes a meaningful fraction of users from a meaningful fraction of issuers. This PR fixes that.
## What
Two surfaces, one engine, telemetry baked in:

### Menu addition
- **"Use a Claim Link"** — visible on every platform (native + web).
- **"Claim Credential" → renamed "Scan a QR Code"** — still native-only, unchanged behavior.

### Modal
A single `<PasteOrUploadClaimModal>` with stacked (not tabbed) affordances:
- Paste field with auto-prefill from clipboard if the clipboard content matches a known claim scheme (privacy-safe — silent on permission denial, only triggers on recognized prefixes).
- File picker + drag-and-drop image upload, decoded client-side via `qr-scanner`.
- Inline friendly error copy keyed off `UnrecognizedReason` ("That doesn't look like a claim link. Try copying the whole link, starting with `https://`, `openid-credential-offer://`, or similar.").
- Loading spinner + disabled Continue button while processing.
### Engine
The format-detection logic that used to live inside `QRCodeScannerListener.handleScan` is now a pure function (`parseClaimInput`) + a React dispatch hook (`useClaimInputRouter`). The camera scanner and the new modal both consume the same router — change the parser once, every input path benefits.

### Supported input formats
The router accepts the same set of formats the QR scanner already handled — paste/upload doesn't expand the surface, it just reuses it:
Input	parsed_kind
openid-credential-offer://…	oid4vci
openid4vp://…	oid4vp
dccrequest://, msprequest://, asuprequest://	vc-api-custom-scheme
https://lcw.app/…	lcw-https
?boostUri=&challenge= (URL or bare query)	boost-claim
?iuv=1 interaction URL	interaction-url
?did=did:web:…:users:<id>	connection-request
Raw VC JSON (with @context + type: VerifiableCredential)	raw-vc-candidate
Precedence is documented in code and pinned by tests — changing the order changes user-visible behavior, and the test suite makes that visible in PR diffs.

## UX walkthrough
Desktop user gets an email with an openid-credential-offer://… link

Before this PR: clicking the link in a desktop browser does nothing (no custom-scheme handler). Dead end.

After this PR:
1. User copies the link from the email.
2. Opens LearnCard in their browser, taps "+" → "Use a Claim Link".
3. Modal opens; clipboard auto-detected → paste field already filled.
4. Taps Continue. Spinner → routed to /oid4vci?offer=… → consent screen → credential lands.

Mobile PWA user with no camera permission
- Before this PR: "Claim Credential" hidden entirely. Dead end.
- After this PR: "Use a Claim Link" is visible. Paste flow works identically.

User has a screenshot of a QR code on their desktop

Before this PR: dead end.
After this PR: drag the image onto the modal's drop zone → qr-scanner decodes client-side → routed.

## Reviewer checklist
- [ ] On desktop with no camera: open menu → "Use a Claim Link" → paste an openid-credential-offer://… URL → consent screen renders.
- [ ] On desktop with no camera: drag a QR image onto the drop zone → routed correctly.
- [ ] Clipboard contains an OID4VCI offer when the modal opens → paste field auto-prefilled.
- [ ] Clipboard contains plain text ("hello world") → paste field NOT prefilled (privacy guard).
- [ ] Paste an unknown URL (https://random-website.com/page) → friendly error inline, modal stays open.
- [ ] Paste malformed JSON → friendly error, modal stays open.
- [ ] On native iOS/Android: existing QR scanner still works identically (regression check on the refactor).
- [ ] On native: both "Scan a QR Code" AND "Use a Claim Link" are in the menu (paste is a complement, not a replacement, for users with broken cameras or in-app browsers that don't honor custom schemes).
- [ ] Analytics: claim_input_routed event fires once per route call with the expected payload shape.
- [ ] qr-scanner adds ~60 KB to the bundle — acceptable for the UX gain. (Could be lazy-loaded if size becomes a concern; not worth it today.

---


Custard7:
@claude please do a deep-review

Custard7:
@claude please re-deep-review

Custard7:
@claude please re-deep-review and recommend merge

PR #1250, opened 2026-05-20
Pt 1 — feat: add CredentialFormat types

Custard7:
Adds TypeScript types and pure helpers for tagging credentials by their format (W3C VC, SD-JWT-VC, JWT-VC, mDoc).

Nothing in the app or SDK consumes these yet — that happens in #1258. Existing W3C VC code is unaffected.

Background: `internal-docs/adr/0001-format-tagged-credential-storage.md` explains why we need this. Short version: we're about to add SD-JWT-VC (and later mDoc), and a JSON wrapper can't honestly represent either. The types lock in the shape we'll standardize on.

### What's in here

- `CredentialFormat` union: `'w3c-vc-2.0' | 'w3c-vc-1.1' | 'jwt-vc-json' | 'dc+sd-jwt' | 'vc+sd-jwt' | 'mso_mdoc'`
- `StoredCredential` discriminated union with the payload type narrowed by `format`
- `toStoredCredential(record)` — projects a `CredentialRecord` into the right format-tagged read view
- Legacy shape inference inside `toStoredCredential` for W3C VC versions, compact JWT/SD-JWT strings, and JWT-bearing proof wrappers
- Public exports from `@learncard/types` and `@learncard/helpers`

`mso_mdoc` is in the union even though we don't decode it yet — including it now means the rest of the SD-JWT-VC stack already accepts it, so when mDoc lands we don't have to widen signatures everywhere.

### How to test

```bash
pnpm --filter @learncard/helpers test -- credential-format.test.ts --runInBand
pnpm --filter @learncard/types build
pnpm --filter @learncard/helpers build
```

Jira: LC-1796.

### Stack

Base of a 3-PR stack: this → #1258 → #1264.

### ✨ PR Description
Purpose: Introduce format-tagged credential storage types and a migration projector for ADR-0001 Phase 1, enabling format discrimination and backward compatibility for diverse credential formats.

Main changes:
- Added `CredentialFormat` enum and `StoredCredential` discriminated union with format-aware wire-form handling for W3C VCs, JWT-VC, SD-JWT-VC, and mDoc
- Implemented `toStoredCredential()` projector supporting explicit format metadata and legacy shape inference with graceful fallback
- Extended `CredentialRecord` with optional `format`, `semanticType`, and `rawWireForm` fields for format-aware writers while maintaining backward compatibility

_Generated by LinearB AI and added by gitStream._
<sub>AI-generated content may contain inaccuracies. Please verify before using.
[...]

PR #1246, opened 2026-05-18
feat: Wire SD-JWT-VC into OID4VC

Custard7:
> **Stack position:** This PR is the second in a four-PR stack.
> 1. [#1250](https://github.com/learningeconomy/LearnCard/pull/1250) — `adr-0001-phase-1-format-tagged-types` → `main` (format-tagged credential types)
> 2. **this PR** — `lc-1796-2` → #1250
> 3. [#1258](https://github.com/learningeconomy/LearnCard/pull/1258) — `lc-1796-3` → this PR (credential envelope + mixed-format presentation)
> 4. [#1264](https://github.com/learningeconomy/LearnCard/pull/1264) — `lc-1796-4` → #1258 (resilience + display polish)
>
> Merge order: #1250 → **this PR** → #1258 → #1264.

#### 🎟 Relevant Jira Issues

LC-1796

---

A LearnCard initialized via any seed-based `initLearnCard(...)` call can now claim an SD-JWT-VC from any OID4VCI issuer. The credential lands in the wallet's index under a sensible category and renders through the existing credential card with **no app-side changes**.

This is the second-to-last slice of the SD-JWT-VC epic. This PR wires both into the OID4VCI receipt pipeline and bundles the plugin so apps pick it up on upgrade.

## How it fits together

When an OID4VCI issuer returns a `dc+sd-jwt` (or legacy `vc+sd-jwt`) credential, the openid4vc plugin now:

1. Detects the format and **delegates parse to the sd-jwt-vc plugin via `learnCard.invoke.parseSdJwtVc()` at runtime** — no hard dependency. If sd-jwt-vc isn't installed, the credential is recorded in `result.failures` with a clear install hint instead of crashing.
2. Synthesizes a W3C-VC-shaped object so the existing store/index/read pipelines work without per-format branches:
    - `type: ['VerifiableCredential', 'SdJwtVcCredential']`
    - Original compact SD-JWT preserved under `proof.jwt` (`proof.type: 'SdJwtCompactProof'`)
    - `vct` exposed under a `sdJwtVct` extension field
3. **Categorizes via `learnCard.invoke.categorizeSdJwtVct(vct)`** when the SD-JWT marker is present; falls back to the existing W3C-`type[]` heuristic otherwise.

The sd-jwt-vc plugin is **auto-wired into all 4 seed-based `@learncard/init` initializers**, between the existing LearnCard plugin and openid4vc. Existing apps upgrade with zero code changes.

## Architectural notes
[...]

PR #1258, opened 2026-05-21
Pt 2 — feat: SD-JWT-VC support across the wallet

Custard7:
Adds SD-JWT-VC support across the wallet: receive over OID4VCI, store in native format, present over OID4VP with per-claim user consent.

### Why

Partners (California Career Passport, EUDI Wallet) need SD-JWT-VC. ADR-0001 commits us to storing credentials in their native format from day one — no transitional W3C wrapper, no migration cycle later. This PR is where that policy gets implemented.

### What's in here

**Receive (OID4VCI)**

The openid4vc plugin detects `dc+sd-jwt` / `vc+sd-jwt` credentials and delegates parsing to the sd-jwt-vc plugin via `learnCard.invoke.parseSdJwtVc()`. Runtime feature-detect, no hard import — apps that don't want SD-JWT can omit the plugin.

The sd-jwt-vc plugin is auto-wired into all four `initLearnCard()` initializers, so existing apps get support on upgrade with zero code changes.

**Store**

SD-JWT-VC credentials persist as `{ format, data }` envelopes. Read-time projection at the storage plugin boundary returns a W3C-shape VC, so all 32 existing `wallet.read.get(...)` callsites work unchanged.

Brain-service and learn-cloud-service Zod inputs accept the envelope on the wire.

**Present (OID4VP)**

`presentSdJwtVc()` builds the holder-bound compact `<JWT>~<disclosures>~<KB-JWT>?` presentation. KB-JWT only when the source credential has a `cnf` binding. Wired into both DCQL and PEX flows.

The consent screen in the wallet shows one checkbox per disclosable claim — users pick what to share. Required claims (issuer, vct, expiration) are shown read-only.

**Resilient signer fallback**

The wallet tries `did:web` first and falls back to `did:key` if the issuer can't resolve the holder's public DID. Works against walt.id and Sphereon as-is.

> **Heads up:** this fallback has three bugs against EUDI's strict OID4VCI 1.0 Final endpoints. Those are fixed in #1264. Merge them back-to-back, no pause.

**Embedded playground + e2e**

`examples/openid4vc-playground/server/embedded.ts` runs an in-process issuer+verifier so we don't depend on external test servers during development. `tests/e2e/tests/envelope.spec.ts` exercises the full claim → store → present round trip.

### How to test

```bash
pnpm install
pnpm --filter @learncard/sd-jwt-vc-plugin test     # 82/82
pnpm --filter @learncard/openid4vc-plugin test     # 524/524
pnpm --filter @learncard/helpers test              # 75/75
```

**Receiving a credential (walt.id):**

1. Go to https://portal.walt.id → Issuer → SD-JWT-VC profile → Create Offer.
2. Copy the `openid-credential-offer://...` URI.
3. `pnpm nx start cli --skip-nx-build`
   ```ts
   const lc = await initLearnCard({ seed: 'a'.repeat(64) });
   const result = await lc.invoke.acceptAndStoreCredentialOffer('openid-credential-offer://...');
   console.log(result.stored[0].format);   // → "dc+sd-jwt"
   ```
4. Confirm the credential is in `lc.index.LearnCloud.get(...)` under the right category.

**Presenting with consent (embedded playground):**

1. Start the embedded openid4vc-playground.
2. Run the SD-JWT-VC presentation scenario.
3. On the consent screen, uncheck a few claims.
4. Verify the verifier only sees the checked claims.

### What to focus on in review

- `packages/plugins/openid4vc/src/vp/sign.ts` + `dcql/respond.ts` — SD-JWT presenter wiring. Trace one request from `presentCredentials` down to `presentSdJwtVc`.
- `packages/plugins/learn-cloud/src/plugin.ts` + `packages/plugins/learn-card-network/src/plugin.ts` — read-time projection. Small but load-bearing; if it's wrong, every existing consumer breaks.
- `apps/learn-card-app/src/pages/oid4vp/components/RequestConsent.tsx` — consent UI. Native checkboxes (not IonCheckbox), grayscale tokens per AGENTS.md.
- `tests/e2e/tests/envelope.spec.ts` — the full flow in one file.

### Known gaps

- Mixed-format PEX submissions are rejected with `unknown_credential_format`. DCQL handles mixed formats; no verifier we know of needs PEX-mixed today.
- Status-list checks run at receipt and on `verifyCredential`, not at presentation time.
- The envelope binary path is pre-wired for mDoc but the decoder isn't here yet.
- EUDI interop needs #1264 in the same merge.

Jira: LC-1796.

### Stack

`lc-1796-3` → #1250 → `main`. Children: #1264 (must merge together). #1246 was closed and absorbed here so the transitional W3C wrapper never reaches `main`.


Custard7:
@claude please do a deep-review and recommend merging

Custard7:
@claude comments addressed- please do a final review

Custard7:
@claude please review and recommend merge

packages/learn-card-core/src/types/planes.ts
```diff
@@ -48,7 +54,10 @@ export type GetPlaneProviders<
 // --- Read --- \\
 
 export type ReadPlane = {
-    get: (uri?: string, options?: PlaneOptions) => Promise<VC | VP | undefined>;
+    get: (
+        uri?: string,
+        options?: PlaneOptions
+    ) => Promise<VC | VP | StoredCredentialEnvelope | undefined>;
```
TaylorBeeston, line 60: This is perfect! Exactly how I would do it =)

packages/learn-card-helpers/src/credential-format.ts
```diff
@@ -1,4 +1,40 @@
...
+    'iss',
+    'iat',
+    'exp',
+    'nbf',
+    'sub',
+    'vct',
+    'cnf',
+    '_sd_alg',
+    '_sd',
+    '...',
```
TaylorBeeston, line 21: I thought this was a mistake and had omp look it up! Nope, that is definitely a correct reserved claim! 😂

https://www.rfc-editor.org/rfc/rfc9901.html#section-4.1-3.7

packages/learn-card-types/src/credential-format.ts
```diff
@@ -27,6 +26,64 @@ export const CredentialFormatValidator = z.enum([
...
+ * Storage-plane envelope for native (non-W3C) credential formats.
+ *
+ * The storage plane previously accepted only `VC | VP`, which forced
+ * SD-JWT-VC to synthesize a JSON-LD wrapper around the compact form
+ * just to satisfy the type — even though the underlying transport
+ * (LearnCloud tRPC) accepts any JSON-serializable value.
+ *
+ * The envelope is the on-wire shape for any credential that is NOT a
+ * W3C VC/VP. Consumers identify the envelope by the presence of both
+ * `format` and `data` fields (use `isStoredCredentialEnvelope`).
```
TaylorBeeston, line 39: This is _maybe_ TMI here, but could also be valid/useful for agents that wanted to look up this type and understand it better, so I think it's _okay_ to leave it in for now, but as this becomes the standard, bespoke format, it may feel kind of weird to leave this comment in here as its referencing like a specific point in history.

packages/learn-card-types/src/credential-format.ts
```diff
@@ -27,6 +26,64 @@ export const CredentialFormatValidator = z.enum([
...
+ */
+export const isStoredCredentialEnvelope = (
+    value: unknown
+): value is StoredCredentialEnvelope => {
+    if (!value || typeof value !== 'object') return false;
+    const candidate = value as Record<string, unknown>;
+    if (typeof candidate.format !== 'string') return false;
+    if (!CredentialFormatValidator.safeParse(candidate.format).success) return false;
+    return typeof candidate.data === 'string' || candidate.data instanceof Uint8Array;
+};
```
TaylorBeeston, line 85: ❤️ Either here or in learncard/helpers, but these kinds of type guards are _so_ helpful and it's awesome that this is showing up exposed at this level =)

packages/learn-card-types/src/credential-format.ts
TaylorBeeston, line 156: This is awesome! There are places in learn-card-base (I believe it's credentialHelpers) that sort of do something similar (attempt to get/define a standardized set of display information from an arbitrary credential). We should potentially discuss this in like a design meeting and nail down _exactly_ what kind of standardized information we want to display, like what's option, etc, then create rigorous, standardized helpers for getting this info! (I _believe_ this is already being done by `toDisplayViewModel` referenced above!)

packages/learn-card-types/src/learncard.ts
```diff
@@ -90,23 +90,20 @@ export type CredentialRecord<Metadata extends Record<string, any> = Record<never
...
      */
     semanticType?: string;
     /**
      * On-the-wire representation for formats whose `vc` field cannot
      * hold the canonical form (SD-JWT compact, JWT-VC compact, base64
-     * mDoc CBOR). Undefined for W3C VCs (`vc` IS the wire form). Added
-     * in ADR-0001 Phase 1.
+     * mDoc CBOR). Undefined for W3C VCs (`vc` IS the wire form).
      */
     rawWireForm?: string;
```
TaylorBeeston, line 111: So I recoiled a bit at first seeing these forcibly added on the index record, but thinking about it more it makes sense. Omp said that we aren't efficiently using these in this branch though when doing the OIDC4VP flow where it can actually request eg the semanticType, so maybe something to consider (omp said we're basically just resolving everything and looking at it that way, though I haven't confirmed). It also seems odd that we need _three_ keys to represent this. What is the real difference between these? Would it be better to collapse them into one?

packages/plugins/openid4vc/src/vci/store.ts
```diff
@@ -14,14 +16,35 @@ import { VciError } from './errors';
...
+     * Format-tagged metadata. Populated by format-aware writers
+     * (currently the SD-JWT-VC path); legacy W3C paths leave these
+     * fields undefined. Format-aware readers filter by these fields
+     * without needing to read the credential body from storage.
+     */
+    format?: import('@learncard/types').CredentialFormat;
+    semanticType?: string;
     /** Index record schema version, matches existing wallet writes. */
     __v: 1;
 }
```
TaylorBeeston, line 50: nit (existing): this should just use the type from learncard types

packages/plugins/sd-jwt-vc/src/plugin.ts
```diff
@@ -17,5 +54,28 @@ export const getSdJwtVcPlugin = (learnCard: SdJwtVcDependentLearnCard): SdJwtVcP
...
+            const proof = pickProof(credential);
+            if (
+                proof?.type === SD_JWT_PROOF_TYPE &&
+                typeof proof.jwt === 'string' &&
+                proof.jwt.length > 0
+            ) {
+                return verifySdJwtVc(learnCard, proof.jwt, {});
+            }
+            return learnCard.invoke.verifyCredential(credential, options);
+        },
```
TaylorBeeston, line 79: ❤️ the beauty of the plugin system shining through! This is my favorite part of how learncard SDK works

packages/plugins/sd-jwt-vc/src/plugin.ts
```diff
@@ -17,5 +54,28 @@ export const getSdJwtVcPlugin = (learnCard: SdJwtVcDependentLearnCard): SdJwtVcP
...
+            }),
+
+        verifyCredential: async (_lc, credential, options) => {
+            const proof = pickProof(credential);
+            if (
+                proof?.type === SD_JWT_PROOF_TYPE &&
+                typeof proof.jwt === 'string' &&
+                proof.jwt.length > 0
+            ) {
+                return verifySdJwtVc(learnCard, proof.jwt, {});
```
TaylorBeeston, line 76: One thing about this is that it outputs like very user-unfriendly checks. We should map those in the learn-card plugin to be more friendly for users! 

Image

TaylorBeeston (approved):
Okay, this LGTM in my testing! Left a bunch of comments. I _should_ mention though that OMP was a little more strict about it, so I'm going to paste its review and when I'm done reviewing PR 3 I can kind of dive deeper into its feedback.

   I would not merge this yet. The architecture is good, but I found a few correctness/security edge cases that should be fixed in this slice.

   ### Blocking

   1. Require SD-JWT verification before storing

   `packages/plugins/openid4vc/src/vci/sd-jwt-vc.ts:122-123`

   `callPluginVerify()` returns `undefined` when `verifySdJwtVc` is missing, and `synthesizeSdJwtVc()` then proceeds to store the credential after only `parseSdJwtVc()` ran.

   Parsing reconstructs claims/disclosures; it does not prove issuer signature validity. That means a LearnCard with a parser but no verifier can accept and persist tampered `dc+sd-jwt` / `vc+sd-jwt` credentials as native records.

   This should fail closed: if the SD-JWT plugin can parse but cannot verify, receipt should throw `unsupported_format` rather than store.

   2. Switching SD-JWT candidates can submit a disclosure frame that does not match the UI

   `apps/learn-card-app/src/pages/oid4vp/components/RequestConsent.tsx:183-184`

   `picks.disclose` is keyed only by row id. Once the first selected candidate initializes `picks.disclose[rowId]`, parsing a later candidate returns early:

   ```ts
   if (prev.disclose[rowId]) return prev;
 ```

 For a row with multiple SD-JWT candidates, the UI renders the new candidate’s claim keys as checked via discloseFrame?.[key] ?? true, but buildChosenList() still submits the old row-level frame. The presenter treats absent keys as hidden, so the UI can say a claim is selected while the actual presentation omits it.

 Disclosure state needs to be candidate-aware, or reset/reinitialize when rowId::candidateId changes.

 3. Cancelled parses can permanently disable Share for a candidate

 apps/learn-card-app/src/pages/oid4vp/components/RequestConsent.tsx:208-215

 If a user switches away while a parse is in flight, the cleanup sets cancelled = true, and the final cleanup that removes parsingSdJwts[cacheKey] is skipped. If the user switches back, that cache key is still marked parsing, the effect refuses to enqueue another parse, and canShare remains false indefinitely.

 The parse flag should be cleared for newParsing keys even when the effect was cancelled, or parsing state should be derived from an effect-generation token rather than stored permanently by cache key.

 4. Malformed cnf.jwk can produce an unbound presentation

 packages/plugins/sd-jwt-vc/src/present.ts:215-221

 This accepts any object containing a jwk property, even when jwk is null or a scalar. parseSdJwtVc() only sets holderPublicKey when cnf.jwk is object-valued, so a payload like cnf: { jwk: null } passes the cnf check, sets needsKeyBinding = false, and presents without a KB-JWT despite declaring key binding metadata.

 If cnf.jwk is present, it should be validated as a non-null object or rejected with unsupported_cnf_confirmation_type.

 5. Multiple new SD-JWT paths only inspect the first proof object

 There are several copies of the same bug Part 1 just fixed: array-valued proof picks the first object instead of the supported SD-JWT proof.

 Examples:
 - packages/plugins/sd-jwt-vc/src/plugin.ts:35-37
 - packages/plugins/openid4vc/src/vp/present.ts:400-402
 - packages/plugins/openid4vc/src/vp/select.ts:302-305
 - packages/plugins/openid4vc/src/vp/select.ts:560-562
 - apps/learn-card-app/src/pages/oid4vp/components/RequestConsent.tsx:710-713

 A valid wrapper like:

 ```ts
   proof: [
     { type: 'DataIntegrityProof', ... },
     { type: 'SdJwtCompactProof', jwt: compact }
   ]
 ```

 will fail matching, consent parsing, presentation, or SD-JWT verification depending on which path sees it first. Each extractor should scan for type === 'SdJwtCompactProof' with a non-empty string jwt.

 ### Non-blocking but worth fixing

 packages/learn-card-helpers/src/credential-format.ts:468-473

 projectJwtVcCompactToVc() returns only payload.vc, dropping the compact JWS. The existing normalizeIssuedCredential() path intentionally synthesizes a JwtProof2020 with proof.jwt so read results remain verifiable and VC-validator compatible. If jwt-vc-json envelopes are part of this storage envelope contract, this projector should preserve
 the compact JWT the same way.

TaylorBeeston:
@claude can you please review, considering the review I pasted from oh-my-pi above?

packages/learn-card-types/src/credential-format.ts
Custard7: Yes, I completely agree - that we should do a more robust standardization/normalization for display info. 👍

packages/plugins/sd-jwt-vc/src/plugin.ts
Custard7: The next PR, pt 3 fixes this!

Custard7:
Addressed omp's blockers in 74d4046 (landed on lc-1796-4, which merges back into this branch).

- **B1 (store unverified SD-JWT):** receipt now fails closed — if `verifySdJwtVc` is missing, it throws `unsupported_format` instead of storing after a parse-only pass.
- **B2 (disclosure frame mismatch):** the frame is now keyed by row + candidate, on both the consent side and `buildChosenList`, so switching candidates can't submit the wrong frame.
- **B3 (cancelled parse disables Share):** the parse flag is cleared even when the effect is cancelled, so switching back re-enqueues and `canShare` recovers.
- **B4 (malformed cnf.jwk):** a present-but-null/non-object `cnf.jwk` is now rejected instead of producing an unbound presentation.
- **B5 (proof[0] across SD-JWT paths):** every extractor now scans for `SdJwtCompactProof` with a string `jwt`, so an LDP-first proof array still matches.
- **NB (projectJwtVcCompactToVc):** preserves the compact JWS under a `JwtProof2020` proof so the projected VC stays verifiable.

Tests added for each. openid4vc 539, sd-jwt-vc 90, helpers 48, base 31 — all green.

PR #1264, opened 2026-05-26
Pt 3 — fix: EUDI issuer interop + SD-JWT-VC claims display

Custard7:
Two unrelated fixes that close out the SD-JWT-VC epic:

1. Make the EUDI reference issuer (https://issuer.eudiw.dev/) actually work end-to-end.
2. Add the "Selectively Disclosable Claims" widget and friendlier verification labels.

> **Must merge in the same train as #1258.** #1258 introduces a signer-fallback feature that has bugs against EUDI; the fixes are here. walt.id and Sphereon work without this PR; EUDI doesn't.

### EUDI fixes

#1258 added a `did:web → did:key` fallback when the issuer can't resolve the holder's DID. Three things break against strict OID4VCI 1.0 Final issuers:

- **`invalid_grant: Code inactive` on retry.** The fallback was re-sending the OAuth code on the second attempt. EUDI (and any RFC 6749-strict issuer) rejects that. Split the exchange into `exchangeAuthCodeForToken` (called once) and `requestCredentialsFromAuthCodeToken` (called per signer attempt). Same split for the pre-auth flow.

- **Can't build a proof JWT.** EUDI omits `c_nonce` from the token response and advertises a `nonce_endpoint` instead (OID4VCI 1.0 Final §7.1). When metadata advertises `nonce_endpoint`, POST to it for a fresh nonce; otherwise fall back to the token-response `c_nonce`.

- **Generic `invalid_credential_request` 400s surfaced unfriendly errors.** EUDI returns this code for proof verification failures (e.g. holder DID not resolvable). Added to the signer-failure dispatch table so it triggers `did:key` fallback instead.

A small CORS proxy (`examples/openid4vc-playground/server/eudi-proxy.ts`) lets the playground hit `issuer.eudiw.dev` directly.

### Display polish

- **SdJwtVcClaimsBox** — new section in BoostDetailsSideBar / SideMenu that shows the disclosed claims as line items with a "Privacy First" badge. Renders for SD-JWT-VCs only; W3C VCs are unaffected.

- **verificationPrettifier** — humanizes raw check codes in the Verifications tab (`disclosure_hash_integrity` → "Selective Disclosure: Claims verified", etc.). Preserves messages that upstream already humanized (so a W3C VC expiring 28 Feb 2023 still shows the date instead of "Does Not Expire").

- **`@learncard/react` hooks** — `useCredentialFormat` and `useParsedSdJwtVc` plus `humanizeClaimLabel` / `formatClaimValue` helpers, for consumer apps that want to build their own SD-JWT-aware widgets.

### How to test

```bash
pnpm --filter @learncard/openid4vc-plugin test    # 34 new tests
pnpm --filter learn-card-base test                 # 23 prettifier tests, 21 decideRecovery
pnpm --filter learn-card-app test                  # resilientVci.test.ts
```

**EUDI claim flow:**

1. Start the openid4vc-playground.
2. Pick the "EUDI Issuer" scenario (proxied through `eudi-proxy.ts`).
3. Claim a credential.
4. It should land without `invalid_grant: Code inactive` or any "we can't read this issuer" prompts. The verifications tab should show all green.

**Display widget:**

1. Claim any SD-JWT-VC (the playground or walt.id will work).
2. Open the credential's details.
3. Confirm a "Selectively Disclosable Claims" section shows up with each disclosed field on its own line, and a "Privacy First" pill in the corner.
4. Open the Verifications tab. Labels should read like English ("Issuer signature is valid"), not like check codes (`issuer_signature`).

**W3C VC regression check:**

1. Open any existing W3C VC in the wallet.
2. Verifications tab should still show expiration with the real date ("Expires 28 FEB 2023" or similar), not a generic "Does Not Expire".

### What to focus on in review

- `vci/auth-code.ts` + `vci/accept.ts` — the split methods. Trace one call from `resilientCompleteCredentialOfferAuthCode` to confirm the auth code is exchanged exactly once across retries.
- `vci/nonce.ts` — fallback chain when `nonce_endpoint` is absent should not double-fetch.
- `verificationPrettifier.ts` — `isMessageRaw` is the heuristic that prevents clobbering humanized upstream messages. There's a regression test for it.

CI: the `Interop (walt.id + Sphereon + EUDI)` job passes (3m19s).

Jira: LC-1796.

### Stack

`lc-1796-4` → #1258 → #1250 → `main`. Must merge back-to-back with #1258.

### ✨ PR Description
Purpose: Enable EUDI issuer interoperability and improve SD-JWT-VC credential display with verification prettification and selective disclosure claim rendering.

Main changes:
- Implement OID4VCI 1.0 Final nonce_endpoint support and split auth-code/pre-auth flows for resilient signer retry without re-consuming single-use grants
- Add SdJwtVcClaimsBox widget displaying disclosed claims with selective disclosure metadata and holder key binding status
- Create verification prettifier translating raw SD-JWT-VC and W3C checks into user-friendly labels; apply to BoostPreview and VerificationsBox displays


Custard7:
@claude please review and recommend merge

packages/learn-card-base/src/helpers/verificationPrettifier.ts
```diff
@@ -0,0 +1,113 @@
...
+            check: label.check,
+            message: isMessageRaw(item.message, item.check) ? label.message : item.message,
+        };
+    }
+
+    return item;
+};
+
+export const prettifyVerificationItems = (items: VerificationItem[]): VerificationItem[] =>
+    items.map(prettifyVerificationItem);
```
TaylorBeeston: This should be moved to the learncard plugin!

TaylorBeeston (approved):
Similar to Part 2, everything here looks good to me! I was able to verify issuance/verification with the playground and with portal.walt.id and everything went smooth! This is really, really awesome work! I left a comment that we really should move from like _another_ prettifying step to putting that logic in the learncard plugin where we already do it, and omp found some issues as well, which I will copy/paste here again!

   I’d hold merge for a few correctness fixes.

   ### Blocking

   1. Token/code exchange happens before signer construction

   `apps/learn-card-app/src/helpers/oid4vc-resilience/resilientVci.ts:81`
   `apps/learn-card-app/src/helpers/oid4vc-resilience/resilientVci.ts:133-137`

   The new split correctly avoids re-exchanging the token across signer fallback, but it now exchanges the pre-auth code / OAuth code before `buildSignerForStrategy()` has succeeded.

   So if local signer setup fails before any `/credential` request is made, the one-time grant is already consumed:

   ```ts
   const tokenResponse = await wallet.invoke.exchangeAuthCodeForToken(...);

   return runWithRecovery(async ({ strategy }) => {
     const signer = await buildSignerForStrategy(wallet, currentSignerId);
     return wallet.invoke.requestCredentialsFromAuthCodeToken({ tokenResponse, signer });
   });
 ```

 Same shape for pre-auth.

 The old convenience plugin path builds the signer before token exchange. This should keep that invariant while still memoizing the token after first successful signer construction:

 ```ts
   let tokenResponse: TokenResponse | undefined;

   return runWithRecovery(async ({ strategy }) => {
     const signer = await buildSignerForStrategy(wallet, currentSignerId);

     tokenResponse ??= await wallet.invoke.exchangeAuthCodeForToken(...);

     return wallet.invoke.requestCredentialsFromAuthCodeToken({ tokenResponse, signer });
   });
 ```

 That preserves EUDI’s single-use code semantics across issuer retry, without burning the grant on local signer setup failures.

 2. Real failed verification items are not prettified

 packages/learn-card-base/src/helpers/verificationPrettifier.ts:81-87

 The tests exercise:

 ```ts
   { check: 'signature_invalid', message: 'signature_invalid: ...' }
 ```

 But the real verifyCredential(..., true) path builds failures from packages/plugins/learn-card/src/verify.ts like:

 ```ts
   {
     status: Failed,
     check: transformErrorCheck(error, ...),
     details: transformErrorMessage(error, ...)
   }
 ```

 For an SD-JWT error like signature_invalid: bad signature, transformErrorCheck returns the whole string as check, message is undefined, and details has the diagnostic. Then isAlreadyPrettified() returns true because the check contains whitespace, so the UI still shows raw machine text.

 The prettifier should extract codes from item.check and item.details too, and it should do that before the “already prettified” whitespace guard.

 3. New SD-JWT display hook only inspects proof[0]

 packages/react-learn-card/src/hooks/useCredentialFormat.tsx:8-10
 apps/learn-card-app/src/components/boost/boostCMS/BoostPreview/SdJwtVcClaimsBox.tsx:39-41

 Both paths choose the first proof object:

 ```ts
   const proofObj = Array.isArray(proof) ? proof[0] : proof;
 ```

 That misses supported wrappers where an LDP/DataIntegrity proof comes first and the SdJwtCompactProof entry comes later. We already fixed this class of bug in the Part 1 helper; these new UI/hook paths should scan the proof array for a supported { type: 'SdJwtCompactProof', jwt: string } entry.

 Otherwise the “Selectively Disclosable Claims” box silently doesn’t render for a valid SD-JWT wrapper.

 ### Non-blocking

 - packages/react-learn-card/src/hooks/useParsedSdJwtVc.tsx:35-47 leaves old data in state while a new compact is parsing, and keeps that stale data if the new parse fails. The app’s current component gates on isLoading/error, but the exported hook should clear data when starting a new parse and on failure.
 - verificationPrettifier.ts maps credentialStatus but the real W3C success path normalizes both status and credentialStatus to check: 'status', so status rows remain lowercase.

TaylorBeeston:
@claude Please review with omp (oh-my-pi)'s review from my last comment in mind

Custard7:
Pushed fixes for everything omp flagged (74d4046).

**Blockers**
- **B1 (code burned before signer):** the grant is now exchanged lazily inside `runWithRecovery`, after `buildSignerForStrategy` succeeds, memoized with `??=`. A local signer failure no longer spends the code. Added a test that asserts the exchange isn't called when the first signer build throws.
- **B2 (real failures not prettified):** the prettifier now reads the code from `check`/`details` (not just `message`), resolves it before the whitespace guard, and preserves humanized `details` like 'Expired 28 FEB 2023' / 'Status: Revoked' instead of overwriting them with a generic label.
- **B3 (proof[0]):** `useCredentialFormat` and `SdJwtVcClaimsBox` now scan the proof array for the `SdJwtCompactProof` entry, so an LDP-first wrapper still renders.

**Non-blocking**
- **NB1:** `useParsedSdJwtVc` clears `data` on new parse and on failure.
- **NB2:** added a `status` alias so W3C status rows get labeled.

Moving the prettifier into the learncard plugin is a fair point — left it as a fast-follow to keep this PR scoped.

Tests: openid4vc 539, sd-jwt-vc 90, learn-card-base prettifier 31, helpers 48, resilientVci 3 — all green.

Custard7:
Moved the prettifier into the learncard plugin (a84fc2c), so the label logic lives next to verify.ts now instead of being a separate pass in learn-card-base.

learn-card-base re-exports it, so useVerifyCredential / VerificationsBox / BoostPreview didn't change. Pure move — no logic or label changes, the 31 tests moved with it and pass.

Custard7:
@claude all comments addressed - please re-review

PR #1266, opened 2026-05-26
feat: [LC-1898] LCA Paraglide i18n core translations

goblincore:
# Overview

#### 🎟 Relevant Jira Issues

https://welibrary.atlassian.net/browse/LC-1898

Adds multi-language (i18n) support across the LearnCard App using **`@inlang/paraglide-js`** (compile-time message generation). Components call typed `m['key']()` functions, compiled from the existing `public/locales/{en,es,fr,ar}/translation.json` files via `@inlang/plugin-i18next`.

Scope is **frontend only** — notifications, emails, and AI-passport translations are separate PRs branched off this one. The whole feature is gated behind a LaunchDarkly flag with per-language gating.

Includes the App Store admin + Developer Portal i18n from #1418, **now merged into this branch**, so this is the single ship-to-main unit for the rollout.

#### 💡 What's included

- **Library**: `@inlang/paraglide-js` — ~2 KB runtime, compile-time inlining, typed message keys. Adapter in `src/i18n/` (`LocaleProvider`, `useLocale`, `useChangeLocale`, RTL handling, `TransP` for inline-markup strings). Generated `src/paraglide/` is gitignored.
- **Coverage** (EN / ES / FR / AR, soft-RTL for Arabic): side menu, wallet/passport, login, onboarding, claim flows, boost CMS & detail, Family CMS, consent flows, AI Insights/Pathways, Dashboard Home Screen, settings, skills frameworks, LearnCard ID editor, App Store admin + Developer Portal, and a broad core-app button sweep.
- **Language-selector gating** via LaunchDarkly (`hideLanguageSelector`): JSON array of language codes to hide (`[]` = show all, `["fr","ar"]` = hide those, `["*"]` = hide selector). Auto-falls a user back to English if their active language gets hidden. Defaults to showing everything when LD is unreachable/malformed. Pure resolver unit-tested (13 cases).
- **`mDynamic`** — typed safe-accessor for dynamic key lookups (replaces all `(m as any)[key]()`; no `any` casts remain); degrades to the key instead of throwing on a missing key.
- **`formatLocaleDate/Time`** — date/time formatting follows the active locale.

#### 🛡️ CI guards

Paraglide is type-safe but nothing in this repo runs `tsc`, so these guards fail the build / a PR check on i18n regressions:

- **Missing keys / missing `m` import** — `check-i18n-keys.mjs` (`I18nKeys` job): diffs every `m['…']` literal against `en/translation.json` and flags files that use `m` without importing it.
- **Verbatim-English values** — `check-i18n-untranslated.mjs`: fails on locale values left in English (the gap key-parity misses).
- **Corrupted interpolation** — `check-i18n-markers.mjs`: fails on broken `{{var}}` / `<n>` markers.
- **Frozen-locale ESLint rules** — `no-frozen-i18n-memo` + `no-module-scope-i18n`: fail on `m['key']()` resolved eagerly in a `useMemo` (no locale dep) or at module scope, which freezes the translation on language switch.
- **Build-time** — rollup `onwarn` turns Paraglide "not exported" warnings into hard build failures.

#### 📸 Feature Breakdown

Screenshot 2026-07-16 at 11 20 44 AM
Screenshot 2026-07-16 at 11 20 16 AM
Screenshot 2026-07-16 at 11 20 22 AM
Screenshot 2026-07-16 at 12 06 25 PM
Screenshot 2026-07-16 at 12 08 14 PM
Screenshot 2026-07-16 at 12 08 07 PM

#### ⚠️ For native-speaker review

Two Arabic strings were machine-retranslated by hand and would benefit from an MSA spot-check: `developerPortal.components.diagnosticsPanel.warningDesc` and `developerPortal.components.reviewStep.minimumAgeDesc`.

#### 🔍 Types of Changes

# Testing

1. Side menu → language picker → switch EN / ES / FR / AR; confirm strings + RTL layout (AR).
2. Verify the login screen footer language picker.
3. **LD flag** (`hideLanguageSelector`): try `[]` (all visible), `["fr"]` (French hidden), `["*"]` (selector hidden on login + side menu); confirm switching to a then-hidden language falls back to English.

#### 📱 🖥 Devices

iOS Native / Android Native / Chromium — native webview parse time + language-switch latency are the key risks.

# ✅ PR Checklist

-   [x] Related to a Jira issue (LC-1831)

[LC-1831]: https://welibrary.atlassian.net/browse/LC-1831?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


goblincore:
@claude please review

goblincore:
@claude please rereview

TaylorBeeston:
Follow-up pushed in `ab69a77fc`.

What changed:
- Removed the temporary Storybook repro-only stories.
- Fixed the AI Insights crash by passing the Paraglide message function into `TransP` instead of passing a resolved string.
- Constrained the dynamic qualitative-factor message keys through a typed key union and replaced the invalid highlight classes with app palette classes (`text-red-700` / `text-emerald-700`).
- Fixed locale bootstrap so `LocaleProvider` initializes Paraglide without a page reload and keeps React state in sync.
- Kept `SharedI18nProvider` mounted around the app route tree so shared package i18n doesn't get torn down by modal/route transitions.
- Localized the missed hardcoded strings in:
  - `AiSessionsLayout`
  - `AiSessionsEmptyState`
  - `BoostDescriptionInput`
  - `FinishedAiSessionAssessmentPreviewDetails`
- Pinned the inlang plugin URL and added the new keys to `en`, `es`, `fr`, and `ar`.

Verification:
- `node scripts/check-i18n-keys.mjs` passed.
- `node scripts/check-i18n-imports.mjs` passed.
- `bunx vitest run src/App.test.tsx src/i18n/languageSelectorConfig.test.ts src/components/boost/NewBoostButton.test.ts src/components/ai-passport/helpers/personalizationI18n.test.ts` passed: 4 files / 36 tests.
- `bun run build-storybook` passed.
- `packages/react-learn-card`: `bun run build` passed, confirming the SDK dist export includes the new i18n adapter.
- `apps/learn-card-app`: `bun run build` passed after resolving against the freshly built workspace package dist.

One note: I also tried `bunx nx build react`, but that currently fails in the dependent `network-brain-service:build` target with Neo4j type errors outside these touched files. The direct React SDK package build plus the LCA production build passed.

goblincore:
@TaylorBeeston I think this is ready for re-review! Merged in all the new redesigned components and had them translated.

apps/learn-card-app/src/pages/wallet/CredentialPage.tsx
```diff
@@ -45,7 +46,7 @@ const categoryToConfig: Record<string, CategoryConfig> = {
         boostCategory: CredentialCategoryEnum.socialBadge, // category
         subheaderType: SubheaderTypeEnum.SocialBadge, // header type
 
-        title: 'Badges',
+        title: m['wallet.categories.socialBadges'](),
```
TaylorBeeston, line 51: **[P1] Resolve this message during render instead of module initialization.** `categoryToConfig` stores the translated string when this chunk first loads, so changing locale later leaves all seven wallet category titles in the old language. A structural scan found the same eager pattern in 12 files / 97 message calls, including `RecoveryFlowModal`, `TransactionHistoryItem`, and several developer guides. I verified the behavior directly: the stored title remained `Badges` after switching to Spanish while a fresh invocation returned `Insignias`. Store a message function/key and invoke it from a locale-dependent render/factory, and extend the frozen-i18n guard to cover top-level calls.

apps/learn-card-app/src/i18n/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+ *   3. tenant `i18n.defaultLanguage` — set via `setTenantDefaultLocaleCache()`
+ *      from `TenantConfigProvider` before React mounts
+ *   4. `'en'` — hard fallback
+ *
+ * Native device locale (Capacitor `Device.getLanguageCode()`) is async and
+ * handled in the effect below — it takes priority over (2)/(3) but slots in
+ * after first render when running on iOS/Android.
+ */
+function resolveInitialLocale(): SupportedLanguage {
+    const detected = detectInitialLocaleSync(SUPPORTED_LANGUAGES, 'en');
```
TaylorBeeston: **[P1] Intersect compiled locales with the active tenant's `i18n.supportedLanguages`.** Both detection calls currently receive global `SUPPORTED_LANGUAGES`, and `useLanguageSelectorConfig` also starts from that global set. The tenant schema/default only enables `["en"]`, while only LearnCard overrides all four; VetPass therefore can auto-select and expose Spanish/French/Arabic despite its tenant configuration. Derive the effective locale set from `useTenantConfig().i18n.supportedLanguages` for detection, enforcement, and selector visibility, with a single-language-tenant test.

apps/learn-card-app/src/i18n/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+export const LocaleProvider: React.FC<{ children: React.ReactNode }> = ({ children }) => {
+    const [locale, setLocaleState] = useState<SupportedLanguage>(() => {
+        const initial = resolveInitialLocale();
+        paraglideSetLocale(initial, { reload: false });
+        return initial;
+    });
+    const changeLocale = useCallback((lang: SupportedLanguage) => {
+        // 1. Persist to localStorage
+        if (typeof localStorage !== 'undefined') {
+            localStorage.setItem('i18n.language', lang);
```
TaylorBeeston: **[P2] A storage failure must not abort the in-memory locale switch.** `detectLocale.ts` correctly notes that localStorage may throw in private/restricted contexts, but this write happens before both `paraglideSetLocale` and `setLocaleState`. On the deploy preview, forcing `Storage.setItem` to throw `SecurityError` left the page in Arabic after selecting English. Catch persistence failure independently and still update Paraglide/React state.

apps/learn-card-app/src/pages/login/forms/EmailForm.tsx
```diff
@@ -332,7 +336,7 @@ const EmailForm: React.FC<EmailFormProps> = ({
                 <input
                     aria-label="Email"
                     className={`${emailInputBaseClassName} ${resolvedEmailInputClassName} ${emailInputErrorClassName}`}
-                    placeholder="Email address"
+                    placeholder={m['login.email.placeholder']()}
```
TaylorBeeston, line 339: **[P2] Translate the accessible label along with the placeholder.** The adjacent input still has `aria-label="Email"`. In the Arabic deploy preview the visible placeholder was Arabic, but Chromium exposed the textbox's accessible name as `Email`, so screen-reader users do not get the selected locale. Use a localized `login.email.label` value for `aria-label`.

TaylorBeeston (changes requested):
Requesting changes for the locale-switch correctness and tenant-scoping issues noted inline.

Two additional findings could not be attached to changed lines:

- **[P2] Stale Docker bind mount:** `apps/learn-card-app/compose-local.yaml:44` still mounts `./vite.config.ts` after this PR renames the file to `vite.config.mts`. On a fresh checkout the actual config is no longer live-mounted and Compose may create a `vite.config.ts` directory. Update both sides of the mount to `.mts`.
- **[P3] Generated artifacts:** remove the committed `.astro/*` and `dist/*` output under `examples/app-store-apps/6-basic-sync-app/` and ensure those paths are ignored.

Verification performed: the four focused i18n test files passed (36 tests); key, marker, untranslated-value, and frozen-memo guards passed; a production Vite build from PR workspace sources passed; and the Arabic login flow was exercised in the deploy preview.

goblincore:
@TaylorBeeston 

Thanks for the thorough review — the items are addressed. Summary below.

### [P1] Module-scope translation freeze — fixed across all 12 files
Every module-level config that resolved `m['key']()` at import time is now a render-time getter, so titles/labels track a runtime locale switch instead of freezing in the pre-locale default:
- Core: `categoryToConfig` (CredentialPage), `ACTION_TEXT` (TransactionHistoryItem), `RECOVERY_COPY`/`DEFAULT_COPY` (RecoveryFlowModal), `RELATIONSHIP_OPTIONS` (EndorsementPanel), `TRUST_TIER_OPTIONS` (RequirementSatisfiedSpec)
- Developer Portal guides: `STEPS`/`FEATURES` (EmbedAppGuide), `PERMISSION_TO_METHODS`/`SCOPE_OPTIONS` (IntegrationGuidePanel), `SCOPE_OPTIONS`/`STEPS` (ConsentFlowGuide), `STEPS` (IssueCredentials / EmbedClaim / VerifyCredentials / ServerWebhooks)

To catch these reliably and prevent regressions, added an AST lint rule **`no-module-scope-i18n`** (flags `m[...]()` resolved outside any function) and wired it into the `I18nMemos` CI job next to `no-frozen-i18n-memo`. The rule surfaces exactly the set you counted (12 files / 97 calls); the whole app is now clean under it.

### [P1] Tenant scoping
Detection (`resolveInitialLocale`, native-upgrade effect) and the language selector now go through `getEffectiveSupportedLanguages()`, which intersects the app's compiled `SUPPORTED_LANGUAGES` with the active tenant's `i18n.supportedLanguages` (seeded in the bootstrap before React mounts, mirroring the existing `defaultLanguage` cache). A tenant that only enables `['en']` (VetPass) can no longer auto-select or expose es/fr/ar. Falls back to all compiled languages if the tenant cache is unset.

### [P2] localStorage failure aborting the switch
`changeLocale` now switches in-memory first (`paraglideSetLocale` + `setState`) and persists in a `try/catch` afterwards, so a `SecurityError` in private/restricted contexts can't strand the UI in the old locale.

### [P2] `aria-label="Email"`
Wrapped with a new `login.email.label` (es/fr/ar), so the accessible name follows the selected locale.

### [P2] Stale Docker bind mount
`compose-local.yaml` mount updated `./vite.config.ts` → `.mts` on both sides.

### [P3] Committed generated artifacts
Removed the tracked `.astro/` and `dist/` output under `examples/app-store-apps/6-basic-sync-app/` and added them to that example's `.gitignore`.

---

### On the two gitStream items
- **`import { m }` "crashes all AI-session components"** — this one isn't a bug in our setup. The `@inlang/plugin-i18next` build emits a named `m` export, `import { m }` is used in ~106 files, `check-i18n-keys.mjs` documents it as a supported pattern, and the app runs fine in the deploy preview (you exercised the Arabic login flow). No change made here — happy to reconsider if you're seeing an actual runtime error.
- **`personalizationI18n` runtime key lookup** — left as-is for now; can route it through the `mDynamic` safe accessor (returns the key instead of `''` on a miss) as a follow-up if you'd like it in this PR.

One cosmetic note: the CI job is still named "Check for locale-frozen i18n memos" though it now also enforces module-scope — will rename on the next touch.

TaylorBeeston (approved):
Pulled the current head (c294bd14) locally and re-ran the focused verification. The fixes from my earlier review look good, so approving.

Verified:
- i18n key, interpolation-marker, and untranslated-value guards pass
- no-frozen-i18n-memo + no-module-scope-i18n guards pass
- focused suites pass: 4 files / 36 tests
- manual edge checks pass: an English-only tenant resolves to exactly `['en']`, and a `localStorage.setItem` SecurityError does not prevent the in-memory/DOM locale switch
- Paraglide compilation succeeds

Two non-blocking asterisks:
- Please add permanent regression coverage for the single-language tenant path (`setTenantSupportedLanguagesCache` / `getEffectiveSupportedLanguages`); I verified it manually, but it is not represented in the committed tests.
- The native-locale mount effect still reads `localStorage.getItem` outside a try/catch. The reported write-failure behavior is fixed and tested, but wrapping that read too would make restricted-storage handling complete.

I could not complete the entire production build in this fresh checkout because workspace artifacts were not prebuilt and the dependency-aware Nx build could not start its plugin worker in this environment. Vite reached and completed the Paraglide compilation stage, and the deploy-preview status is green.

PR #1267, opened 2026-05-27
[LC-1621] Check Age Gate First on Sign Up

smurflo2:
# Overview

#### 🎟 Relevant Jira Issues

[LC-1621](https://welibrary.atlassian.net/browse/LC-1621)

#### 📚 What is the context and goal of this PR?
We used to check age gate conditions at the very end of our onboarding process, after we'd generated a key and after the user had already entered things like role, name, profile pic, and user id. Not great if they do all that and then get told they're too young. Now we ask for age and country **first**, gate access in the same way we did before, and _then_ go on to ask about role, name, etc.

#### 🥴 TL; RL:
Enforce age gate before profile creation.

#### 💡 Feature Breakdown (screenshots & videos encouraged!)
image

Under 13:
image

U.S. 13-17 (also non-EU)
image

EU 13-(17 / country specific age)
image

18+
Allowed through with no modal

DOB / Country removed from profile screen:
image

#### 🔍 Types of Changes

-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [ ] Chore (refactor, documentation update, etc)

#### 💳 Does This Create Any New Technical Debt? ( If yes, please describe and [add JIRA TODOs](https://welibrary.atlassian.net/jira/software/projects/WE/boards/2) )

-   [x] No
-   [ ] Yes

# Testing

#### 🔬 How Can Someone QA This?

- Log in with an existing account: should be no change in behavior
- Log in with a new account: should see age gate screen first
- Check age and country combinations match the "Feature Breakdown" section

#### 🧪 Code Coverage

# Documentation

#### 📝 Documentation Checklist

**User-Facing Docs** (`docs/` → [docs.learncard.com](https://docs.learncard.com))

-   [ ] **Tutorial** — New capability that users need to learn (`docs/tutorials/`)
-   [ ] **How-To Guide** — New workflow or integration (`docs/how-to-guides/`)
-   [ ] **Reference** — New/changed API, config, or SDK method (`docs/sdks/`)
-   [ ] **Concept** — New mental model or architecture explanation (`docs/core-concepts/`)
-   [ ] **App Flows** — Changes to LearnCard App or ScoutPass user flows (`docs/apps/`)

**Internal/AI Docs**

-   [ ] **AGENTS.md** — New pattern, flow, or context that AI assistants need
[...]

PR #1270, opened 2026-05-29
Fix age-gate-first onboarding: EU guardian email + under-13 redirect

goblincore:
## Summary

Follow-up fixes for PR #1267 (LC-1621), which moves the age gate to run **before** key/profile creation. That reordering broke two first-time-user flows that depended on a profile (or a login redirect) which no longer exists at that point. Branched off `lc-1621-streamline-signup`; targets that branch.

## What changed

**Fix 1 — EU teen guardian-approval flow (defer email to post-profile)**
- `EUParentalConsentModalContent.tsx` gains an optional `onSubmit(guardianEmail)` override. The age-gate flow uses it to *collect* the parent email without sending — the network profile the approval is keyed to doesn't exist yet. Re-prompt callers (which run after a profile exists) keep the existing internal `sendGuardianApprovalEmail` behavior.
- `OnboardingContainer.tsx` stores `guardianEmail` in `formData` and advances onboarding after consent.
- `OnboardingNetworkForm.tsx` creates the EU minor's profile with `approved: false` and, once `createProfile` succeeds, sends the deferred `sendGuardianApprovalEmail` (try/catch, non-fatal). Profile now exists → send + later guardian approval succeed instead of throwing "Requester profile not found".

**Fix 2 — under-13 "I'm an Adult" logout/redirect**
- Dropped the stale `underageFamily=1` param from `routeChildToParentFlow` and `familyInviteUrl`. The child is already logged out, so we rely on the existing `redirectTo` query handling, which fires correctly once the guardian logs in (the old param triggered a second logout that bounced the freshly-logged-in guardian).

**Fix 3 — EU consent modal double `closeModal`**
- The Done button called both `onComplete()` and `onClose()`. Since `onComplete` (age-gate flow) already closes the modal, the extra `onClose()` popped the underlying onboarding modal too, so onboarding never advanced and no profile/email was created. Now calls `onComplete` when present, else falls back to `onClose`.

## Test plan

- [ ] **EU 14yo / Germany:** age gate → enter parent email → Done → onboarding advances to role selection → profile created with `approved=false` → guardian email actually sent → guardian approval link succeeds (no "Requester profile not found").
- [ ] **Under-13 → "I'm an Adult":** child logged out → lands on `/login?redirectTo=/families?createFamily=true` → guardian logs in → lands on `/families?createFamily=true` (no second logout). Repeat via copied `familyInviteUrl`.
- [ ] **Regressions:** US 13–17, non-EU, 18+, and existing-account login unchanged.

### DIFF
--- a/apps/learn-card-app/src/components/onboarding/OnboardingContainer.tsx
+++ b/apps/learn-card-app/src/components/onboarding/OnboardingContainer.tsx
@@ -48,6 +48,7 @@ const OnboardingContainer: React.FC<OnboardingContainerProps> = ({ onSuccess, in
         photo: currentUser?.profileImage ?? '',
         usMinorConsent: false,
         euParentalConsentRequested: false,
+        guardianEmail: '',
         profileId: '',
     });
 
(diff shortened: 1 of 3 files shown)


smurflo2 (approved):
Thanks Donny! This looks great! Elegant solution to the flow issue 🏄

PR #1285, opened 2026-06-05
fix: CJS Bundler Issues

Custard7:
# Overview

Our post-publish smoketests caught two of our packages failing to load once installed from npm, and the PR check that validates packages was timing out. This fixes all of it.

The packages worked fine inside our repo but broke for outside users, because the published builds weren't quite right. None of our normal CI caught it — only the smoketests did, because they install from npm the way a real user would.

#### 🥴 TL; DR:

Three packaging bugs in `@learncard/helpers` and `@learncard/crypto-plugin` that broke `import`/`require` for npm users, plus a CI check that was too slow and got auto-cancelled.

#### 💡 What changed

**1. `@learncard/helpers` crashed with "require() of ES Module" on Node 18.**
The build pulled in `query-string`, which is now ESM-only and can't be `require()`d from our CommonJS build. We only used it in two image-helper files for basic URL parsing, so we replaced it with the built-in `URLSearchParams` (new `url.helpers.ts`) and dropped the dependency.

**2. `@learncard/crypto-plugin` and `@learncard/helpers` threw "Named export not found" when imported.**
Our esbuild config wasn't telling esbuild to target Node, so the CommonJS output didn't include the marker Node needs to expose named exports to `import`. Adding `platform: 'node'` to the CommonJS builds fixes it.

**3. `@learncard/helpers` couldn't find `zod`.**
It imports `zod` but never declared it as a dependency — it only worked locally because another package happened to install it. Added it to `dependencies`.

**4. The "Validate Published Package Contracts" PR check kept timing out.**
It validated ~36 packages one at a time (each shells out to `npm pack`), which ran past the 15-minute limit and got cancelled — it looked like a hang but was just too slow. We now validate them in parallel (≈4–5x faster) and raised the timeout for headroom.

Also moved `@learncard/helpers` into the smoketest's "must pass" list so these fixes stay fixed.

None significant. The `URLSearchParams` replacement covers exactly what we used `query-string` for (URL parsing + rebuilding, preserving insertion order, fragments, and `%20` space encoding). One intentional difference: `query-string` sorted query keys alphabetically by default; we preserve insertion order instead. The call sites are image-CDN URLs (`w`/`q`/`auto=format`) where key order doesn't matter, and no tests assert exact URL strings.

#### 🔍 Types of Changes

-   [x] Chore (CI: parallelize + retime the package-validation check)

-   [x] No

> Note: 7 other packages (ceramic, didkey, idx, lca-api, learn-cloud, network, simple-signing) still have the same `platform: 'node'` issue and remain on the smoketest "known broken" list. They're untouched here — a follow-up can apply the same one-line build fix to each.

# Testing

The published-package smoketests are the real test, but you can verify locally:

1. `pnpm install` then build the two packages:
   `pnpm exec nx run-many --target=build --projects=crypto-plugin,helpers`
2. Confirm the named import that used to fail now works:
   ```
   node --input-type=module -e "import { CryptoPlugin } from './packages/plugins/crypto/dist/crypto-plugin.cjs.development.cjs'; console.log(typeof CryptoPlugin)"
   ```
   → prints `object` (previously threw "Named export 'CryptoPlugin' not found").
3. Confirm helpers loads via `require` without the ESM/zod errors:
   ```
   node -e "console.log(typeof require('./packages/learn-card-helpers/dist/helpers.cjs.development.cjs').isHex)"
   ```
   → prints `function`.
4. Run the package validator and confirm both pass:
   `node scripts/validate-packages.mjs --only=crypto-plugin,helpers`
   → both show ✓.

CI: the "Validate Published Package Contracts" check should finish in a few minutes instead of timing out.

#### 🧪 Code Coverage

No new unit tests — the published-package smoketests and the validate-packages check are the coverage for this. Existing tests unaffected.

# Documentation

#### 💭 Documentation Notes

Internal build/packaging fix, no public API changes. The non-obvious "why" (e.g. why `platform: 'node'` is required) is captured in code comments next to each change.

# ✅ PR Checklist

### ✨ PR Description
Purpose: Fix CommonJS bundler issues by replacing ESM-only query-string dependency with native URLSearchParams and adding platform: 'node' to esbuild CJS builds for proper named export annotations.

Main changes:
- Created url.helpers.ts with URLSearchParams-based parseUrl/stringifyUrl functions replacing query-string dependency
- Added platform: 'node' to esbuild CJS configurations in helpers and crypto-plugin build scripts
- Removed query-string from dependencies; updated validation scripts with async concurrency pool and moved helpers from KNOWN_BROKEN to GATING packages


Custard7:
@claude please review

Custard7:
@claude please re-review

TaylorBeeston (approved):
LGTM! Thank you!

PR #1297, opened 2026-06-11
fix: load init's CJS bundle via createRequire in Node ESM entry

Custard7:
## What

`@learncard/init`'s Node ESM entrypoint (`dist/node-esm.mjs`) used to load the package's CommonJS bundle with a static `import init from './index.cjs'`. This changes it to load that same CJS bundle through `createRequire` instead. No behavior change for normal Node consumers — they still get the exact same CJS bundle.

I also added a check to `scripts/validate-packages.mjs` that fails if any package's ESM entrypoint statically imports a `.cjs`/`.cts` file.

## Why

A static `import` of a `.cjs` file from an ESM module is the one shape that breaks bundlers. When a bundler (Astro/Vite SSR, etc.) processes `@learncard/init` as a symlinked workspace dependency, it follows that import into `index.cjs` and tries to transform the CommonJS file as ESM, crashing with `module is not defined`. This was hit by `examples/app-store-apps/5-northstar-learning`.

`createRequire` keeps the CJS load opaque to the bundler and hands it to Node's native loader, where `module`/`require` exist. So the package now works in all three modes — published installs, plain Node/production, and bundler-transformed symlinked workspace deps — without consumers needing `ssr.external` / `optimizeDeps` workarounds.

The validation check exists because our current package validators (publint + attw) don't catch this: attw's bundler check is static and never transforms symlinked source, so it reported `init` as fine while it was actually broken. The new check guards the structural invariant directly so this class can't regress.

## How to review

1. Read `packages/learn-card-init/scripts/nodeEsmEntrypoint.mjs` — confirm it loads `./index.cjs` via `createRequire` and re-exports the same names as before.
2. Read the new check in `scripts/validate-packages.mjs` (the `checkNoEsmToCjsImport` / `STATIC_CJS_IMPORT` addition) — it scans ESM-side export entries and fails on a static `.cjs`/`.cts` import.
3. Build init and confirm the example app works without any bundler workaround:
   - `pnpm --filter @learncard/init build`
   - `cd examples/app-store-apps/5-northstar-learning && pnpm build` (and `pnpm dev`, then trigger a credential action) — should succeed with no `module is not defined`.
4. Run `node scripts/validate-packages.mjs --skip-publint --skip-attw` — all packages pass. To confirm the check actually bites, temporarily restore the old `import init from './index.cjs'` line and re-run; it should fail.

## Notes

This is a build-script + tooling change only; no shipped source code changes. `dist/` is regenerated by the build, so nothing in `dist/` is committed.

---

### ✨ PR Description
Purpose: Fix ESM-to-CJS static import that breaks bundler-transformed symlinked packages by switching to runtime `createRequire` loading.

Main changes:
- Replace static ESM import of `index.cjs` with `createRequire` in Node ESM entrypoint to prevent bundler transform failures
- Add validation guard in scripts to detect and reject static CJS imports from ESM entries across all packages
- Restructure package.json exports with explicit node/import/require conditions and type definitions per entry point


Custard7:
@claude please review

TaylorBeeston (approved):

PR #1313, opened 2026-06-18
fix: Stale Auth Cache

Custard7:
# Overview

Users on Android (and web) kept landing in a "mixed login state": after being logged in for a while, reopening the app would show "Complete Profile", pop the age gate, and present the Join Network modal — even though they'd already onboarded and have a profile.

We thought a previous PR (LC-1738, "Stale Auth Error on Android Resume") had fixed this, but it only hardened the `AuthCoordinator`. The onboarding prompts don't read the coordinator — they read a separate react-query hook (`useGetProfile` / `useIsCurrentUserLCNUser`), which still had the race.

#### 🥴 TL; RL:

On app resume the persisted user rehydrates instantly, so `isLoggedIn` flips to true *before* the private key/wallet is rebuilt. The profile query ran too early, the wallet wasn't ready, the failure got swallowed and cached as "this user has no profile" for 5 minutes — which is what triggered all three prompts.

#### 💡 Feature Breakdown

Three coordinated changes in `packages/learn-card-base/src/react-query/queries/queries.ts`:

1. **Wait for the wallet.** The "my own profile" query now stays disabled until the wallet is actually present, so it can't run during the pre-key window and cache a bogus empty result.
2. **Don't swallow failures.** If the wallet can't resolve, the query now throws (so react-query marks it as an error and retries) instead of quietly returning "no profile".
3. **Treat errors as "still loading".** `useIsCurrentUserLCNUser` now counts an error as "still resolving", so a transient hiccup never reads as "user has no profile" and the onboarding prompts stay hidden.

A genuinely new user (wallet ready, profile really is empty) still gets the prompts as before.

#### 🔍 Types of Changes

-   [ ] New feature
-   [ ] Breaking change
-   [ ] Chore

-   [x] No

# Testing

1. Log in normally and confirm you reach the wallet with your profile intact.
2. Background the app for a while (or force-kill it on Android), then reopen.
3. Confirm you do **not** see "Complete Profile", the age gate, or the Join Network modal — you should land straight in your logged-in state.
4. As a brand-new user (no profile yet), confirm the "Complete Profile" / onboarding prompts still appear correctly.

Android Native (primary), plus Chromium / iOS for the resume case.

#### 🧪 Code Coverage

No new tests in this PR — the change is in the gating/error handling of an existing query. Recommend running `pnpm exec nx test learn-card-base` before merge.

# Documentation

#### 💭 Documentation Notes

Internal bug fix, no API changes. Rationale is captured in inline comments on the non-obvious lines (why errors are treated as "loading" and why the query waits for the wallet).

### ✨ PR Description
Purpose: Fix stale auth cache on app resume by preventing transient wallet failures from falsely caching "no profile" and triggering premature onboarding prompts.

Main changes:
- Modified `useGetProfile` to gate queries on wallet readiness and distinguish between unresolved errors and settled "no profile" states
- Updated error handling in `useIsCurrentUserLCNUser` and onboarding components to treat fetch errors as unresolved rather than conclusive
- Added test case validating existing users aren't redirected to age gate when profile fetch fails

### DIFF
--- a/apps/learn-card-app/environments/learncard/assets/config/capacitor.config.json
+++ b/apps/learn-card-app/environments/learncard/assets/config/capacitor.config.json
@@ -41,7 +41,7 @@
 		},
 		"CapacitorUpdater": {
 			"appId": "com.learncard.app",
-			"autoUpdate": false
+			"autoUpdate": true
 		}
 	},
 	"packageClassList": [
--- a/apps/learn-card-app/environments/learncard/assets/config/capacitor.config.json
+++ b/apps/learn-card-app/environments/learncard/assets/config/capacitor.config.json
@@ -41,7 +41,7 @@
 		},
 		"CapacitorUpdater": {
 			"appId": "com.learncard.app",
-			"autoUpdate": true
+			"autoUpdate": false
 		}
 	},
 	"packageClassList": [
--- a/.changeset/witty-pillows-cheer.md
+++ b/.changeset/witty-pillows-cheer.md
@@ -1,0 +1,5 @@
+---
+"learn-card-base": patch
+---
+
+fix: prevent mixed login state on app resume — a transient wallet/private-key reconstruction failure no longer caches a spurious "no profile" result, which previously surfaced "Complete Profile", the age gate, and the JoinNetworkModal for users who had already onboarded
--- a/apps/learn-card-app/capacitor.config.ts
+++ b/apps/learn-card-app/capacitor.config.ts
@@ -52,7 +52,7 @@ const config: CapacitorConfig = {
         },
         CapacitorUpdater: {
             appId: 'com.learncard.app',
-            autoUpdate: true,
+            autoUpdate: false,
             // SINGLE SOURCE OF TRUTH for the Capgo channel.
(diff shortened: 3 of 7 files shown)


Custard7:
@claude please review

Custard7:
@claude re-review

gerardopar (approved):
@Custard7 tested on android, working for me 👍🏽

https://github.com/user-attachments/assets/ddc2edb6-7582-42dd-aec1-d5140c2cca7e

1. was able to confirm onboarding still works as expected 
2. cold starts on android work as expected 
3. user's auth is persisted with no issues 🚀

apps/learn-card-app/environments/learncard/assets/config/capacitor.config.json
```diff
@@ -41,7 +41,7 @@
 		},
 		"CapacitorUpdater": {
 			"appId": "com.learncard.app",
-			"autoUpdate": true
+			"autoUpdate": false
```
gerardopar: @Custard7 was this meant to be turned off?

Custard7: Ah I was testing it with a live push! Thank you for testing and approving!! 😎👍🏻🙌🏻

PR #1327, opened 2026-06-24
Upgrade Astro apps for Zod 4

TaylorBeeston:
# Overview

#### 🎟 Relevant Jira Issues

N/A — dependency maintenance for Astro example apps.

#### 📚 What is the context and goal of this PR?

Astro apps were pinned to older Astro versions that pulled Zod 3, forcing an `astro>zod` pnpm override while the monorepo uses Zod 4. This PR moves every Astro workspace app to the Zod 4-compatible Astro 6 line, removes the Astro-specific Zod 3 override, standardizes repo Node pins on 24.12.0, and updates the dependency review path that was failing on vulnerable transitive packages.

#### 🥴 TL; RL:

Upgrade Astro apps to `astro@^6.4.8`, keep official integrations compatible, remove `astro>zod`, lock Zod to 4.4.3, standardize Node to 24.12.0, and override vulnerable `@opentelemetry/core`/`esbuild` transitive versions.

#### 💡 Feature Breakdown (screenshots & videos encouraged!)

- Upgraded app-store Astro examples from Astro 5 to `^6.4.8`.
- Upgraded existing Astro 6 examples from `^6.0.6` to `^6.4.8`.
- Upgraded the legacy Snap dapp from Astro 1 to Astro 6 and matching `@astrojs/react`/`@astrojs/tailwind` versions.
- Removed `pnpm.overrides["astro>zod"]`; the lockfile now resolves Zod 4 only.
- Standardized Node pins in `.nvmrc`, Dockerfiles, package `engines`, and GitHub Actions to Node 24.12.0.
- Updated dependency review actions and lockfile resolutions for `@opentelemetry/core@2.8.0` and `esbuild@0.28.1`.
- Updated the Snap CHAPI Vite config's node-globals polyfill import for the upgraded Astro/Vite stack.

#### 🛠 Important tradeoffs made:

- The Snap examples' direct `astro build` is still intentionally blocked by their existing skipped MetaMask Snap output; their package `build` scripts remain the verification path.
- The smoketest matrix now runs package-manager coverage on the standardized Node version instead of old LTS Node versions.

#### 🔍 Types of Changes

-   [x] Bug fix (non-breaking change which fixes an issue)
-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
-   [x] Chore (refactor, documentation update, etc)

[...]

PR #1331, opened 2026-06-27
fix: repair OpenAPI generation under Zod 4.4 (brain + cloud services boot)

goblincore:
## Summary

The brain-service and cloud-service crash at **startup** under the current Zod 4.4.3 pin, taking down staging (`/api/health-check` → 500 on `staging.network.learncard.com` and `staging.cloud.learncard.com`) and timing out the e2e runner. `generateOpenApiDocument()` runs eagerly at module load (imported by both `docker-entry.ts` and `lambda.ts`), so a generation failure crashes **every** Lambda at cold start. Prod is unaffected only because it still serves the pre‑Zod‑4.4 artifact.

### Root cause
The deploy that broke it (`#1327`) bumped `pnpm.overrides.zod` `4.1.13 → 4.4.3`. **Zod 4.3** tightened two behaviors that `trpc-to-openapi@3.1.0` / `zod-openapi` relied on during OpenAPI generation (confirmed by version-matrix repro: 4.1.13 and 4.2.x generate fine, 4.3.0+ throw):

1. `.omit()` is no longer allowed on object schemas containing refinements — and `trpc-to-openapi` calls `.omit()` internally on **every** route input (`getRequestBodyObject`). This is what threw `Error: .omit() cannot be used on object schemas containing refinements` in brain-service.
2. `z.custom()` can no longer be represented in OpenAPI — threw `Zod schema of type 'custom' ... cannot be represented in OpenAPI` for the custom-storage routes in cloud-service.

## Fix (stays on Zod 4.4.3 — no schema rewrites, no `@learncard/types` changes)

- **Bump `trpc-to-openapi` override `3.1.0 → 3.3.0`.** 3.3.0 handles refined route inputs under Zod 4.4 (verified for both path-param and non-path-param cases), fixing all the brain-service refined-input crashes without touching any validator.
- **cloud-service `customStorage.ts`:** the `count` / `update` / `delete` query schemas swap `z.custom<Filter<…>>()` → `z.record(z.string(), z.any())`, matching the already-working `read` route. Handlers already cast `as any`, so runtime behavior is unchanged.

## Verification

- Repro with `zod@4.4.3` + `trpc-to-openapi@3.3.0` + `zod-openapi@5.4.5`: both failing patterns now generate successfully; the cloud `z.record` shape generates cleanly.
- `pnpm install --frozen-lockfile` passes; lockfile resolves a single `zod@4.4.3` and single `trpc-to-openapi@3.3.0` (no version mixing).
- Lockfile diff is trpc-only (no unrelated transitive drift).

## Recommended follow-up (not in this PR)

Make `generateOpenApiDocument()` lazy / fault-tolerant in both services' `openapi.ts` so a future schema incompatibility degrades the `/docs` endpoint instead of crashing the whole service at boot. That hardening is what turns "one bad schema" from a full outage into a non-event.

---

### ✨ PR Description
Purpose: Fix OpenAPI generation failures under Zod 4.4 by replacing incompatible custom/instanceof validators and add MongoDB query injection protection.

Main changes:
- Replaced `z.instanceof(RegExp)` with `z.preprocess` converting RegExp to string representation for OpenAPI compatibility
- Removed `z.custom()` validators from custom-storage query schemas, replaced with `z.record(z.string(), z.any())`
- Added `assertSafeMongoQuery()` helper to reject MongoDB server-side-JavaScript operators ($where, $function, $accumulator) and enforce depth limits
- Upgraded `trpc-to-openapi` to 3.3.0 and added `@parcel/watcher` 2.5.1 pnpm overrides for Zod 4.4 compatibility

### DIFF
--- a/.changeset/fix-zod4-openapi-generation.md
+++ b/.changeset/fix-zod4-openapi-generation.md
@@ -1,0 +1,16 @@
+---
+"@learncard/network-brain-service": patch
+"@learncard/learn-cloud-service": patch
+---
+
+Fix OpenAPI document generation under Zod 4.4 so the brain and cloud services boot.
+
+Zod 4.3 tightened two behaviors that broke `generateOpenApiDocument` (which runs
+eagerly at service startup, so a failure crashed every Lambda at cold start):
+
+- `.omit()` is no longer allowed on object schemas containing refinements, which
+  `trpc-to-openapi` calls internally on every route's input. Bumping the
+  `trpc-to-openapi` override to `3.3.0` resolves this for all refined route inputs.
+- `z.custom()` can no longer be represented in OpenAPI. The custom-storage
+  `count`/`update`/`delete` query schemas now use `z.record(z.string(), z.any())`,
+  matching the already-working `read` route.
--- a/.changeset/fix-zod4-openapi-generation.md
+++ b/.changeset/fix-zod4-openapi-generation.md
@@ -1,6 +1,7 @@
 ---
 "@learncard/network-brain-service": patch
 "@learncard/learn-cloud-service": patch
+"@learncard/types": patch
 ---
 
 Fix OpenAPI document generation under Zod 4.4 so the brain and cloud services boot.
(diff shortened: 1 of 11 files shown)


goblincore:
Thanks — good catch on the NoSQL operator surface. Assessment + fix:

**Not a regression (confirmed):** the old `z.custom<Filter<…>>(…)` used `z.record(z.string(), z.any()).parse` as its runtime validator, so what reaches Mongo is byte-for-byte identical before and after this PR. The change only relaxed the TS type.

**Cross-tenant reads aren't possible:** the access layer applies the `did` scope *after* the caller's query (`{ ...query, did: { $in: dids } }`), so a caller can't override it — top-level operators like `$or`/`$and` get ANDed with the `did` filter.

**Real residual vector — addressed:** authenticated users could still inject server-side-JS operators (`$where`, `$function`, `$accumulator`) within their own scope → DoS / arbitrary JS on the DB. Added `assertSafeMongoQuery()` ([helpers/query.helpers.ts](services/learn-card-network/learn-cloud-service/src/helpers/query.helpers.ts)) which recursively rejects those operators (at any nesting depth, incl. inside `$expr`/`$and`) with a `BAD_REQUEST`, wired into all four query routes (`read`/`count`/`update`/`delete`). Normal operators (`$gt`, `$in`, equality, …) pass through unchanged.

Pushed in 2ed497d. ReDoS via `$regex` is a separate, broader hardening I've left out of this hotfix.

goblincore:
@claude please review

Custard7 (approved):
@goblincore okay, all tests passing locally for me after latest patches I pushed up. If you can verify on your end, then we should be g2g ✅

Custard7:
@goblincore pushed up small changes to address review bots comments

goblincore:
On the `RegExpValidator` finding (issue 5) — I believe this is a **false positive**. The premise is that the old union let plain strings through "without a format check," but `RegExpStringValidator` has always had the `.refine` requiring `/pattern/flags`, so bare strings were rejected *before and after* this change.

Verified empirically (zod 4.4.3), old `z.instanceof(RegExp).or(RegExpStringValidator)` vs new `z.preprocess(…, RegExpStringValidator)`:

| input | OLD | NEW |
|---|---|---|
| `RegExp` instance `/hello/i` | OK | OK |
| delimited string `"/hello/i"` | OK | OK |
| bare string `"hello"` | REJECTED | REJECTED |
| bare string `"somepattern"` | REJECTED | REJECTED |

The `z.preprocess` rewrite only changed how `RegExp` **instances** are handled (converting them to a `/source/flags` string to avoid the `z.instanceof` → zod-openapi crash that this PR is about); the accept/reject semantics for strings are unchanged. So no regression and no fix needed here.

Separately/orthogonally: bare-string `$regex` (e.g. `{ $regex: "pattern" }`, valid in Mongo) is rejected by both the old and new validators. If we *want* to support that, it's a pre-existing enhancement we can take in a follow-up — but it's not something this PR broke.

goblincore:
@Custard7 cool yeah all tests green for me!

PR #1335, opened 2026-06-30
fix [LC-1939] : repair published-packages smoketest (zod v4 hoist break + pnpm pin)

goblincore:
# Overview

#### 🎟 Relevant Jira Issues

Fixes: [LC-1939](https://welibrary.atlassian.net/browse/LC-1939)

The scheduled **Smoketest Published NPM Packages** workflow installs our freshly published packages into a clean throwaway project across npm / yarn / pnpm / esbuild. It catches packaging problems the monorepo hides (export maps, ESM/CJS interop, dependency resolution, bundler compat). It had **three independent failures** (yarn, pnpm, esbuild). This PR fixes all three.

The original version of this PR fixed the first two and deferred the esbuild one to a follow-up. That follow-up is now included here, so the branch is complete.

#### 💡 What changed

**Fix 1 — `@learncard/types` fails to import under yarn/pnpm** (`Cannot read properties of undefined (reading 'datetime')`)

The types ESM build externalizes `import { z } from 'zod'`, so it depends on which zod the consumer's tree resolves. `expo` transitively drags in the zod `3.25.76` bridge release, whose default export is still v3 (no `z.iso`). Hoisting non-deterministically bound types' import to v3, so `z.iso.datetime()` in `lcn.ts` threw at import time.

Fix: import from the `zod/v4` subpath across `@learncard/types` source. That subpath resolves the v4 API on `3.25.76`, `4.1.13`, and `4.4.3`, regardless of hoisting.

**Fix 2 — pnpm smoketest job: `No pnpm version is specified`**

`pnpm/action-setup@v6` has no bundled default; it reads the version from a `packageManager` field in the cwd's `package.json`. The smoketest runs in a throwaway dir with no such field, so the step errored before testing anything. Pinned `version: 9` to match the repo's pnpm major (mirrors how the yarn step pins `yarn@stable`).

**Fix 3 — esbuild bundler job crashes bundling `@learncard/init`**

Root cause: the four generated tRPC client packages (`lca-api-client`, `simple-signing-client`, `network-brain-client`, `learn-cloud-client`) shipped only a CJS entry with no `exports` map. Under Node's ESM loader this bound to the CJS file, whose named exports (`getClient`, `getApiTokenClient`) the CommonJS lexer couldn't see through the `module.exports = require(...)` indirection — so `import { getClient } from ...` threw "getClient not found". To work around that, `@learncard/init` routed its Node ESM entry through a `createRequire(import.meta.url)` shim (`node-esm.mjs`) that loaded the CJS bundle. Bundling that shim with `esbuild --platform=node` compiled `import.meta` away → `createRequire(undefined)` → crash.

Fix (the real one, not a workaround):

- Gave each of the four client packages a proper dual-format setup, matching what `@learncard/types` and `@learncard/helpers` already do: `"type": "module"`, CJS bundles emitted with a `.cjs` extension, and a dual `exports` map (`import` → real ESM build, `require` → CJS entry). They now expose clean named exports to Node's ESM loader.
- Added the `files: ["dist"]` field these four packages were missing. `dist` is gitignored, so without it `pnpm pack` shipped zero build artifacts. This was a pre-existing latent bug surfaced while validating the change.
- Pointed `@learncard/init`'s Node ESM entry directly at `init.esm.js` and deleted the `createRequire` shim. `init.esm.js` now loads natively under Node ESM and bundles cleanly under esbuild/Vite.
- Extended `scripts/validate-packages.mjs` to cover the three network clients so this class of regression is caught before publish going forward.

The didkit backend is unchanged: it's still selected at call time via the `didkit: 'node'` argument (a lazy `import('@learncard/didkit-plugin-node')`), independent of the entry module format. Both the ESM and CJS bundles carry both didkit variants, so swapping init's entry does not change which backend runs.

#### 🔍 Types of Changes

-   [x] Chore (CI workflow fix)

-   [x] No — the esbuild follow-up that was previously deferred is now done. All three smoketest jobs are addressed.

# Testing

Everything below passes locally on this branch.

1. Build the affected packages:

```bash
pnpm exec nx run-many -t build \
  --projects=lca-api-client,simple-signing-client,network-brain-client,learn-cloud-client,init \
  --skip-nx-cache
```

2. Run the repo's packaging validator (publint + are-the-types-wrong, against the actual `pnpm pack` tarballs). This is the check that would have caught the original init regression before publish:

```bash
node scripts/validate-packages.mjs \
  --only=lca-api-client,network-brain-client,learn-cloud-client,simple-signing-client,init
# expect: 5 gating package(s) valid; 0 advisory.
```

3. Confirm Node now sees the client named exports, and init resolves/bundles in every mode:

```bash
# clients under Node ESM
node --input-type=module -e "for (const p of ['@learncard/network-brain-client','@learncard/learn-cloud-client','@learncard/lca-api-client','@learncard/simple-signing-client']) { const m = await import(p); console.log(p, Object.keys(m).filter(k=>k!=='default').join(', ')); }"

# init: ESM, CJS, and the esbuild bundle that used to crash
node --input-type=module -e "console.log(typeof (await import('@learncard/init')).initLearnCard)"
node -e "console.log(typeof require('@learncard/init').initLearnCard)"
node --input-type=module -e "import esbuild from 'esbuild'; const r=await esbuild.build({stdin:{contents:\"import {initLearnCard} from '@learncard/init'\",resolveDir:process.cwd(),loader:'js'},bundle:true,platform:'node',format:'esm',write:false}); console.log('esbuild node bundle OK', r.outputFiles[0].contents.length)"
```

4. Runtime smoke through the ESM path:

```bash
node --input-type=module -e "import { initLearnCard } from '@learncard/init'; const lc = await initLearnCard({ seed: 'a'.repeat(64) }); const vc = await lc.invoke.getTestVc(); const signed = await lc.invoke.issueCredential(vc); console.log(await lc.invoke.verifyCredential(signed));"
# expect: { checks: ['proof','expiration'], warnings: [], errors: [] }
```

After merge and republish, the daily **Smoketest Published NPM Packages** workflow's yarn, pnpm, and esbuild jobs should all go green.

#### 🧪 Code Coverage

No new unit tests. The published-packages smoketest and `scripts/validate-packages.mjs` are the regression tests for this class of issue. The zod change is a mechanical import-specifier swap; the client and init changes are packaging config validated by the tooling above.

# Documentation

Internal packaging/CI fix — no API or behavior changes for SDK consumers, so no user-facing or AGENTS.md docs needed.

# ✅ PR Checklist

-   [x] Related to a Jira issue (LC-1939)
-   [x] My code follows style guidelines (eslint / prettier)
-   [x] I have manually tested common cases — verified via clean external installs and the validator (see Testing)
-   [x] I have reviewed my code
-   [x] I have commented my code, particularly where ambiguous
-   [ ] New and existing unit tests pass locally — N/A (packaging + CI config)
-   [x] I have completed the Documentation Checklist above (or explained why N/A)

-   [x] Code is backwards compatible (CJS `require` and bundler `import` both still resolve)
-   [x] There is not a "Do Not Merge" label on this PR
-   [x] I have thoughtfully considered the security implications of this change
-   [x] This change does not expose new public facing endpoints

[LC-1939]: https://welibrary.atlassian.net/browse/LC-1939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### ✨ PR Description
Purpose: Fix published-packages smoketest by upgrading zod imports to v4 subpath and correcting CommonJS/ESM export maps across tRPC client packages.

Main changes:
- Updated all zod imports from bare `zod` to `zod/v4` subpath to ensure v4 API stability regardless of hoisted zod major version
- Added dual `exports` maps, `.cjs` file extensions, and `"type": "module"` to tRPC clients (lca-api-client, brain-client, cloud-client, simple-signing-client) for proper ESM/CJS resolution
- Removed `node-esm.mjs` shim and simplified `@learncard/init` exports to directly reference ESM build instead of CJS bridge that breaks bundlers


goblincore:
@Custard7 You can take over this one if you'd like - as noted in the PR description, this branch currently fixes the first two issues (the pnpm test failure because of no version/the yarn test because of zod issue) as those are more straightforward - but does not fix the esbuild one, which is abit more complicated:

**The problem**: @learncard/init's Node ESM entry (node-esm.mjs) uses createRequire(import.meta.url) to load its CJS build. When esbuild bundles it for node, it compiles import.meta away → createRequire(undefined) → crash.

**Why it's not a quick fix:** every shortcut trades one break for another —

- Re-export the ESM build (init.esm.js) → esbuild works, but it breaks plain Node ESM, because init.esm.js reaches CJS-only @learncard/*-client packages whose named exports Node's ESM loader can't see (getClient not found).
- Guard import.meta + require the package subpath → passes the smoketest only because node_modules happens to be present; a real bundle-and-ship (Lambda) still breaks. That's masking.
- Re-inline everything into one self-contained bundle → works, but reintroduces the exact bloat + double-didkit hazard that externalizing the plugins was designed to avoid.

The real fix: those CJS-only *-client packages (learn-cloud-client, lca-api-client, network-brain-client, simple-signing-client) are the actual root cause — and not by coincidence: they're also the smoketest's entire "known broken" advisory list. Fix them to export ESM-cleanly (dual build / lexer-friendly named exports — likely modest, since they're small generated clients). 

Then:
init.esm.js loads natively under Node ESM and bundles cleanly under esbuild/Vite,
the createRequire hack (node-esm.mjs) can be deleted entirely,
the advisory failures clear too.

One decision remains: keep native didkit on the CJS require path, wasm on the ESM path (which bundling requires anyway).

So: fixing the CJS-only client packages is the correct solution — it collapses three separate failures into zero instead of shuffling them around. (Interim, if that's too broad right now: ship a dedicated self-contained @learncard/init/bundled entry for the Lambda case, leaving the default deduped.)

That's captured in [LC-1939](https://welibrary.atlassian.net/browse/LC-1939) as the remaining work beyond [PR #1335](https://github.com/learningeconomy/LearnCard/pull/1335).

[LC-1939]: https://welibrary.atlassian.net/browse/LC-1939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Custard7:
@goblincore okay, I went ahead and worked on the technical debt issues you described to fix those other packages. If you can confirm build on your end and passing tests, this shuold be g2g 👍

goblincore:
@Custard7 can confirm it looks good to me, local probes green think we can merge

Custard7 (approved):

PR #1343, opened 2026-07-02
[LC-1940] ✨ Ship native DIDKit to Lambda via a Layer (follow-up to #1341)

goblincore:
## TL;DR

**Follow-up to #1341** (stacked on that branch — merge #1341 first; GitHub will retarget this to `main` automatically). #1341 made the WASM fallback resolvable; this PR makes the **native-first path actually succeed in Lambda**, restoring the pre-Bun-migration behavior where the fast napi addon serves DIDKit ops instead of silently falling back to WASM on every cold start.

📄 Context: [Confluence — Staging Smoketest Failure investigation](https://welibrary.atlassian.net/wiki/spaces/LC/pages/993820674) (see "packaging native DIDKit correctly").

## Why a Layer

The native wrapper locates its binary at runtime via:

```js
const packageJson = require.resolve('@learncard/didkit-plugin-node/package.json');
const nodeFile = fs.readdirSync(path.dirname(packageJson)).find(f => f.startsWith('index.') && f.endsWith('.node'));
```

So the package must exist as a **real, resolvable package on the module path** — it can't be bundled by esbuild or dropped next to the handler (unlike the WASM). Lambda mounts layers at `/opt` and puts `/opt/nodejs/node_modules` on `NODE_PATH`, which satisfies this exactly. And the layer is small in scope: the compiled `dist/` has **zero runtime deps beyond node builtins** (`@learncard/*` imports are type-only), so the layer contains just this one package (~19.6 MB unzipped, ~7.7 MB zipped, well under limits).

## What's in the PR

| Piece | What it does |
|---|---|
| `scripts/prepare-didkit-native-layer.cjs` | Assembles `<service>/didkit-native-layer/nodejs/node_modules/@learncard/didkit-plugin-node/` (package.json + dist + `index.*.node`) from the package root. In CI the existing **didkit-native-linux-x64 artifact** already provides dist + binary there — zero new CI plumbing. Locally the binary is absent → warns → runtime falls back to WASM as today. |
| `serverless.yml` ×3 (brain, lca, learn-cloud) | Defines the `didkitNative` layer, attaches via `provider.layers` (all functions), and adds the package to `custom.esbuild.exclude` — which in serverless-esbuild both marks it esbuild-external **and** stops the packager from installing a binary-less registry copy that would shadow the layer (`/var/task/node_modules` wins over `NODE_PATH`). `aws-sdk` kept in exclude (it's the default being overridden). |
| `serverless-deploy` npm scripts ×3 | Run the prep script before `serverless deploy` (nx-appended `--stage/--region` args still land on the deploy command). |
| `learnCard.helpers.ts` ×3 | `console.warn` when falling back to WASM — the **silent catch** is what hid "native never loads in Lambda" for months. Post-deploy, absence of this warning in CloudWatch = native confirmed live. |

**Untouched on purpose:** simple-signing-service (never uses the native plugin — it's WASM-only by design) and the scouts lca config (`serverless-scouts.yml`, keeps today's WASM-fallback behavior).

## Verification done locally

- ✅ Layer assembly tested **both ways**: without a binary (warn + WASM-fallback path) and with the real CI artifact binary dropped in
- ✅ Full resolution chain simulated from a neutral cwd with `NODE_PATH=<layer>` (mimics `/var/task` + `/opt`): `require.resolve` of package.json → readdir finds `index.linux-x64-gnu.node` → main entry resolves to `dist/index.js`
- ✅ YAML parses ×3, `tsc --noEmit` clean ×3
- ⏳ The final step — dlopen of the linux binary — is **deploy-only**. Verify on staging via CloudWatch: DIDKit-touching requests should show **no** `[didkit] native plugin unavailable` warning.

## Notes / risks for review

1. **Safe-by-construction fallback:** if the layer fails to attach or the binary won't load, the existing try/catch falls back to WASM (which #1341 fixed) — worst case is today's behavior, now with a log line.
2. **`provider.layers` + serverless-lift:** brain's notifications-queue worker is created by lift; if it doesn't inherit provider layers, that one function just stays on WASM fallback. Graceful.
3. **Behavior note:** the native wrapper *ignores* `allowRemoteContexts` (`_allowRemoteContexts` param unused) — this matches pre-migration prod behavior (native was live then), but flagging for awareness on remote-context verification flows.
4. **Perf claim to verify post-deploy:** compare CloudWatch Duration on signing/did routes before/after — this PR is only worth its infra if that delta is real.

---

### ✨ PR Description
Purpose: Deploy native DIDKit binary to Lambda via layers instead of bundling, fixing silent fallback to WASM and Bun workspace module resolution issues.

Main changes:
- Added error logging to DIDKit initialization in three services, exposing fallback behavior that was previously silent
- Created `prepare-didkit-native-layer.cjs` script to assemble Lambda Layer with native plugin binaries and dist files
- Updated serverless configs to exclude didkit-plugin-node from bundles, attach native layer to all functions, and removed broken node_modules copy patterns

### DIFF
--- a/.changeset/shy-spies-strive.md
+++ b/.changeset/shy-spies-strive.md
@@ -1,0 +1,7 @@
+---
+"@learncard/network-brain-service": patch
+"@learncard/lca-api-service": patch
+"@learncard/learn-cloud-service": patch
+---
+
+✨ Ship native DIDKit to Lambda via a Layer (follow-up to #1341)
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,9 @@ internal-docs/*
 .worktrees/
 CLAUDE.local.md
 
+# Assembled at deploy time by scripts/prepare-didkit-native-layer.cjs
+services/learn-card-network/*/didkit-native-layer/
+
 # Local Netlify folder
 .netlify
 # Local environment configuration
--- a/scripts/prepare-didkit-native-layer.cjs
+++ b/scripts/prepare-didkit-native-layer.cjs
@@ -1,0 +1,66 @@
+#!/usr/bin/env node
+/**
+ * Assembles a Lambda Layer directory containing @learncard/didkit-plugin-node so the
+ * native DIDKit addon is resolvable in Lambda.
+ *
+ * Why a layer: the native wrapper locates its binary via
+ * `require.resolve('@learncard/didkit-plugin-node/package.json')` + readdir, so the
+ * package must exist as a real, resolvable package on the module path — it cannot be
+ * bundled by esbuild or dropped next to the handler (unlike the WASM). Lambda mounts
+ * layers at /opt and puts /opt/nodejs/node_modules on NODE_PATH, which satisfies that.
+ *
+ * Run from a service directory before `serverless deploy` (see the service's
+ * `serverless-deploy` npm script). Output: <service>/didkit-native-layer/nodejs/
+ * node_modules/@learncard/didkit-plugin-node/ — referenced by the `layers:` block in
+ * the service's serverless.yml.
+ *
+ * The compiled dist/ only requires its own relative files + node builtins (the
+ * @learncard/* imports are type-only), so the layer needs just this one package.
+ * In CI, the "Build Native DIDKit Plugin" job's artifact (dist/ + index.linux-x64-gnu.node)
+ * is downloaded into the package root before deploy. Locally the .node binary is
+ * usually absent — the layer then ships without it and the runtime falls back to WASM,
+ * same as before this change.
+ */
+const fs = require('node:fs');
+const path = require('node:path');
+
+const serviceDir = process.cwd();
+const layerRoot = path.join(serviceDir, 'didkit-native-layer');
+const layerPkgDir = path.join(layerRoot, 'nodejs', 'node_modules', '@learncard', 'didkit-plugin-node');
+
+const packageRoot = path.dirname(
+    require.resolve('@learncard/didkit-plugin-node/package.json', { paths: [serviceDir] })
+);
+
+fs.rmSync(layerRoot, { recursive: true, force: true });
+fs.mkdirSync(layerPkgDir, { recursive: true });
+
+fs.copyFileSync(path.join(packageRoot, 'package.json'), path.join(layerPkgDir, 'package.json'));
+
+const distDir = path.join(packageRoot, 'dist');
+if (!fs.existsSync(distDir)) {
+    console.error(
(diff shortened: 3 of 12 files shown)


goblincore:
Good catch on the leftover `package.patterns` globs — addressed in f16e4b1.

**Why it mattered:** today those globs match nothing (service-relative paths, and Bun workspaces hoist everything to the repo root — that gap is the #1341 root cause). But they were a landmine: if a service-level `node_modules` ever reappeared (installer change, tooling shift), the force-include would put a copy of `didkit-plugin-node` into `/var/task/node_modules`, which wins over `NODE_PATH` and silently shadows the Layer — the exact failure mode this PR exists to prevent, and it would fail *silently* (native "works" but from a stale/binary-less copy, or falls back to WASM with no signal beyond the new warn log).

Removed the didkit-plugin-node globs **and** the equally-dead WASM glob (the wasm ships inside the bundle via `esbuildPlugins.cjs` since #1341), left `'!node_modules/**'` (+ lca's `src/swagger-ui`) in place, and added a comment explaining the history — plausible-looking dead config is what fueled the original 5-commit symptom chase, so the comment is there to stop the next responder from re-adding these.

TaylorBeeston (approved):

goblincore:
## ✅ Post-deploy verification complete — native DIDKit confirmed live on staging

- Deploy run succeeded for all three services; **Smoketest Staging green** with the layer attached.
- CloudWatch Logs Insights across all three `*-dev-didWeb` log groups (1h window, with fresh DIDKit-touching traffic generated): **zero `[didkit] native plugin unavailable` fallback warns**.
- Positive control: **7 cold starts** in the same window (several clearly post-deploy) — each fresh instance served did.json requests, so a native load failure would have warned. Silence = native loaded.
- All staging probes 200 post-deploy (brain/cloud did.json, LCA + brain health).

Remaining follow-up from the PR notes: eyeball the wasm-era (≈15:20–15:50 UTC today) vs native-era request `Duration` on the didWeb lambdas for the perf delta.

goblincore:
## 📉 Perf follow-up closed — measured wasm vs native on staging (didWeb, 12h window)

| Era (UTC today) | worst_ms | avg_ms |
|---|---|---|
| WASM era (15:20–15:50, hotfix only) | **~1,920** | ~300 |
| Native era (post-15:55, layer live) | **~130** | ~150 |

The ~1.9s worst-case is the per-instance WebAssembly compile on first DIDKit request — the exact cost this layer was built to remove. **~14× improvement on cold-instance did:web resolution**, avg roughly halved. Small sample (22 reqs), but the worst-case signal is mechanistic, not statistical.

Layer: deployed ✅ · native confirmed loading ✅ · perf delta measured ✅

PR #1303, opened 2026-06-12
Migrate monorepo development to Bun workspaces

TaylorBeeston:
https://www.loom.com/share/92b2fe81fa2a434280dbe9ba8a52ad81

# Overview

#### 🎟 Relevant Jira Issues

N/A — no Jira ticket for this migration.

#### 📚 What is the context and goal of this PR?

Move the monorepo from pnpm-managed installs and build-first local development to Bun workspaces and Bun-driven TypeScript source development.

The goal is to keep publish-time builds intact for npm artifacts while making local LearnCard App, Brain Service, Cloud Service, LCA API, and CLI development run against workspace TypeScript directly.

#### 🥴 TL; RL:

- Replaces pnpm workspace metadata/lockfile with Bun workspace metadata and `bun.lock`.
- Updates CI, Docker, Netlify configs, scripts, and docs to install/run with Bun.
- Adds `development` export conditions so local dev resolves workspace package source instead of built `dist`.
- Simplifies local Docker dev by removing package build/watch services and mounting workspace packages directly.
- Fixes source-mode issues found during proof: Vite CSS import/style handling, ESM `require` usage, Brain Service model cycle, CLI Bun runtime path, and Docker HMR mounts.
- Keeps npm publish builds in place and preserves npm Trusted Publishing/OIDC instead of long-lived npm tokens.

#### 💡 Feature Breakdown (screenshots & videos encouraged!)

- Root workspace/package-manager migration:
    - Adds Bun workspaces to root `package.json`.
    - Adds `bunfig.toml` with hoisted linker.
    - Replaces `pnpm-lock.yaml` / `pnpm-workspace.yaml` with `bun.lock`.
- Local Docker development:
    - `apps/learn-card-app/compose-local.yaml` now runs services from TS source with `bun --conditions=development`.
    - Removes build watcher complexity from local app dev.
    - Mounts `packages/` and root `node_modules` into app/service containers so dependent package edits hot-reload without rebuilding images.
- Package source-mode imports:
    - Adds development export conditions across workspace packages.
    - Keeps normal package exports pointed at built `dist` for published consumers.
- LearnCard App/Vite:
    - Excludes workspace packages from dependency prebundling during dev.
    - Imports React package source CSS explicitly so Tailwind utilities from `@learncard/react` are generated correctly.
- CLI:
    - `packages/learn-card-cli` dev/start paths run TS directly with Bun.
    - Adds a Bun-compatible REPL fallback for dev source execution.
- CI/deployment:
[...]

PR #1355, opened 2026-07-03
fix: Tanstack Resolution

Custard7:
## What

The LearnCard App white-screened on staging (built from `main`) with `No QueryClient set, use QueryClientProvider to set one`, thrown from `AuthCoordinatorProvider`. This fixes the underlying build issue so the app boots.

## Why

`AuthCoordinatorProvider` is correctly nested inside `PersistQueryClientProvider`, so this was not a provider-ordering bug. The real cause was `@tanstack/react-query` ending up as **two separate module instances** in the production bundle — one backing the provider, a different one backing `useQueryClient()`. Since React Context identity is per-module-instance, the hook couldn't find the client.

Two things combined to cause this, and it only showed up in a real `vite build` (not the dev server, which is why local dev looked fine):

- `learn-card-base` imports react-query in ~20 files but never declared it as a dependency.
- The app aliases `learn-card-base` to its raw TypeScript source, so those bare `@tanstack/react-query` imports get pulled into the app's own bundle. Without a dedupe pin, Rollup could resolve them to a second copy.

The recent Bun workspaces migration (#1303) changed hoisting enough to tip this over into duplication.

## Changes

- `apps/learn-card-app/vite.config.ts` — added `@tanstack/react-query` to `resolve.dedupe` so the build always resolves a single instance (with a comment explaining why, to prevent regression).
- `packages/learn-card-base/package.json` — declared `@tanstack/react-query` as a `peerDependency` (and `devDependency` for local resolution), since it uses the hooks but never declared them.
- `bun.lock` — lockfile update. Most of the churn is unrelated version bumps picked up from `main`; the relevant change is the new react-query entry under `learn-card-base`.

## How to review

The bug only reproduces in a production build, not the dev server. To verify:

1. `bun install`
2. `cd apps/learn-card-app && bun run build`
3. `bunx vite preview --outDir build` and confirm the app loads past the initial screen (no `No QueryClient set` error in the console).

To confirm the dedupe worked, check that `build/assets/` contains exactly one `vendor-tanstack-*.js` chunk and that the react-query internals live only in that chunk.

### ✨ PR Description
Purpose: Resolve duplicate React Query instances by pinning @tanstack/react-query in build and package dependencies to prevent QueryClient context errors.

Main changes:
- Added @tanstack/react-query to Vite optimize.include to ensure single resolved instance during prod build
- Declared @tanstack/react-query as peerDependency and devDependency in learn-card-base package.json

---

### DIFF
--- a/.changeset/salty-parrots-bow.md
+++ b/.changeset/salty-parrots-bow.md
@@ -1,0 +1,6 @@
+---
+"learn-card-app": patch
+"learn-card-base": patch
+---
+
+fix: Tanstack Resolution
--- a/apps/learn-card-app/vite.config.ts
+++ b/apps/learn-card-app/vite.config.ts
@@ -229,6 +229,11 @@ export default defineConfig(async ({ mode, command }) => {
                 'react-router-dom',
                 'history',
                 '@ionic/react-router',
+                // Required: learn-card-base is aliased to raw TS source and imports
+                // react-query with bare specifiers. Without this pin the prod build
+                // resolves two react-query instances, splitting the QueryClient React
+                // Context and throwing "No QueryClient set" at runtime.
+                '@tanstack/react-query',
             ],
         },
         server: {
--- a/bun.lock
+++ b/bun.lock
@@ -64,7 +64,7 @@
     },
     "apps/learn-card-app": {
       "name": "learn-card-app",
-      "version": "1.96.3",
+      "version": "1.96.5",
       "dependencies": {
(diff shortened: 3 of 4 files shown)


Custard7:
Addressed in eb0cbcad5 — widened the `@tanstack/react-query` peerDependency to `^5.0.0`. There's no 5.80-specific API in use here, so the broader v5 range is correct. Kept the devDependency pinned for local install.

PR #1356, opened 2026-07-03
feat: Android Auth + Offline Resiliency

Custard7:
# Overview

https://www.loom.com/share/a8e8d71a98384884b450d8ef89ef9d36

Makes the native apps usable on airplane mode and flaky connections. Today, opening the app with no connection can hang on the splash forever, drop you into a half-broken "logged in but nothing works" state, or (on Android) throw a Firebase error modal you can't get past. This PR makes boot never hang, lets returning users into the app offline with their real wallet, shows clear offline states the user can tap to reconnect, and recovers data automatically when the connection comes back.

## What changed and why

**1. Boot can no longer hang on the network.**
Every network call during startup (auth session check, key-status fetch, wallet build) now has a hard timeout. If a call stalls, boot moves on instead of spinning forever. See `withDeadline` in `packages/learn-card-base/src/helpers/withDeadline.ts`.

**2. Network errors at boot are treated as "offline", not fatal.**
A connection failure during startup (e.g. Firebase `auth/network-request-failed`) used to become a hard error — an error modal plus a retry loop. It now resolves to the normal logged-out/offline state and re-runs cleanly when the connection returns.

**3. Returning users get into the app offline — with their real wallet.**
If the networked wallet can't be built (offline or timed out), we build a fully-local wallet so the app still loads, then quietly upgrade to the networked wallet when the connection is back. This required fixing several real bugs found while testing on Android:
- The private key wasn't reliably saved (the SQLite write didn't wait for the DB and could silently no-op).
- The key read back was flaky on cold start (transient empty reads / picking the wrong row) — now retries and prefers a row that actually holds a key.
- Deriving the wallet's ID offline failed because the bundled DIDKit crypto file wasn't passed to the SDK, so it tried to fetch it over the network. Now passed explicitly.
- See `packages/learn-card-base/src/security/platformPrivateKeyStorage.ts` and `helpers/walletHelpers.ts`.

**4. Onboarding prompts can't misfire offline.**
The shared auth-gate selector treats "no profile" as a genuine absence only when a full networked wallet is up and the device is online. On the local/offline wallet, the network's `getProfile` returns nothing without erroring — which previously looked like "no profile" and wrongly triggered the Complete Profile / age-gate / join-network flow. It now resolves to "couldn't confirm" instead.

**5. Data recovers automatically on reconnect.**
React Query's online state is now driven by our connectivity model (fed by Capacitor Network) instead of the browser's `navigator.onLine`, which is unreliable in the Android webview. So when the connection returns, queries refetch and paused mutations resume — data that went stale offline (notifications, etc.) reloads on its own. See `apps/learn-card-app/src/FullApp.tsx`.

**6. Clear offline states the user can act on.**
- A slim banner appears when offline (or running the local wallet); tapping it re-checks the connection and upgrades the wallet — no reload. See `components/network-listener/OfflineBanner.tsx`.
- Logged-out users who open offline see a calm "You're offline" screen whose Try Again re-runs boot. See `OfflineBootGate.tsx`.
- **Embedded partner apps** (the app listings that open an external site in an in-app frame) now show an on-brand offline state instead of a broken frame or a raw browser "can't connect" error. It triggers when the device is offline, and also when the app fails to load or times out (~12s) while "online but unreachable". "Try Again" re-checks the connection and reloads the app. See `pages/launchPad/AppEmbedOfflineState.tsx`, wired into `EmbedAppFullScreen.tsx` and `EmbedIframeModal.tsx`.

**7. Reusable connectivity hooks + dev tooling.**
`useIsOffline()` / `useOnReconnect()` are exported from `learn-card-base` for any component that needs to react to connectivity. A new **Offline** tab in the Dev Debug panel adds: a connectivity flag to preview offline UI, a fault injector that stalls/fails the real boot calls (reproduce airplane-mode behavior in a normal browser/simulator), and a Key Persistence readout showing whether the private key is actually on disk.

## Types of changes

-   [x] Bug fix (boot hang, offline error modal/loop, private-key persistence, false onboarding)
-   [x] New feature (offline wallet fallback, reconnect UX, offline states for embedded apps)

# Testing

## How to QA this

**On a real device / simulator (the important one):**
1. Log in while online, then fully quit the app.
2. Turn on airplane mode and open the app. Expected: it boots to your wallet in offline mode (offline banner) — no endless loader, no error modal, and it does NOT try to make you sign up / show the age gate.
3. Turn airplane mode off and tap the banner's **Reconnect**. Expected: it reconnects, the banner clears, and stale data reloads.
4. Logged out + airplane mode → open app: expect the "You're offline" screen; Try Again re-runs boot.
5. Open an installed app listing (embedded partner app) while offline → expect the on-brand "You're offline" state, not a broken frame. Tap Try Again after reconnecting to load it.

**Using the Dev Debug panel (no airplane mode needed):**
1. Bug button (bottom-right) → **Offline** tab.
2. **Fault Injector** → **Delay**/**Fail**, then reload → app still loads (local wallet), no hang.
3. **Key Persistence** → while logged in, **Check disk key** → expect "found".
4. Set the injector back to **Off** for normal boot.

## Devices to test on

Android Native / iOS Native / Chromium

## Code coverage

Unit tests for the timeout helper, the auth-gate selector (exhaustive, incl. offline + local-wallet cases), and the coordinator's offline/network-error handling. Run: `bunx vitest run src/auth-status src/auth-coordinator src/helpers/withDeadline.test.ts` in `packages/learn-card-base`.

# Documentation

Internal changes to app boot behavior — no user-facing docs needed. Debug-panel usage is self-explanatory in the UI.

## Notes

-   A generic **image** fallback (broken-image placeholders + image caching, including a CSS `img::before`/`::after` baseline and an `onError` component for iOS) is intentionally left for a follow-up PR.
-   The same offline wiring still needs to be mirrored into the Scouts app in a follow-up.

### ✨ PR Description
Purpose: Implement offline-resilient Android authentication with deadline-bounded boot calls and wallet fallback to prevent indefinite hangs during poor connectivity.

Main changes:
- Added deadline utilities and network fault injection for boot-time calls; AuthCoordinator now times out opportunistic auth checks and server fetches gracefully
- Implemented offline wallet fallback that rebuilds on reconnect; platform private key persistence now retries SQLite reads to survive cold-start races
- Created connectivity and wallet-mode stores bridging React Query's online state; UI shows offline banner, boot gate, and debug panels for testing resilience


gerardopar (approved):
tested the following scenarios

1. toggled wifi on/off + background app
2. toggled airplane mode on/off + background app
3. toggled airplane mode on/off + cold start app 

### Cold Start
https://github.com/user-attachments/assets/dda27529-5bcf-40fa-8153-3d901c82d443

### Airplane Mode
https://github.com/user-attachments/assets/61ac0c36-b064-4314-a654-2829aafb6186

### Wifi off
https://github.com/user-attachments/assets/eefeb286-c2cc-4fb3-8c28-71cd22eb5593

PR #1360, opened 2026-07-07
chore: [LC-1905] Upload VC resiliency in Build My LearnCard

goblincore:
# Overview

#### 🎟 Relevant Jira Issues

Fixes: [LC-1905](https://welibrary.atlassian.net/browse/LC-1905)

The **Add Credentials** screen in Build My LearnCard lets a user upload or paste a raw JSON Verifiable Credential. The existing flow was brittle: the file picker only accepted `.json`, and both the paste box and the file path funneled into a plain `JSON.parse` that expected either an OpenBadge v2 object or a single VC. A `VerifiablePresentation` (VP) that wraps one or more VCs — the common export shape from other wallets — failed and could not be imported.

This PR makes the upload as forgiving and resilient as possible, making a best effort to extract usable credentials from whatever the user provides.

#### 🥴 TL; RL:

Upload now accepts `.txt` files, forgivingly extracts JSON (even embedded in surrounding text), and imports a `VerifiablePresentation` that wraps one or more VCs — adding the good ones and reporting the rest.

Screenshot 2026-07-06 at 11 13 56 PM

https://github.com/user-attachments/assets/4a14b25e-1934-473d-b126-72b5af296e5f

- **New pure module `extractCredentialsFromText.ts`** — owns all parsing/unwrapping so it's independently unit-testable:
    - trim → `JSON.parse` → on failure, substring-scan for the first balanced `{…}`/`[…]` block (string/escape-aware, so braces inside string literals don't truncate it).
    - Classify: a `VerifiablePresentation` (`type` includes `VerifiablePresentation`, or a `verifiableCredential` field is present) is unwrapped to its `verifiableCredential`(s), normalized to an array; anything else passes through as a single candidate (single VC or OpenBadge v2, unchanged).
    - Never throws — returns `{ credentials, errors }`.
- **`useUploadVcFromText` refactor** — the per-credential store logic is extracted verbatim into `uploadSingleCredential`; `uploadVcFromText` now extracts candidates, uploads each, and returns an `AggregateUploadResult` (`addedCount` / `failedCount` / per-item `results` / `errors`). **Partial success**: good VCs are stored even when a sibling fails. `validateTextVC` (the live paste-box check) no longer false-flags a VP.
- **`useUploadFile.getJsonFiles`** — flattens multi-VC-per-file results and counts credentials (not files) for the success/partial/failure toast.
- **`CheckListUploadRawVC`** — picker now `accept=".json,.txt"`; paste + upload surface multi-VC and partial-success messaging.

- **Scope kept tight (YAGNI):** top-level bare arrays with no VP wrapper, and nested VPs, are intentionally out of scope. A VP's inner `verifiableCredential` may be a single object or an array, so the array case is covered by VP handling.
- **Imported credentials are stored as received** (already-signed VCs are not re-issued/re-signed).
- **Best-effort over strict:** the substring-scan means we favor extracting a credential from noisy input over rejecting it.

#### 🔍 Types of Changes

-   [x] No
-   [ ] Yes

# Testing

1. Open Build My LearnCard → **Add Credentials**.
2. **Paste path:** paste the contents of the attached `SharedCredential.txt` (a VP from MIT's Learner Credential Wallet wrapping an OpenBadge v3 VC) into the paste box → **Add**. The "Digital Credential Explorer" credential should appear in your wallet with a success toast. 
See JIRA for the sample artifact you can use for testing (credential in txt file):

3. **Upload path:** click **Upload** and select the same file as a `.txt` → same result.
4. **Partial success:** craft a VP with one valid VC and one broken entry (e.g. `{}`), upload it → the valid one is added and a partial-success toast reports the failure.
5. **Resilience:** paste JSON with some surrounding prose (e.g. `Here's my credential: {…}`) → it still extracts.

Chromium (LearnCard App)

#### 🧪 Code Coverage

Unit tests for the pure `extractCredentialsFromText` module (9 tests, vitest), including the **real `SharedCredential.txt` acceptance fixture** (VP → single OpenBadge v3 VC), VP-with-array, VP-with-single-object, bare single VC, JSON-embedded-in-prose, brace-inside-string, empty input, no-JSON input, and empty-VP. The wallet-touching upload paths are covered by the existing manual/e2e flows (unchanged store logic, just looped).

# Documentation

**User-Facing Docs**

-   [ ] Tutorial / How-To / Reference / Concept / App Flows — n/a

**Internal/AI Docs**

-   [ ] AGENTS.md
-   [x] Code comments/JSDoc — added JSDoc to the extraction module explaining the parse/scan/classify pipeline.

**Visual Documentation**

-   [ ] Mermaid diagram

#### 💭 Documentation Notes

No public API or SDK surface changed — this is an app-internal resiliency improvement to an existing UI flow. No user-facing docs required.

# ✅ PR Checklist

-   [ ] I have **manually tested** common end-2-end cases — ⚠️ unit-tested + typechecked; in-app manual smoke still recommended by a reviewer (see QA steps)
-   [ ] I have considered **product analytics** for user-facing features — no new analytics events added

[LC-1905]: https://welibrary.atlassian.net/browse/LC-1905?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### ✨ PR Description
Purpose: Enhance VC upload resiliency by extracting credentials from various JSON formats, supporting partial uploads, and displaying user-friendly error messages.

Main changes:
- Implement `extractCredentialsFromText` utility to resilient extract VCs from text, handle VerifiablePresentation unwrapping, and scan for JSON in prose
- Refactor `uploadVcFromText` to support batch credential processing with per-credential validation and aggregate success/failure counts
- Replace raw validation errors with friendly user-facing copy while logging technical details; accept `.txt` file uploads


apps/learn-card-app/src/hooks/useUploadFile.tsx
```diff
@@ -228,58 +228,66 @@ export const useUploadFile = (uploadType: UploadTypesEnum) => {
...
-                            log.error('Error uploading file:', err);
-                            throw err;
-                        }
-                    } catch (innerErr) {
-                        failedUploads++;
-                        log.error('❌ Error processing file:', innerErr);
-                        return null;
+                            });
+
+                        if (aggregate.errors.length > 0) onFail?.(aggregate.errors);
```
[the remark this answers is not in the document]
goblincore: Fixed in 73fba98 — `onFail` now only fires when `aggregate.addedCount === 0`, so a partially-successful file no longer shows the inline error box alongside the partial-success toast.

goblincore:
Addressed the gitStream/LinearB review in `73fba98`:

**Issue 1 — double error feedback (fixed).** `getJsonFiles` now guards `onFail?.(aggregate.errors)` with `&& aggregate.addedCount === 0`, so a partially-successful file surfaces only the summary toast, not the inline error box as well.

**Issue 2 — validation bypass (fixed).** Added `validateCredentialObject(parsed)`, which validates an already-parsed credential directly (OpenBadge-v2 check + `VCValidator`). `uploadSingleCredential` and `uploadVcFromTextAndAddToWallet` now call it instead of `validateTextVC`, so an object is no longer `JSON.stringify`'d and re-run through `extractCredentialsFromText` — closing the path where a malformed outer object wrapping a valid inner VC could have its children validated in its place. `validateTextVC` remains the text/paste entry point and delegates per-candidate to the same helper.

**Issue 3 — findFirstJsonBlock depth counter (not a bug).** The concern was that `}` inside objects would decrement depth against a top-level `[`. It can't: when `open === '['`, `close === ']'`, and `{`/`}` match neither the `open` nor the `close` branch, so they're skipped entirely. `}` only affects depth when `close === '}'` (i.e. `open === '{'`). Rather than change correct code, I added two regression tests locking the behavior in: a VP-in-prose whose `verifiableCredential` is a nested array of objects, and a top-level array of objects embedded in prose — both extract fully without truncation.

All 11 unit tests pass (9 prior + 2 new). Also added a `learn-card-app: patch` changeset to clear the changeset check.

goblincore:
@claude please review

apps/learn-card-app/src/hooks/extractCredentialsFromText.ts
```diff
@@ -0,0 +1,100 @@
...
+        const block = findFirstJsonBlock(trimmed);
+        if (block === null) {
+            return { credentials: [], errors: ['Could not find valid JSON in the provided text.'] };
+        }
+        try {
+            parsed = JSON.parse(block);
+        } catch (err: any) {
+            return {
+                credentials: [],
+                errors: [`Could not parse the JSON found in the provided text: ${err.message}`],
```
TaylorBeeston: This fallback only tries the first balanced `{...}` / `[...]` block. If the surrounding prose has any earlier bracketed text, the valid credential later in the input is never considered.

Repro I ran locally against this branch:

```ts
extractCredentialsFromText(`from [wallet export]: ${validVcJson}`)
// => { credentials: [], errors: ["Could not parse..."] }
```

That undercuts the new JSON-in-prose resilience path. Please keep scanning after a candidate block fails to parse/classify, and only return failure after all balanced candidates are exhausted. A robust version should also avoid surfacing `err.message` directly to the user; log the raw parser detail and return friendly copy.

apps/learn-card-app/src/components/learncard/checklist/checklist-steps/CheckListUploadRawVC.tsx
```diff
@@ -176,15 +176,32 @@ export const CheckListUploadRawVC: React.FC = () => {
...
+                        type: ToastTypeEnum.Success,
+                        hasCheckmark: true,
+                        autoDismiss: false,
+                    }
+                );
             } else {
-                setRawTextErrors([`Failed to parse JSON VC. ${result?.error}`]);
+                setRawTextErrors(
+                    result?.errors?.length ? result.errors : ['Failed to parse JSON VC.']
+                );
```
TaylorBeeston: `result.errors` now includes raw parser / validation / storage messages from the helper layer, and this component renders them directly in the user-facing error box. That can expose messages like `Unexpected identifier`, Zod field paths, or `Failed to add credential to index: ${error.message}`.

Per the app UI guidance, keep detailed failures in logs and map the UI to stable friendly copy. For this paste path, something like this is safer:

```suggestion
                setRawTextErrors(['We could not add that credential. Please check the text and try again.']);
```

apps/learn-card-app/src/hooks/useUploadFile.tsx
```diff
@@ -228,58 +228,70 @@ export const useUploadFile = (uploadType: UploadTypesEnum) => {
...
-                        failedUploads++;
-                        log.error('❌ Error processing file:', innerErr);
-                        return null;
+                            });
+
+                        // Only surface inline file errors when nothing was added from this
+                        // file. On partial success the summary toast already reports failures,
+                        // so avoid double-signaling (inline error box + toast).
+                        if (aggregate.errors.length > 0 && aggregate.addedCount === 0)
+                            onFail?.(aggregate.errors);
```
TaylorBeeston: Same raw-error issue on the file path: `aggregate.errors` is rendered by the caller as inline file errors, but it can contain parser internals and storage/index exception messages. Please log the raw details and pass friendly UI copy through `onFail`.

```suggestion
                        if (aggregate.errors.length > 0 && aggregate.addedCount === 0)
                            onFail?.(['We could not add that credential. Please check the file and try again.']);
```

TaylorBeeston (changes requested):
(OMP made this whole review, apparently it attempted to LARP as me or something idk lol)

Thanks for the resiliency work here — the VP unwrapping and aggregate upload shape look solid.

I found two user-facing issues that are worth tightening before merge: the fallback scanner stops after the first non-JSON bracketed block, and raw parser/storage errors can now reach the UI. Left inline comments with concrete repro/suggestions.

apps/learn-card-app/src/hooks/extractCredentialsFromText.ts
goblincore: Fixed in 24504a0c. `extractCredentialsFromText` now scans every balanced JSON block left-to-right (via an `iterateJsonBlocks` generator) and returns the first block that yields usable credentials, instead of bailing after the first block fails to parse. Your `from [wallet export]: {vc}` repro now resolves — added it as a regression test, plus one for an *unbalanced* earlier bracket and one for an empty-VP-then-valid-VC. Also stopped surfacing `err.message`: the scanner just `continue`s past unparseable blocks and returns stable friendly copy.

apps/learn-card-app/src/components/learncard/checklist/checklist-steps/CheckListUploadRawVC.tsx
goblincore: Fixed in 24504a0c. The paste path now logs the raw reasons (`log.error('Paste VC upload failed', result?.errors)`) and renders stable copy — 'We could not add that credential. Please check the text and try again.' The live `validateTextVC` useEffect and the catch branch were leaking Zod paths / `error.message` too, so both were switched to the same friendly constant.

apps/learn-card-app/src/hooks/useUploadFile.tsx
goblincore: Fixed in 24504a0c. Both `getJsonFiles` failure paths (the `addedCount === 0` branch and the per-file catch) now `log.error` the raw `aggregate.errors`/exception and pass friendly copy through `onFail` — 'We could not add that credential. Please check the file and try again.' No parser/index internals reach the inline error box.

goblincore:
@TaylorBeeston Had claude address the issues above and reply, so should be ready for rereview!

apps/learn-card-app/src/hooks/extractCredentialsFromText.ts
```diff
@@ -0,0 +1,144 @@
...
+ * Flatten a parsed JSON value into individual credential candidates:
+ * - a VerifiablePresentation → its verifiableCredential(s)
+ * - a bare array → each element, flattened (VP elements are unwrapped too)
+ * - a single object → itself
+ * Non-object values (strings, numbers, null) yield nothing.
+ */
+const collectCandidates = (parsed: any): any[] => {
+    if (isVerifiablePresentation(parsed)) return unwrapPresentation(parsed);
+    if (Array.isArray(parsed)) return parsed.filter(Boolean).flatMap(collectCandidates);
+    return parsed && typeof parsed === 'object' ? [parsed] : [];
```
TaylorBeeston: This still stops the scan too early when the first parseable block is ordinary JSON metadata rather than a credential. `collectCandidates` treats any object as usable, so `extractCredentialsFromText` returns that metadata object and never reaches a valid VC later in the text. The upload path then validates only the metadata object and fails.

Repro I ran against `24504a0c` / current PR worktree:

```ts
extractCredentialsFromText(`metadata: {"source":"x"} credential: ${validVcJson}`)
// => { credentials: [{ source: "x" }], errors: [] }
// validVcJson later in the same input is never considered
```

This is the same class of issue as the previous comment: keep scanning after blocks that do not yield a credential candidate. Either make this classifier stricter (VP, OBv2, or VC-like object only) or have the upload path continue to later extracted blocks when validation rejects an earlier candidate. Please add a regression test for parseable non-credential JSON before a valid VC.

TaylorBeeston (changes requested):
Re-reviewed the fix. The raw-error UI issue is addressed, and the original `[wallet export]` repro now works.

One scanner edge remains: extraction still stops before the valid credential when the earlier noisy block is parseable JSON but not a credential. Left an inline repro.

apps/learn-card-app/src/hooks/extractCredentialsFromText.ts
goblincore: Fixed in c73a86a18. `collectCandidates` now gates lone objects through a `looksLikeCredential` check — the object must carry at least one of `@context` / `type` / `credentialSubject` / `verifiableCredential` / `proof`. So a parseable-but-non-credential block like `{"source":"x"}` yields no candidate, and the scanner keeps going and returns the real VC later in the text. Your `metadata: {...} credential: {vc}` repro now resolves to the vc; non-credential elements are likewise dropped from bare arrays. Added three regression tests (metadata-before-VC, non-credential array element, lone non-credential object). The gate is intentionally lenient — every real VC/VP/OpenBadge carries one of those fields — so it filters junk without risking false rejections of valid credentials.

goblincore:
@TaylorBeeston Addressed the re-review in `c73a86a18` (rebased onto the latest `main` merge).

**Scanner stopped at the first parseable non-credential block** (e.g. `metadata: {"source":"x"} credential: {vc}` returned the metadata). `collectCandidates` now gates lone objects through `looksLikeCredential` — the object must carry at least one of `@context` / `type` / `credentialSubject` / `verifiableCredential` / `proof`. Non-credential JSON yields no candidate, so the scanner keeps going and returns the real credential later in the text; junk elements are also dropped from bare arrays. The gate is intentionally lenient (every real VC/VP/OpenBadge has one of those fields) so it filters noise without risking false rejections of valid credentials.

Added 3 regression tests (metadata-before-VC, non-credential array element, lone non-credential object) — 21/21 unit tests pass.

That covers all three threads from the last review (raw-error UI, `[wallet export]` scan, and this parseable-metadata edge). Thanks for the thorough passes — ready for a re-request whenever.

TaylorBeeston (changes requested):
Sorry, OMP is being too nitpicky! LGTM now!

TaylorBeeston (approved):

PR #1378, opened 2026-07-09
fix: App page tile icons + More Apps launch behavior (LC-1928)

goblincore:
## Summary

BEFORE:

Screenshot 2026-07-09 at 12 06 57 AM

AFTER:

Screenshot 2026-07-09 at 12 07 00 AM

- **Fix tiny LearnCard app icons at tablet / in-between widths.** The launchpad grid stayed at 3 columns until `md` (768px) while every tile was capped at `max-w-[100px]`, so between ~400 and 767px the tiles sat tiny inside much wider grid columns and the icons looked lost. Cap the tile at `max-w-[160px]` at every width (not just desktop) and move the 3→4 column breakpoint from `md` down to `sm` (640px), so the original `p-[18%]` icon padding stays proportional across all viewports.
- **More Apps tiles now launch installed apps directly instead of always opening the info page.** Previously every tile in the More Apps section opened `AppStoreDetailModal`. Now installed apps launch on tap — mirroring the existing "Open" button on the browse list — and only not-installed / suggested apps still open the detail page (so the user can install first).
- **Refactor: extract shared launch flow into a `useAppLaunch` hook.** The launch logic (consent redirect + delegate VP, AI tutor, embedded iframe, direct / second-screen links, age hard-block, fallback) was already duplicated in `AppStoreListItem` and `AppStoreDetailModal`. Rather than add a third copy for the new `MoreAppTile`, extract the `AppStoreListItem` flavor into `useAppLaunch` and have both `AppStoreListItem` and `MoreAppTile` consume it.

## Test plan

- [ ] Desktop (≥992px, sidebar visible): LearnCard app tiles + icons look proportional and match the design.
- [ ] Tablet / in-between (~400–991px, no sidebar): LearnCard app icons are no longer tiny inside oversized tiles.
- [ ] Mobile (<400px): LearnCard tiles look the same as before this PR.
- [ ] Tap a LearnCard shortcut tile → navigates to that feature (`Skill Insights`, `Pathways`, etc.).
- [ ] Tap an installed app tile in **More Apps** → **launches the app** (opens the embed/redirect/etc, not the detail modal).
- [ ] Tap a not-installed / suggested app tile in **More Apps** → opens the app detail / install page.
- [ ] Age-restricted apps still show the hard-block modal on tap (unchanged behavior).
- [ ] Browse list ("Browse More +" → app store) is unchanged: Get / Install / Open buttons still work.

### ✨ PR Description
Purpose: Extract app launch logic into reusable hook to fix tile icons and enable consistent launch behavior across app grid and store list.

Main changes:
- Created `useAppLaunch` hook encapsulating consent-flow, AI Tutor, iframe, and direct-link launch paths with age-gating
- Refactored `AppStoreListItem` to use new hook, removing 246 lines of embedded launch logic
- Updated `AppGrid` responsive breakpoints (3-col phone → 4-col tablet) and created `MoreAppTile` component using shared launch hook

### DIFF
--- a/.changeset/calm-poets-marry.md
+++ b/.changeset/calm-poets-marry.md
@@ -1,0 +1,5 @@
+---
+"learn-card-app": patch
+---
+
+fix: App page tile icons + More Apps launch behavior 
--- a/apps/learn-card-app/src/pages/launchPad/AppStoreListItem.tsx
+++ b/apps/learn-card-app/src/pages/launchPad/AppStoreListItem.tsx
@@ -1,21 +1,12 @@
-import React, { useMemo } from 'react';
+import React from 'react';
 import type { AppStoreListing, InstalledApp } from '@learncard/types';
 
 import { IonItem, IonSpinner } from '@ionic/react';
-import { useModal, ModalTypes, useWallet } from 'learn-card-base';
-import { ShieldAlert } from 'lucide-react';
 import { ThreeDotVertical } from '@learncard/react';
 
 import useTheme from '../../theme/hooks/useTheme';
 import { ColorSetEnum } from '../../theme/colors';
-import AppStoreDetailModal from './AppStoreDetailModal';
-import { EmbedIframeModal } from './EmbedIframeModal';
-import { useConsentFlowByUri } from '../consentFlow/useConsentFlow';
-import GuardianConsentLaunchModal from './GuardianConsentLaunchModal';
-import AiTutorConnectedView from './AiTutorConnectedView';
-import { useGuardianGate } from '../../hooks/useGuardianGate';
-import { checkAppInstallEligibility } from '@learncard/helpers';
-import { openExternalLink } from '../../helpers/externalLinkHelpers';
+import useAppLaunch from './useAppLaunch';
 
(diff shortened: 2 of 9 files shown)


goblincore:
Reviewed the gitStream/LinearB findings:

**Issue 1 (`MoreAppTile` doesn't hide `isHardBlocked` apps) — fixed in `e27d4d00a`.** Legitimate parity gap with `AppStoreListItem`. `MoreAppTile` now destructures `isHardBlocked` from `useAppLaunch` and returns `null` when true, matching the existing behavior for age-restricted apps.

**Issue 2 (skeleton `max-w-[100px]` vs new `max-w-[160px]`) — dismissed as false positive.** The bot didn't distinguish `AppGridTile`'s two branches:

- The **image** branch (used by `MoreAppTile` → tiles the skeleton replaces) is still `max-w-[100px]` — this PR did not change it.
- Only the **gradient** branch (LearnCard shortcut tiles) went to `max-w-[160px]`, and those don't have a loading state — the shortcuts are static and render immediately.

So the skeleton correctly matches the tile it stands in for; there is no layout jump.

PR #1383, opened 2026-07-09
fix(notifications): decrement header alerts-island unread count on read

goblincore:
## Problem

The header **alerts island** (megaphone icon) unread badge did not decrement/clear after a notification was read. Repro (desktop web):

1. New credential arrives → island shows unread badge (e.g. `1`).
2. Open the notifications modal → click the credential/boost notification → it opens the claim view and is visibly marked read.
3. Close the modal(s).
4. **Bug:** the island badge still shows the same count.

## Root cause

`useUpdateNotification.onSuccess` and `useMarkAllNotificationsRead.onSuccess` read the switched DID with `switchedProfileStore.use.switchedDid()` — a **zustand React hook** — from *inside* the react-query mutation callback, i.e. **outside a React render**.

`store.use.X()` (zustood) resolves to the zustand `useStore` hook (`@udecode/zustood` `index.js:58`). Calling it outside render throws *"Invalid hook call"*, which aborts the `invalidateQueries`/`refetchQueries` for `['useGetUnreadUserNotifications', did]` **before it runs**. So:

- The island's unread query never refetched → badge never cleared.
- Even the `markAllNotificationsRead` fired on modal-close was a silent no-op (its `onSuccess` threw and was swallowed by a `try/catch`).

The rest of the codebase already reads the store from callbacks with the non-reactive `switchedProfileStore.get.switchedDid()` (`index.js:97`, `getState()`) — see `boosts.ts` / `mutations.ts`. Only these two notification callbacks used `.use` in a callback.

Server-side read persistence was verified **correct** and ruled out: `NotificationMetaValidator` declares `read` and `actionStatus: 'COMPLETED'`, and `updateNotificationMeta` does `$set: meta`, so a `{ actionStatus, read }` claim update persists both fields.

## Fix

- **Use `.get.switchedDid()` in both `onSuccess` callbacks** so invalidation/refetch actually runs (the real fix for the reported bug).
- **Fix dead code in `onMutate`:** `isArchiving = payload?.archived` made `else if (!isArchiving)` swallow every non-archive update, so the trailing `else if (payload.read)` branch was unreachable. Now distinguishes the `archived` tri-state (`true`/`false`/`undefined`) so read updates flip `read` in the active-list cache.
- **Optimistic island decrement:** `onMutate` now removes the read notification from the `useGetUnreadUserNotifications` cache so the badge decrements *instantly*, not only after the refetch (defense-in-depth against any eventual-consistency lag).

## Tests

Adds `notifications.test.tsx` (4 tests) using the **real** zustood store so `.use.*()` genuinely throws outside render — reproducing the exact failure. All fail on `main`, pass with this change. Rest of the `learn-card-base` suite is unaffected (one pre-existing, unrelated `dateHelpers` age test failure is not touched by this PR).

## Notes / out of scope

- Separate from the notifications-desktop-modal PR (#1380); the bug pre-dates that branch (it's in the shared mutation layer).
- `useMarkNotificationRead` (a *different*, unused-in-this-path mutation) has pre-existing type errors / dead code (`notification` undefined reference). Left untouched here.

### ✨ PR Description
Purpose: Fix header alerts-island unread count failing to decrement by using non-reactive store access in mutation callbacks and implementing optimistic cache updates with rollback.

Main changes:
- Replace reactive `.use.switchedDid()` hook calls with `.get.switchedDid()` in mutation callbacks to prevent throws outside React render context
- Add optimistic updates and rollback logic to unread notification cache on read/archive mutations with proper `onError` handler
- Refactor `useMarkNotificationRead` with proper success callback implementation and unread cache management


goblincore:
@claude please review

goblincore:
Thanks @gitstream-cm — good catch, addressed in cf73a174b.

`useUpdateNotification.onMutate` writes optimistically (active/archive lists + the unread island cache) *before* the server call, but the hook had **no `onError` at all** — so on failure the badge stayed at the decremented value until the next refetch (and `previousData` was likewise never restored).

Fix:
- Snapshot the active, archive, and unread caches in `onMutate` and return them as a typed `UpdateNotificationContext`.
- Add `onError` to restore all three snapshots on failure.
- Added a regression test that rejects `updateNotificationMeta` and asserts the unread count is restored.

Note: the sibling `useMarkNotificationRead` does its optimistic writes in `onSuccess` (post-success only), so it has no equivalent rollback gap.

goblincore:
Fixed in 0c78defa0 — removed the dead `archiveData` assignment in the unarchive branch; all reads use the earlier `currentArchiveData` snapshot. Thanks @gitstream-cm.

PR #1394, opened 2026-07-10
perf(clients): eagerly prefetch DID-Auth challenges without duplicate requests

goblincore:
# Overview

#### 🎟 Relevant Jira Issues

N/A — performance follow-up from staging boot-trace analysis (Trace-20260709T165929.json).

Not super high priority relative to the other UI/UX bug fixes so doesn't need to get reviewed soon, just thought it as a nice chore cleanup task

A hard-refresh boot trace of the LearnCard app on staging showed **22 calls to `/trpc/utilities.getChallenges` between 3.9s and 6.3s of boot** (~95 challenges each). Investigation found two compounding causes:

1. **Eager challenge prefetch** — every `getClient()` in the tRPC client packages fired `getChallenges()` at construction, even for clients that never issued a request.
2. **Duplicate wallet construction** — boot was building **three full wallets** (each with its own brain/cloud/lca clients and plugin init traffic) when it only needed one.

#### 🥴 TL; RL:

Challenges are now fetched lazily (on first request, with in-flight dedup), and boot builds exactly one wallet instead of three. Measured on a real logged-in dev boot against staging:

| Stage | `getChallenges` requests | Wallet builds |
|---|---|---|
| Original trace | ~22 | 3 |
| After lazy challenge fetching | 13 | 3 |
| After wallet dedup | **5** | **1** |

Challenge traffic also finishes by ~9s instead of trickling out past 16s.

BEFORE:
Screenshot 2026-07-10 at 12 26 08 PM

Screenshot 2026-07-10 at 12 26 23 PM

AFTER:

Screenshot 2026-07-10 at 10 32 43 AM

Screenshot 2026-07-10 at 10 33 00 AM

**Commit 1 — lazy challenge fetching (4 client packages)**

- Removed the eager `getChallenges().then(...)` at construction in `@learncard/network-brain-client`, `@learncard/learn-cloud-client`, `@learncard/simple-signing-client`, and `@learncard/lca-api-client` (the 4th found by repo-wide grep; the trace analysis only flagged 3).
- The pool now fills lazily in the `headers` callback, guarded by an in-flight `refillPromise` so concurrent first requests share one fetch instead of each pulling ~95 challenges.
- The 401 path is untouched: `callbackLink` still refetches and wholesale-replaces the pool before retrying.
- First unit tests for these packages: vitest + a batch-aware `fetch` stub, 4 tests × 4 packages (lazy construction, single fetch + pool reuse, concurrent dedup, 401 refetch/retry).

**Commit 2 — wallet construction dedup (`learn-card-base` + app entry)**

- **Concurrent-construction race**: `getBespokeLearnCard`/`getSigningLearnCard` cached the *resolved* wallet, which takes seconds to build — so `useWallet`, `AuthCoordinatorProvider`, and react-query hooks racing at boot all missed the cache and built duplicates. They now cache the **in-flight promise** (evicted on failure so retries still work).
- **Dummy fallback wallet**: `getWalletOrFallback` trusted the not-yet-settled `isLoggedIn` flag and built a throwaway seed-`'a'` wallet for logged-in users during early boot. It now checks secure storage for the real key first; the fallback remains for genuinely logged-out browsing.
- **Hardcoded `'aaa'` wallet**: `useGetProofOfLoginVp` built a full network wallet from a hardcoded 3-char seed on every boot (also the long-standing source of the "seed is less than 32 bytes" console warnings). The server route (`firebase.getProofOfLoginVp`) is **open** — it only verifies the Firebase token and ignores the caller's DID — so it now reuses the logged-in user's cached wallet. `'aaa'` is kept for the genuinely pre-login flows (`sendLoginVerificationCode`, `verifyLoginCode`, `verifyNetworkHandoffToken`), which are untouched.
- Boot observability: raised the resource-timing buffer in the app entry (default 250 overflows under Vite dev's module count) so boot network activity can be audited from the console, and added a debug-level log per wallet construction so duplicate builds are visible as regressions.

- The first real request per client pays **one extra serialized round-trip** for challenges (the pool is no longer pre-warmed at construction). Intended trade: construction is free; only clients that are actually used pay.
- A cross-client shared challenge pool was considered and rejected: challenges are **DID-bound server-side** (`setValidChallengesForDid`), so pools can only be shared between clients with the same DID — which wallet dedup already achieves, without needing the DID inside `getClient`.
- **Remaining 5 fetches** = 1 brain + 2 cloud + 2 lca: the LearnCloud and LCA plugins rebuild their tRPC client when the wallet's DID upgrades from `did:key` to `did:web` mid-init, discarding the (DID-bound, so legitimately unusable) pool; LearnCloud also fires eager `getDids`/`getDid` queries at construction. Fixing that is a plugin-lifecycle change tracked as a follow-up task rather than expanding this PR. Target there is 3 (one per service).

#### 🔍 Types of Changes

-   [x] No
-   [ ] Yes

# Testing

**Unit tests** (each package, from repo root):

```bash
cd packages/learn-card-network/brain-client && bun run test
cd packages/learn-card-network/cloud-client && bun run test
cd packages/learn-card-network/simple-signing-client && bun run test
cd packages/lca-api-client && bun run test
```

**End-to-end boot measurement** (the real success criterion):

1. Check out this branch. For the Vite dev server no rebuild is needed (workspace packages resolve to `src` via the `development` export condition); for a production build, rebuild the four client packages.
2. Run the app against staging, log in, open DevTools → Network, filter `getChallenges` + Fetch/XHR type.
3. Hard-refresh: expect **~5 requests** (was ~22), all completing within the first ~9s. Or from the console: `performance.getEntriesByType('resource').filter(e => e.name.includes('getChallenges')).length`.
4. Console check: no "seed is less than 32 bytes" warnings on a logged-in boot (previously appeared every boot via the `'aaa'` wallet).
5. Auth-adjacent regression pass (these changes touch auth-adjacent paths): log out/in, email contact-method verification, send a boost, switch profiles, view credentials. Verified during development: logged-in boot renders dashboard normally; logged-out `/login` builds zero wallets and fetches zero challenges.
6. 401/expired-challenge path: idle until challenges expire, then act — transparent refetch + retry (also unit-tested).

Chromium (staging web app) is sufficient for the perf claims; a quick native sanity pass wouldn't hurt since `learn-card-base` boot paths are shared.

#### 🧪 Code Coverage

16 new unit tests across the four client packages (vitest, batch-aware `fetch` stub — no live server needed): no fetch at construction, one fetch + pool reuse, concurrent dedup, 401 refetch/retry. The wallet-dedup changes were verified empirically via instrumented live boots (wallet-build counts + resource-timing request counts) rather than unit tests, since `walletHelpers` pulls in WASM/stores that aren't unit-test friendly.

# Documentation

**Internal/AI Docs**

-   [x] **Code comments/JSDoc** — Complex logic that needs inline explanation

**Visual Documentation**

#### 💭 Documentation Notes

Internal perf refactor with no public API changes — `getClient()` and `getBespokeLearnCard()` signatures and behavior contracts are unchanged. Inline comments explain the lazy-fill/dedup pattern in each client, the promise-cache rationale in `walletHelpers`, and why `useGetProofOfLoginVp` can use the real wallet.

# ✅ PR Checklist

### ✨ PR Description
Purpose: Implement lazy challenge prefetching and promise-based caching to eliminate duplicate wallet/client construction during concurrent boot operations across multiple packages.

Main changes:
- Add eager challenge prefetch with single-flight refills in all four client packages (lca-api, brain, cloud, simple-signing) via refillChallenges/takeChallenge pattern
- Convert wallet helper caches from resolved instances to in-flight promises, evicting failed builds to enable retries
- Improve getWalletOrFallback to check secure storage before fallback, and reuse logged-in wallet in useGetProofOfLoginVp instead of dummy seed


goblincore:
## 📉 Round 2: measured live and pushed wallet-construction dedup

Measured a real logged-in dev boot against staging (driven + counted via browser automation, not eyeballed):

| Stage | `getChallenges` requests | Wallet builds |
|---|---|---|
| Original trace (Trace-20260709T165929) | ~22 | 3 |
| After lazy challenge fetching (first commit) | 13 | 3 |
| After wallet dedup (this commit) | **5** | **1** |

The lazy-client change alone couldn't reach the target because **boot was building three full wallets**, each with its own brain/cloud/lca clients:

1. **Concurrent-construction race** — `getBespokeLearnCard` cached the *resolved* wallet (built over several seconds), so `useWallet`, `AuthCoordinatorProvider`, and react-query hooks racing at boot all missed the cache and built duplicates. Now caches the in-flight promise (evicted on failure so retries work).
2. **Dummy fallback wallet** — `getWalletOrFallback` trusted the not-yet-settled `isLoggedIn` flag and built a throwaway seed-`'a'` wallet for logged-in users. Now checks secure storage for the real key first.
3. **Hardcoded `'aaa'` wallet** — `useGetProofOfLoginVp` built a wallet from a hardcoded 3-char seed on every boot (also the source of the "seed is less than 32 bytes" console warnings). The server route is **open** — it only verifies the firebase token and ignores the caller's DID — so it now reuses the logged-in user's cached wallet; `'aaa'` remains for pre-login flows only.

**Remaining 5** = 1 brain + 2 cloud + 2 lca: the LearnCloud and LCA plugins rebuild their tRPC client when the wallet's DID upgrades from `did:key` to `did:web` mid-init, discarding the challenge pool. Carrying the pool across rebuilds (or deferring the plugins' eager init queries) is a deeper plugin-lifecycle change — proposed as a follow-up rather than expanding this PR.

Also: raised the resource-timing buffer in the app entry (default 250 overflows under Vite dev) so boot network activity is auditable from the console, and added a debug-level log per wallet construction to catch regressions.

apps/learn-card-app/src/index.tsx
```diff
@@ -17,6 +17,10 @@ import * as Sentry from '@sentry/browser';
 
 (window as any).Buffer = Buffer;
 
+// Keep resource-timing entries for the whole boot (default 250 overflows under
+// Vite dev's module count) so boot network activity can be audited from the console
+performance.setResourceTimingBufferSize(10000);
+
```
TaylorBeeston:
```suggestion
```

Probably don't want to keep this! It's not a _huge_ deal, but it could lead to a smol increase in memory usage on prod

TaylorBeeston (changes requested):
I need to sit with this more, but I believe that removing the eager challenge fetching is the wrong move! I think the real value-add from this guy is the wallet/request deduping! But the eager challenge fetching is actually a _good_ thing for performance!

TaylorBeeston:
Pushed `ce0463756` with the revised design.

## Why

The original `22 → 13` result combined two changes: removing eager prefetch **and** adding an in-flight `refillPromise`. The old duplicate traffic came from an uncoordinated race: construction started `getChallenges()`, then the first request saw an empty pool and started another fetch. The revised implementation keeps the valuable part—the single-flight promise—but starts it eagerly. A first request that races construction now awaits the same fetch, preserving prewarming without creating a second pool.

I also made challenge acquisition exhaustion-safe, added explicit empty-refill failures, retained the 401 recovery path, added wallet promise-cache regression coverage, reused the active wallet for proof-of-login, stopped secure-storage failures from silently selecting the dummy wallet, and removed the production resource-timing instrumentation.

## Measurements

Measured on an Intel Core i9-14900HX/Linux machine against the production Brain, LearnCloud, and LCA APIs. Five interleaved runs per variant; values below are medians. The wallet scenario requested three concurrent full network wallets with the same deterministic seed, awaited profile readiness, and drained pending network work.

| Variant | Wallets | Challenges | Construction | First authenticated response | Total |
|---|---:|---:|---:|---:|---:|
| Original eager, no promise cache | 3 | 21 | 383 ms | 1,083 ms | 1,909 ms |
| Lazy single-flight, no promise cache | 3 | 9 | 384 ms | 1,105 ms | 1,924 ms |
| Lazy + wallet promise cache | 1 | 2 | 162 ms | 1,122 ms | 1,272 ms |
| **Eager single-flight + wallet promise cache** | **1** | **3** | **156 ms** | **1,109 ms** | **1,272 ms** |

Lazy fetching alone reduced request count but did not improve completion time (`+0.8%`). Wallet promise caching was the material improvement: `3 → 1` wallets and `33.9%` lower total time. Restoring eager prefetch added one unused LCA pool in this no-profile harness, but total time was effectively identical (`+0.03%`) and first authenticated response was 13 ms faster.

I also measured the intended prewarming case by allowing 250 ms of caller/plugin setup between construction and the first authenticated request:

| Variant | Challenges | First request | Total |
|---|---:|---:|---:|
| Original eager race | 4 | 794 ms | 1,094 ms |
| Lazy single-flight | 2 | 803 ms | 1,108 ms |
| **Eager single-flight** | **2** | **551 ms** | **849 ms** |

Eager single-flight preserved the deduped challenge count while reducing first-request latency by 253 ms / 31.4% versus lazy fetching.

Limitation: this is a deterministic full-wallet initialization benchmark, not an authenticated Firebase dashboard boot. Without a real profile it does not exercise `did:key → did:web` client rebuilds, so absolute logged-in challenge counts may differ. It does isolate the behavior changed here and reproduced the original ~22-request symptom (`21` requests).

## Verification

- Four client suites: 24/24 tests passed
- Wallet promise-cache suite: 4/4 tests passed
- NX builds for all four clients and their dependency graph passed
- Full `learn-card-app` production build and its 42 dependencies passed

TaylorBeeston (approved):
Approved after revising the challenge flow to eager single-flight prefetching, adding exhaustion-safe acquisition and cache regression coverage, and re-benchmarking all variants on the same machine. The measurements confirm wallet promise deduplication is the primary boot win while eager single-flight preserves that reduction and avoids first-request latency.

TaylorBeeston:
@claude please re-review

PR #1408, opened 2026-07-13
fix: did:web SAs

Custard7:
# Overview

A partner reported that LearnCard-issued credentials were failing to verify in strict verifiers (`@digitalcredentials/vc`) with:

> Verification method "did:web:network.learncard.com:users:lefdemoschool#lca-sa" not authorized by controller for proof purpose "assertionMethod"

The root cause is in how we build a profile's (and app's) `did:web` document when a signing authority is attached. Two things were wrong, plus the underlying data problem that caused them to pile up.

#### 🥴 TL;DR

- Signing-authority keys had a `controller` that included a fragment (e.g. `...:lefdemoschool#lca-sa`). Per the DID spec, a key hosted in its own document must be controlled by the base DID (no fragment). When a verifier dereferences a fragment controller it only gets the key node back — which has no `assertionMethod` — so authorization fails. Credentials signed with the profile's own `#owner` key were fine; only signing-authority-signed credentials broke.
- The same key showed up multiple times in the document (`lefdemoschool` had it 3×). That's because the "profile uses signing authority" relationship was created with a plain create every time it was registered, so re-registering stacked duplicate relationships, and the document builder appended each one without de-duping.

#### 💡 What changed

1. **`dids.ts`** — signing-authority (and manager) verification methods now use the base DID as their `controller`, never a fragment. Added a small `normalizeDidDocument` pass that runs right before each document is cached/returned. It (a) forces any locally-hosted method's controller to the base DID and (b) de-duplicates `verificationMethod` / `authentication` / `assertionMethod` / `keyAgreement`. This is applied to the profile, app, manager, and root `.well-known` routes, and also cleans up any already-bad data still living in the graph.
2. **`accesslayer/signing-authority/relationships/create.ts`** and **`accesslayer/app-store-listing/relationships/create.ts`** — the "uses signing authority" relationship is now created with a `MERGE` keyed on name + did instead of an unconditional create, so registering the same authority twice no longer produces duplicate relationships.

The `bun.lock` changes are incidental lockfile churn.

#### 🔍 Types of Changes

This is backwards compatible. Public keys, key IDs, and the `assertionMethod`/`authentication` lists are unchanged — only the `controller` values are corrected and duplicates removed. No credential needs to be re-issued; affected credentials simply start verifying once the corrected document is served.

# Testing

Before deploy you can reproduce against production data:

```
curl -s https://network.learncard.com/users/lefdemoschool/did.json
```

Today you'll see `#lca-sa` repeated three times and its `controller` ending in `#lca-sa`.

After deploy, on a profile that has a signing authority attached:

1. Fetch `/users/<profileId>/did.json`.
2. Confirm every `verificationMethod[].controller` equals the document `id` (no `#` fragment).
3. Confirm each key id appears only once across `verificationMethod`, `authentication`, and `assertionMethod`.
4. Verify a credential that was signed via that signing authority against the document using `@digitalcredentials/vc` — it should now pass.

Note: DID documents are cached in Redis (~1h TTL, refreshed on read). To see the fix immediately after deploy, evict the `did-doc:*` / `app-did-doc:*` / `manager-did-doc:*` keys; otherwise they self-heal within the hour.

#### 🧪 Code Coverage

There is a local test file (`services/learn-card-network/brain-service/test/dids.spec.ts`) covering the controller normalization, de-duplication, and relationship idempotency — unit tests for the normalizer plus integration tests through the real route. It is not yet committed to this branch; commit it before merge if we want it in CI.

# Follow-ups (not in this PR)

- A one-time cleanup/migration to collapse duplicate `USES_SIGNING_AUTHORITY` relationships already in Neo4j. The read-time de-dup masks them, but removing them at the source is cleaner.
- There is no `deleteDidDocForApp` cache helper yet (profile and manager have one); worth adding for symmetry.

### DIFF
--- a/.changeset/tiny-impalas-ring.md
+++ b/.changeset/tiny-impalas-ring.md
@@ -1,0 +1,5 @@
+---
+"@learncard/network-brain-service": patch
+---
+
+fix: did:web SAs
--- a/bun.lock
+++ b/bun.lock
@@ -64,7 +64,7 @@
     },
     "apps/learn-card-app": {
       "name": "learn-card-app",
-      "version": "1.96.5",
+      "version": "1.97.3",
       "dependencies": {
         "@capacitor-community/media": "9.1.0",
         "@capacitor-community/sqlite": "8.1.0",
@@ -246,7 +246,7 @@
     },
     "apps/scouts": {
       "name": "scoutpass-app",
-      "version": "1.90.20",
+      "version": "1.90.22",
       "dependencies": {
         "@capacitor-community/sqlite": "8.1.0",
         "@capacitor-firebase/analytics": "8.2.0",
@@ -365,7 +365,7 @@
     },
     "examples/app-store-apps/1-basic-launchpad-app": {
       "name": "@learncard/app-store-demo-basic-launchpad",
-      "version": "0.0.32",
+      "version": "0.0.33",
       "dependencies": {
         "@astrojs/netlify": "^7.0.13",
         "@learncard/init": "workspace:*",
@@ -380,7 +380,7 @@
     },
     "examples/app-store-apps/2-lore-card-app": {
       "name": "@learncard/app-store-demo-lore-card",
-      "version": "0.0.33",
+      "version": "0.0.34",
       "dependencies": {
         "@astrojs/netlify": "^7.0.13",
         "@learncard/init": "workspace:*",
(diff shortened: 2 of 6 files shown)


TaylorBeeston (approved):
Perfect! I can also go in AWS and forcibly clear the did cache if needed!

PR #1410, opened 2026-07-13
fix: Query Param Merging

Custard7:
# Overview

When a partner submits an embedded app with a launch URL that already has query params (e.g. `https://verifier.prettygoodskills.com/jobs?embed=learncard-partner-connect`), we tack on our own `lc_host_override` param with a raw `?`. That produced a broken URL with two `?`:

`.../jobs?embed=learncard-partner-connect?lc_host_override=https://learncard.app`

This isn't just ugly — it silently breaks the override. The partner-connect SDK reads it with `URLSearchParams(...).get('lc_host_override')`, and with the double-`?` the whole thing parses as a single `embed` value, so `lc_host_override` is never found. The embedded app then falls back to its default origin, so staging/cross-tenant embeds don't work.

#### 🥴 TL; RL:

We were string-concatenating query params onto URLs that might already have a query string. Now we merge them properly.

#### 💡 Feature Breakdown

- Added a small `appendQueryParams(url, params)` helper in `packages/learn-card-base/src/helpers/urlHelpers.ts` that uses the native `URL` API to merge params correctly (well-formed output, auto-encoded, overwrites duplicate keys).
- Swapped all 5 places that built these URLs by hand over to the helper:
  - `EmbedAppFullScreen.tsx` — iframe src + notification deep-link navigation
  - `EmbedIframeModal.tsx` — iframe src + notification deep-link navigation
  - `AppPreviewModal.tsx` — preview iframe src
- The two notification handlers also dropped their now-redundant manual `encodeURIComponent`.

#### 🔍 Types of Changes

-   [x] No

# Testing

1. Take a partner app whose launch URL already has a query string, e.g. `https://verifier.prettygoodskills.com/jobs?embed=learncard-partner-connect`.
2. Launch it (both the modal and full-screen views).
3. Inspect the iframe `src`. It should now read `...?embed=learncard-partner-connect&lc_host_override=<host>` — one `?`, an `&` before our param, no double-`?`.
4. Confirm the embedded app actually picks up the host override (works against staging), instead of falling back to its default origin.
5. Send a notification with an `actionPath` and tap it — the resulting navigation URL should also be well-formed.
6. Repeat the check in the app preview modal (`AppPreviewModal`).

#### 🧪 Code Coverage

Added unit tests for the new helper (`urlHelpers.test.ts`): no existing query string, existing query string (no double-`?`), duplicate-key overwrite, fragment preservation, multiple params, and invalid-URL throwing.

# Documentation

#### 💭 Documentation Notes

Internal bug fix, no API or user-facing changes. The new helper has JSDoc explaining the footgun it prevents.

### ✨ PR Description
Purpose: Replace manual query parameter string concatenation with a utility function to properly merge parameters and prevent malformed URLs with duplicate query separators.

Main changes:
- Added `appendQueryParams()` utility function to `urlHelpers.ts` using URL API for safe parameter merging with comprehensive test coverage
- Replaced all manual `${url}?param=value` concatenations with `appendQueryParams()` calls across three embed components
- Updated imports across files to include new `appendQueryParams` from `learn-card-base`

### DIFF
--- a/.changeset/mighty-moose-write.md
+++ b/.changeset/mighty-moose-write.md
@@ -1,0 +1,5 @@
+---
+"learn-card-app": patch
+---
+
+fix: Query Param Merging
--- a/.changeset/modern-pigs-brake.md
+++ b/.changeset/modern-pigs-brake.md
@@ -1,0 +1,6 @@
+---
+"learn-card-app": patch
+"learn-card-base": patch
+---
+
+fix: Query Param Merging
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,6 @@ services/learn-card-network/*/didkit-native-layer/
 services/learn-card-network/lca-api/.env.scouts
 /apps/scouts/.vite-local
 /apps/scouts/.vite-docker
+
+# Sisyphus agent working directory (local only)
+.sisyphus/
--- a/apps/learn-card-app/src/pages/appStoreDeveloper/components/AppPreviewModal.tsx
+++ b/apps/learn-card-app/src/pages/appStoreDeveloper/components/AppPreviewModal.tsx
@@ -4,7 +4,7 @@ import { X, Maximize2, Play, AlertCircle } from 'lucide-react';
 import { getLogger } from 'learn-card-base';
 const log = getLogger('app-preview-modal');
 
-import { useModal } from 'learn-card-base';
+import { useModal, appendQueryParams } from 'learn-card-base';
 import { useLearnCardPostMessage } from '../../../hooks/post-message/useLearnCardPostMessage';
 import { useLearnCardMessageHandlers } from '../../../hooks/post-message/useLearnCardMessageHandlers';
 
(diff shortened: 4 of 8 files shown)


gerardopar:
@claude review pr please

TaylorBeeston (approved):

Custard7:
Thanks for the reviews — the throw-vs-graceful-degradation catch is a good one and I've addressed it in 77ebc54.

**What changed:** `appendQueryParams` is now total — it wraps the `new URL()` parse in a try/catch and returns the input unchanged on failure. This restores the pre-existing behavior: a truthy-but-invalid `embedUrl` flows through untouched, so the components' existing handling (empty `embedOrigin` → error toast / redirect) kicks in as designed, instead of throwing during render.

I went with "make the helper total" over "guard each call site" so there's nothing to forget at future call sites, and the render-time calls (`EmbedIframeModal`, `EmbedAppFullScreen`, `AppPreviewModal`) need no per-site try/catch. Updated the unit test accordingly (the throw case is now "returns input unchanged", plus an empty-string case).

On the optional notes: left `Record<string, string>` typing and the `appendQueryParams` name as-is — the overwrite-on-duplicate behavior is intended and documented in the JSDoc, and renaming a just-added export felt like unnecessary churn. Happy to rename if you feel strongly.

PR #1413, opened 2026-07-14
feat: [LC-1963] Onboarding 2.0

Custard7:
# Overview

Screenshot 2026-07-14 at 3 02 09 PM

Rebuilds the LearnCard sign-up flow. The old flow had four steps (age gate → privacy → role → profile form), let you skip partway through, and then re-prompted you — often with the birthday picker again — the moment you tried to use a feature. People also got stuck on the "User ID" step: they didn't realize it was a public, searchable handle, and they'd stall trying to pick one.

This replaces all of that with a short, guided flow that fills in sensible defaults for you and never traps you in a re-prompt loop.

#### 🥴 TL; RL:

New 3-screen onboarding: (1) birthday + country, (2) your profile card, (3) you're in. Fewer steps, no skips, smart defaults, lands you straight on the dashboard.

-   **Screen 1 — Birthday & country.** Age drives our privacy/consent rules, so it comes first. Birthday uses an iOS-style scroll wheel instead of a calendar (nobody wants to click back 30 years). Country is pre-filled from your device locale but stays editable. Minors see the required consent notice inline (US) or a guardian-email field (EU) — no surprise pop-ups.
-   **Screen 2 — Your card.** A live preview card updates as you type. Name and photo are pre-filled from your Google/Apple login. The handle is auto-generated from your name and keeps updating until you edit it yourself, so most people never touch it. Privacy toggles (AI, analytics, crash reports) are pre-set based on age.
-   **Screen 3 — You're in.** Confetti, your finished card, an optional "what brings you here?" role picker, and one button into the app.
-   On desktop (≥992px) screen 2 is a two-column layout (card left, form right); everything collapses to a single column on mobile.

Key files (all under `apps/learn-card-app/src/components/onboarding/v2/`): `OnboardingFlow.tsx` (the flow), `BirthdayPicker.tsx` (wheel), `countryInference.ts`, `handleGenerator.ts`. The old onboarding components and their tests are deleted. `LoginPage.tsx` and `useJoinLCNetworkModal.tsx` now open the new flow.

-   **No skipping.** The flow is short and mandatory, which is what removes the confusing "it popped up again" behavior. A profile always exists by the end.
-   **Handles can be auto-generated and changed later.** profileId is not immutable, so we generate one and tell users they can change it anytime, instead of blocking them on the decision.
-   **Deleted the old onboarding unit tests** along with the components they covered. The new flow isn't unit-tested yet (see tech debt).

#### 🔍 Types of Changes

-   [ ] No
-   [x] Yes

No automated tests cover the new `v2` flow yet (the old onboarding tests were removed with the old components). Worth a follow-up to add coverage for the birthday picker, handle generation, and consent branching.

# Testing

Sign in as a brand-new user (no profile yet):

1.  **Happy path:** birthday wheel → country is already filled from your locale → Continue → name/photo pre-filled from your login → handle auto-filled → Create → confetti → **Explore** drops you straight on the dashboard (no intro slides, no "what would you like to do" prompt).
2.  **Handle:** type a name and watch the handle track it; edit the handle yourself and it stops auto-updating. Type a handle that's already taken to see the inline validation. Confirm the handle you set is the one that's actually created (no random suffix appears after Create).
3.  **Minors:** enter a birthday under 18. Non-EU country shows an inline parental-consent checkbox; an EU country shows a guardian-email field. AI is off and locked for minors.
4.  **Desktop vs mobile:** at ≥992px screen 2 is two columns; below that it's one.
5.  **Feature-gated entry:** as a user without a profile, trigger a network action (e.g. share/send). Onboarding opens; finishing it should resume that original action rather than dumping you on the dashboard.
6.  **Role:** pick a role on the success screen and confirm it sticks on your profile.

iOS Native / Android Native / Chromium (native is worth a look for the wheel picker and safe-area spacing)

#### 🧪 Code Coverage

None added yet — see tech debt note above.

# Documentation

-   [ ] **App Flows** — Changes to LearnCard App user flows (`docs/apps/`)

#### 💭 Documentation Notes

UI-only change to sign-up; no SDK or API surface changed. If we keep a written onboarding walkthrough in `docs/apps/learn-card-app/`, it should be refreshed to match the new 3-screen flow.

# ✅ PR Checklist

-   [ ] Related to a Jira issue
-   [ ] I have **commented** my code, particularly where ambiguous
-   [ ] New and existing **unit tests pass** locally with my changes

### ✨ PR Description
Purpose: Replace legacy OnboardingContainer with new OnboardingFlow v2 component featuring redesigned UI, birthday picker, and enhanced onboarding logic.

Main changes:
- Implemented OnboardingFlow component with multi-step profile creation (age-country, profile, celebrate) and BirthdayPicker UI
- Updated analytics tracking to reference OnboardingFlow instead of OnboardingNetworkForm in signup flag handling
- Refactored onboarding modal integration in LoginPage and JoinLCNetworkModal to use new OnboardingFlow component


goblincore:
@claude please review

Custard7:
Thanks for the continued review. Notes on the three latest (issue_8/9/10) — none block this PR:

**issue_8 (SQL injection in `updateCurrentUser`)** — Real hygiene issue, but pre-existing and out of scope here. `useSQLiteStorage` builds every statement with raw string interpolation (~15 call sites), and `updateCurrentUser` is already called this way on `main` from `NewJoinNetworkPrompt` and `UserProfileUpdateForm`. It's a local, single-user, on-device SQLite DB, so there's no cross-user or server exposure — the practical impact is a display name containing a literal `"` producing malformed SQL that's caught and logged (the name just doesn't cache locally; it re-syncs from the network on next load). The `uid` fix didn't introduce this: the `SET name="${...}"` interpolation always executed — only the row match changed from 0 rows to 1. The right fix is parameterizing the shared helper across all call sites, which I'll do as a separate hygiene PR rather than band-aid a single caller here.

**issue_9 (empty name/photo for non-Firebase users)** — Not a bug in this flow. Onboarding only opens for users who don't have a profile yet, so a new signup has no stored name/photo to pre-fill in the first place. The Create button stays disabled while `name` is empty (`OnboardingFlow.tsx:986`), so a nameless profile can't be created. OAuth users are pre-filled by the Firebase effect. No data loss.

**issue_10 (double-suffixed handle)** — The suggested output (`josiah-ab12-cd34`) can't actually occur: the collision setter strips a trailing `-xxxx` before appending (`prev.replace(/-[a-z0-9]{4}$/, '')`, line 198), so it re-rolls the suffix rather than stacking it. The re-fire window is also guarded by `!uniqueProfileFetching` and `uniqueProfile != null` (line 195), so a not-yet-fetched (`undefined`) key doesn't trigger it either.

For reference, issue_1, issue_2, issue_4, and issue_6 are addressed in the latest commits.

goblincore (approved):
@Custard7 This is great!! much smoother experience.

A few small things though none of them are like blocking or critical, probably more like things that could be followups:

1) I kinda feel there is a little too much margin/padding on top on mobile
Screenshot 2026-07-15 at 12 55 28 PM

2) So this isn't really an issue with this PR per se, it's existing behavior, but with the new UI i think it becomes a little confusing (see attached screen recording) 

https://github.com/user-attachments/assets/9c12d09f-392e-478c-9139-6f0924800abd

basically this has to do with the existing age gating logic:

Like if you are above 13 but below 18 (or whatever the legal age is set for the that particular EU country) it seems we show this disclaimer:
 
Screenshot 2026-07-15 at 12 53 06 PM

But if you go below 13 it disappears and nothing is shown until you hit the next button:


Screenshot 2026-07-15 at 12 53 09 PM
Screenshot 2026-07-15 at 12 53 12 PM

This is fine, but a little confusing since the messages shown to above 13 but below 18 seems to imply that anyone below 18 would see that message and go through that particular flow (the account approval via guardian email). I guess maybe modifying the text copy might be the simplest though also  I kinda feel maybe we could simplify it where its just email approval? ....anyways this is part of a larger ongoing thing i think with parent and child accounts that we haven't quite sorted out yet.

Somewhat related, for the above 13-below 18 (or whatever legal age as defined per country) currently there isn't really any additional UI that tells the user their account is in pending approval (which is fine, this is how it was before).  Also not entirely related but it's not clear what is gated other than the AI - I think the only thing that is really gated other than the AI stuff is the "Issue Credential" button but the boost a friend on the apps page isn't gated like that I think though I could be wrong.

PR #1416, opened 2026-07-15
feat: Update to final OID4VCI/OID4VP 1.0 spec

Custard7:
## What

Updates the OpenID4VC holder implementation (`packages/plugins/openid4vc`) from the old drafts to the **final OID4VCI 1.0 and OID4VP 1.0 specs**, while keeping the pre-1.0 behavior working for issuers/verifiers that haven't migrated yet.

It's a big diff but it's two halves — issuance (VCI) and presentation (VP) — plus the test harnesses, the playground, and two small wallet-app fixes. The whole thing was validated end-to-end against the LER Interoperability Test Suite (a real 1.0 issuer + verifier we don't control) — both the issuance and presentation flows now pass it.

Screenshot 2026-07-16 at 4 45 15 PM

## Why

The specs changed the wire format between the drafts we targeted and the final 1.0 releases. We hit this against real 1.0 deployments (metadata 404s, rejected proofs, rejected responses) and this brings us in line with 1.0 without dropping the large installed base still on the drafts.

## Issuance (OID4VCI)

- **Metadata discovery URL**: 1.0 inserts `/.well-known/openid-credential-issuer` between host and path (draft appended it to the end). Fixes discovery for issuers whose identifier has a path.
- **Credential request body**: 1.0 uses `credential_configuration_id` + a `proofs` object (draft used `format` + `credential_definition` + a single `proof`).
- **`di_vp` key proofs** (`vci/proof.ts`): the wallet now reads the issuer's `proof_types_supported` per configuration and picks a proof type. Issuers that only accept `di_vp` (a W3C VP with a Data Integrity proof — `proofPurpose: authentication`, `domain` = issuer, `challenge` = `c_nonce`) get one, signed with a mutually supported cryptosuite. Everyone else keeps getting the JWT proof.
- **Draft 13 still works**: discovery detects which spec the issuer speaks and uses the matching request shape (jwt proofs only). All draft-specific code is isolated in `vci/draft13-compat.ts` and tagged `[draft-13-compat]`.
- Stored credentials now report the index `category` back to the caller, and employment-type credentials map to the wallet's real "Work History" category instead of a nonexistent one.

## Presentation (OID4VP)

- **DCQL** is the 1.0 query language and is the primary path; PEX (`presentation_definition`) is kept as `[pex-compat]` backward support.
- **DCQL `vp_token` values are arrays** per 1.0 §8.1 (`{"query_id": [vp]}`); draft 22 sent bare values. Reference verifiers accept both, strict 1.0 ones require arrays.
- **Unsigned Request Objects**: a `request_uri` may return plain JSON instead of a JWS — but only for the `redirect_uri` client-id prefix, which forbids signatures by construction. Any other prefix still requires a verified JWS (`vp/request-object.ts`).
- **Data Integrity VPs**: when the verifier's `client_metadata.vp_formats_supported` restricts `ldp_vp` to `DataIntegrityProof`, the wallet signs the VP with a mutually supported cryptosuite (e.g. `eddsa-rdfc-2022`) and `proofPurpose: authentication` instead of the legacy `Ed25519Signature2020` default.
- **VP `@context` matches its credentials** (`vp/context.ts`): embedding a VC 2.0 credential inside a v1-context presentation fails JSON-LD signing with "protected term redefinition", so the envelope now follows the embedded credentials.
- **DCQL type matching**: 1.0 sends `type_values` as expanded JSON-LD IRIs; held credentials store compact terms. The adapter now matches both.
- **Encrypted responses (JARM)**: negotiate `enc` from `encrypted_response_enc_values_supported` (default `A128GCM`) and take the JWE `alg` from the verifier's key, per 1.0. Old JARM field names still honored as a fallback.
- **`transaction_data`**: rejected with a clear error when present (we don't bind it yet), instead of silently presenting without it.
- **`nonce`** is now required in signed Request Objects.
- **Client-id prefixes**: added `decentralized_identifier:` (1.0 rename of `did`) and `x509_hash`.
- **`request_uri_method=post`**: POST a `wallet_nonce` to the request URI and require the verifier to echo it.

## Wallet app

- After claiming, "Open my <credential>" now lands on that credential's wallet category page (e.g. `/achievements`) instead of the dashboard; multi-credential claims go to `/passport` (`pages/oid4vci/Oid4vciExchange.tsx`).
- `BoostEarnedCard` no longer crashes the wallet grid when a record carries an unknown category — it falls back to Achievement styling.

## Test harnesses + playground

- The in-process e2e verifier (`tests/openid4vc-e2e/server/verifier.ts`) gained DCQL, JARM (`direct_post.jwt`), and transaction_data support, with new roundtrip tests so the 1.0 presentation path is covered on every PR (no Docker).
- The Sphereon strict harness (`tests/openid4vc-interop-e2e/setup/sphereon-verifier.ts`) validates the array-shaped DCQL `vp_token`.
- The e2e mock issuer was updated to the 1.0 request/response shape.
- The playground's EUDI reverse proxy and embedded issuer were fixed to serve metadata at both URL styles and to derive their base URL from the request host (they were hard-coded to a port).

## How to review

Easiest way to review:
- For VCI issuance flow: go to https://ler-tests.prettygoodskills.com/wallet/credential-acceptance/oid4 => Initiate Exchange => Go to dashboard in learncard app => Add to Passport => Use a Claim Link => Paste the exchange link here:

Screenshot 2026-07-16 at 5 05 54 PM

Then, complete the claim flow, and all the tests should pass from the prettygoodskills site! ✅ 

Screenshot 2026-07-16 at 5 04 39 PM

- For VCI presentation flow: go to https://ler-tests.prettygoodskills.com/wallet/credential-presentation/oid4 => Initiate Exchange => Go to dashboard in learncard app => Add to Passport => Use a Claim Link => Paste the exchange link here:

Screenshot 2026-07-16 at 5 05 54 PM

Then, complete the presentation flow, and all the tests should pass from the prettygoodskills site! ✅ 

Screenshot 2026-07-16 at 4 44 37 PM

- Issuance: start in `vci/metadata.ts` (the insert + version-detect), `vci/proof.ts` (proof-type selection + di_vp builder), and `vci/draft13-compat.ts` (the isolated compat, with removal notes at the top). `vci/request.ts` has the new body.
- Presentation: `vp/parse.ts` + `vp/request-object.ts` (new params, client-id prefixes, unsigned request handling, request_uri POST), `dcql/respond.ts` (array vp_token), `vp/encrypt.ts` (JARM negotiation), and the `transaction_data` guard + Data Integrity negotiation in `plugin.ts`.
- Backward-compat is all greppable: `[draft-13-compat]`, `[pex-compat]`, `[jarm-compat]`.
- Manual check: run the app and walk the LER test suite's issuance (`/wallet/credential-acceptance/oid4`) and presentation (`/wallet/credential-presentation/oid4`) flows at ler-tests.prettygoodskills.com — both should complete.
- Ignore `bun.lock` — version bumps from the `main` merge, no new deps.

## Testing

- Plugin unit suite: `bunx nx test openid4vc-plugin` — 586 pass.
- In-process HTTP e2e (`tests/openid4vc-e2e/`) — 12 pass, incl. DCQL / JARM / transaction_data.
- Docker interop (`tests/openid4vc-interop-e2e/`, walt.id + Sphereon + EUDI) — 23 pass / 3 skipped; validates Draft 13 issuance against a real vendor and strict 1.0 DCQL.
- Live: full issuance + presentation runs against the LER Interoperability Test Suite.

## Known follow-ups (not gaps — these are cleanly rejected today)

The `openid_federation` and `verifier_attestation` client-id prefixes, actually binding `transaction_data` into a presentation, and the Digital Credentials API response modes. Each needs its own infra/epic; verifiers using them get the spec-mandated rejection. A dockerized DCC-style verifier tier for the interop suite (to pin the di_vp / unsigned-request / Data-Integrity profile at the wire level) would also be a good addition.

### ✨ PR Description
Purpose: Update OpenID4VCI/OID4VP protocols from Draft 13 to final 1.0 spec with backward compatibility layer.

Main changes:
- OID4VCI: Metadata discovery inserts well-known between host/path (§12.2.2); credential requests use 1.0 shape with `proofs` array and auto-select `di_vp` key proofs
- OID4VP: Reject `transaction_data` requests, negotiate JARM encryption via `encrypted_response_enc_values_supported`, support `x509_hash` and `request_uri_method=post`
- Draft 13 compatibility isolated in `draft13-compat.ts` module for eventual removal once ecosystem migrates


Custard7:
@claude please review

Custard7:
Thanks for the review — addressed most of it:

- **Draft 13 issuance break:** you're right, the backwards-compat claim was overstated. Updated the PR description to call out that issuance is a 1.0-only cutover (only discovery/response-parsing stay tolerant), and flagged that the previously-tested partners need confirming on 1.0 before merge.
- **Changeset:** added (`.changeset/update-oidc-v1.md`, minor bump for `@learncard/openid4vc-plugin`).
- **Accept header:** added `Accept: application/json` to the authorization-server metadata fetch too, for consistency with the issuer fetch. Tests updated.

One correction on the `proofs` vs `proof` point: in OID4VCI **1.0 final** §8.2 there is no singular `proof` parameter — the request defines only `proofs` (an object whose value is a non-empty array). Singular `proof` existed in Draft 13 and was removed in 1.0 (the only `proof:` left in the spec is the unrelated W3C VP `di_vp` property). So `proofs` isn't the "riskier of two allowed shapes" — it's the only conformant one. An issuer that accepts only singular `proof` would itself be non-1.0, so I'd rather not send a non-conformant body speculatively. Happy to revisit if we hit a concrete issuer that requires it.

Custard7:
Follow-up: implemented Draft 13 backward-compat so this is no longer a hard cutover.

- Metadata discovery now detects the issuer's spec revision once (1.0 insert-style URL → `final`; append-style fallback → `draft-13`), and that drives the credential-request shape. Draft 13 issuers (e.g. walt.id under its `/draftNN` path) work again.
- All Draft-13 code is isolated in `vci/draft13-compat.ts`, every call site tagged `[draft-13-compat]`, so it's a clean one-pass removal later.
- Also updated the in-repo mock issuers (`tests/openid4vc-e2e/server/issuer.ts` → 1.0; playground `embedded.ts` → accepts both shapes) and added a changeset.

Plugin suite 545 green, in-process e2e 8 green, plugin type-checks clean. The Docker walt.id interop suite still needs a local run to confirm, but the fallback path is now in place for it.

Custard7:
Ran the full Docker interop suite against real walt.id + Sphereon + EUDI — **all green** (12 files, 23 passed, 3 intentional skips).

The Draft 13 fallback is validated end-to-end against a real Draft 13 vendor: walt.id `jwt_vc_json` issuance, auth-code, multi-credential, and roundtrip all pass via the append-style discovery + Draft 13 request body.

One fix landed while verifying: walt.id's **SD-JWT** issuance was 400ing because the Draft 13 SD-JWT profile requires the request to echo `vct` (not `credential_definition`), which the plugin never sent — a pre-existing gap, not a regression (the old code omitted it too; the byte-identical body only started getting exercised once interop ran). Draft 13 format-specific fields are now centralized in `draft13-compat.ts`: `credential_definition` for W3C VC formats, `vct` for SD-JWT VC formats.

Final state: plugin unit 546 pass, in-process e2e 8 pass, Docker interop 23 pass, type-check + build clean.

Custard7:
@claude please review

Custard7:
Thanks — addressed the actionable items from the latest review:

- **Path-less Draft 13 blind spot (🟡):** documented as an accepted limitation in `fetchIssuerMetadataDocument` (`metadata.ts`). I deliberately did **not** add the `nonce_endpoint` content-sniff tiebreaker: it's only a *1.0-positive* signal, but the problematic case (a path-less Draft 13 issuer) has no `nonce_endpoint` and is byte-indistinguishable from a nonce-less 1.0 issuer — so the heuristic can't disambiguate it and would just add a fragile guess. Given Draft 13 issuers serve under a path in practice (walt.id's `/draftNN`) and the ecosystem is converging on 1.0, defaulting path-less to `final` is the right call; documenting it beats a false-confidence sniff.
- **`originsMatch` misnomer (🟢):** renamed to `issuerIdentifiersMatch` and reworded the comment + the test title to reflect that it compares the full identifier (path included), not just the origin.
- **Embedded mock `token_type` (🟢):** now returns canonical `'Bearer'` so the playground exercises the standard casing.

Left as-is (agree they're not worth changing): the `metadata_fetch_failed`-only fallback gate (correct — don't paper over a misconfigured issuer), AS discovery not being spec-version-aware (robust; only a minor extra round-trip for draft-13), and the `insertWellKnown` catch fallback (unreachable with a valid identifier).

Verification: plugin unit 547 pass, in-process e2e 8 pass, playground type-checks, build clean.

Custard7:
@claude please re-review

Custard7:
Thanks — good catch on the third-pass review. Addressed:

- **🟡 transaction_data guard bypassable on the granular path** — you're right, it only lived in `presentCredentials`. Hoisted it into a shared `assertTransactionDataUnsupported(request)` and call it from both `presentCredentials` and `submitPresentation`, so the §8.4 rejection holds regardless of whether the wallet drives the all-in-one or the `prepare→build→sign→submit` flow. Added an e2e test asserting `submitPresentation` throws `invalid_transaction_data` and submits nothing.
- **🟢 `pickSupportedEnc` took the first advertised enc blindly** — now prefers the first advertised value the wallet can actually perform (intersect against a known JOSE `enc` set), falling back to the first advertised value, then the spec default. An exotic value listed first no longer forces a late `encrypt_failed`.
- **🟢 apv/apu ECDH-only** — added the one-line note that RSA-OAEP / AES-KW enc keys skip the apv nonce binding (ECDH is the expected path).

Left as-is (agree they're intentional): `verifier_info` is pass-through per §5.11, and the `x509_san_dns` port stripping is deliberate (RFC 6125).

Verification: plugin unit 558 pass, in-process e2e 12 pass (incl. the new granular-path guard test), build clean.

TaylorBeeston (changes requested):
## Findings

### 1. [P1] Malformed `transaction_data` in Request Objects bypasses the rejection guard

`packages/plugins/openid4vc/src/vp/request-object.ts:1136-1140` converts malformed `transaction_data` to `undefined`, while an empty array remains `[]`. The downstream guard in `plugin.ts:543-549` only rejects non-empty arrays.

Therefore signed or unsigned Request Objects containing `transaction_data: [42]`, `{}`, or `[]` proceed as though the parameter were absent. The wallet can submit a presentation without transaction binding—the exact behavior this PR intends to prevent.

A targeted runtime reproduction returned `{ hasTransactionData: true, value: undefined }`. Validate Request Object claims with the same strict rules as query parameters, or reject whenever the claim is present since the plugin supports no transaction-data type.

### 2. [P1] JWT proof selection ignores the issuer’s supported algorithms

`packages/plugins/openid4vc/src/vci/proof.ts:100-106` selects JWT solely because a `jwt` entry exists. It never intersects the JWT entry’s `proof_signing_alg_values_supported` with `signer.alg`.

The default LearnCard signer uses `EdDSA`. If an issuer advertises JWT with only `ES256` and `di_vp` with `eddsa-rdfc-2022`, this code selects JWT and sends an algorithm the issuer must reject, even though the wallet and issuer have a mutually supported `di_vp` path. A targeted invocation confirmed `selectKeyProofType` returns `{ proofType: "jwt" }` for that metadata.

Pass the active signer algorithm into selection and choose a proof type with a mutually supported algorithm. If none exists, fail before making the credential request.

### 3. [P1] Unsigned by-value requests can impersonate DID verifiers

`packages/plugins/openid4vc/src/vp/parse.ts:416-434` applies `deriveClientIdPrefix` to unsigned by-value authorization requests. Consequently, the new `decentralized_identifier:` handling in `vp/client-id-prefix.ts` normalizes the client to `did` and accepts it without invoking `verifyAndDecodeRequestObject`, where DID signature verification occurs.

An attacker can construct an unsigned deep link with `client_id=decentralized_identifier:did:example:trusted-verifier` and `response_uri=https://attacker.example/collect`. The wallet treats it as the DID verifier and can disclose selected credentials to the attacker-controlled endpoint. OID4VP 1.0 §5.9.3 requires `decentralized_identifier` requests to be signed with a key associated with that DID.

A targeted runtime reproduction returned `{ kind: "by_value", scheme: "did", responseUri: "https://attacker.example/collect" }`.

Reject unsigned by-value requests whose derived prefix requires a signed Request Object—at minimum `did`/`decentralized_identifier`, `x509_san_dns`, `x509_hash`, `https`, and `verifier_attestation`. Add regression coverage for both canonical `decentralized_identifier:` and legacy `client_id_scheme=did` forms.

### 4. [P2] The app still consumes the old issuer-metadata return shape

`fetchCredentialIssuerMetadata` now returns `{ metadata, specVersion }` at `packages/plugins/openid4vc/src/vci/metadata.ts:25-64`, but `apps/learn-card-app/src/pages/oid4vci/Oid4vciExchange.tsx:206` and `:260` still assign that wrapper directly to `CredentialIssuerMetadata`.

At runtime, `OfferConsent` cannot find `credential_configurations_supported`, and `pickIssuerInfo` cannot find `display`. Claiming can continue, but the consent and success screens lose credential metadata and issuer branding. This is also a TypeScript mismatch. A targeted reproduction confirmed `result.display` is `undefined` while `result.metadata.display` contains the issuer data.

Destructure `{ metadata }` at both callsites. Because this helper is publicly exported, also account for the breaking return-type change for downstream direct consumers.

### 5. [P2] Invalid by-reference `request_uri_method` values silently become GET

`packages/plugins/openid4vc/src/vp/parse.ts:165-169` maps any value other than lowercase `get`/`post` to `undefined`. The strict `parseRequestUriMethod` validation only runs for by-value requests, so a by-reference request containing `request_uri_method=POST` is fetched with GET instead of failing with `invalid_request_uri_method`.

A targeted runtime reproduction returned `{ method: "GET", resolved: undefined }`. Validate the raw outer parameter before selecting the fetch method; OID4VP 1.0 defines only case-sensitive `get` and `post`.

**Verdict:** changes requested. The targeted runtime probes reproduced all five issues; the unsigned DID by-value path is a security blocker.

Custard7:
Thanks for the thorough review — the targeted runtime probes were all accurate. All five findings were legitimate; all five are addressed in ee0e1fc06 (612 plugin tests pass, +26 new regression tests; in-process e2e 12 pass).

**1. `transaction_data` bypass (P1)** — Fixed. `buildRequestFromClaims` now applies the same strict validation as the URL-parameter path: any present-but-malformed `transaction_data` claim (`[42]`, `{}`, `[]`, bare string) throws `invalid_transaction_data` instead of being coerced to `undefined`. Regression tests cover all four malformed shapes plus the valid-array passthrough into the §8.4 guard.

**2. JWT proof selection ignores issuer algs (P1)** — Fixed. `selectKeyProofType(configDef, signerAlg)` now intersects the jwt entry's `proof_signing_alg_values_supported` with the active signer's alg: your exact scenario (jwt=ES256-only, di_vp=eddsa-rdfc-2022, EdDSA signer) now selects `di_vp`, and a jwt-only alg mismatch fails before the credential request with a precise error. Missing/empty alg lists stay lenient (several live issuers omit the field), and both call sites (`accept.ts`, `auth-code.ts`) pass `signer.alg`.

**3. Unsigned by-value DID impersonation (P1, security)** — Fixed, with one deliberate refinement to your suggested rejection set. Unsigned by-value requests now throw the new `invalid_client_id_scheme` for every prefix whose trust model requires a signature: `did` (canonical `decentralized_identifier:`, legacy `client_id_scheme=did`, and bare `did:` URIs), `x509_san_dns`, `x509_hash`, `verifier_attestation`, and explicit legacy `client_id_scheme=https`. Your repro (`client_id=decentralized_identifier:did:example:trusted-verifier` + attacker `response_uri`) is a covered regression test in both forms.

The refinement: a **bare** `https://...` client_id with no explicit scheme is downgraded to `pre-registered` rather than rejected. That's the correct semantics in both spec revisions — draft 22 defaults an absent `client_id_scheme` to pre-registered, and in 1.0 final a prefix-less client_id is pre-registered too (federation requires the explicit `openid_federation:` form; there is no bare-URL federation encoding in 1.0). Rejecting it would break the pre-1.0 installed base this PR explicitly preserves, while the downgrade removes the misleading signature-implying `https` label without dropping the request. The wallet UI's HTTPS trust affordance already excludes both labels, so nothing is presented as verified.

Also hardened the adjacent hole in the same surface: `redirect_uri` client-ids must now match the response target (§5.9.1 — under that prefix the identity *is* the redirect target). Canonical `redirect_uri:<value>` requires exact equality; the legacy `client_id_scheme=redirect_uri` form requires same-origin (deployed draft-22 verifiers routinely use distinct paths for client_id vs response_uri, so exact equality there would break real traffic). Enforced in both the by-value path and the unsigned-JSON `request_uri` path, with cross-origin spoof regression tests.

**4. App consumes old metadata shape (P2)** — Fixed. Both `Oid4vciExchange.tsx` callsites destructure `{ metadata }` now, so consent/success screens get `display` and `credential_configurations_supported` again. On the exported return-type change: it's intentional (the `specVersion` drives draft-13 request-shape selection) and rides the existing minor bump for `@learncard/openid4vc-plugin`.

**5. `request_uri_method` silent GET fallback (P2)** — Fixed. The by-reference path now routes through the same strict `parseRequestUriMethod` as by-value, so `request_uri_method=POST` throws `invalid_request_uri_method` before any fetch (regression test asserts no fetch is made).

New `invalid_client_id_scheme` code is mapped to trust-gap copy in the app's friendly-error layer. One caveat: I haven't re-run the Docker interop suite (walt.id/Sphereon/EUDI) since these changes — the in-process e2e verifier exercises the exact legacy `redirect_uri` shape those vendors use and passes, but I'll do one more interop run before merge.

TaylorBeeston (approved):
Re-reviewed the current head (`ee0e1fc`) after pulling it locally. All five issues from my earlier review are addressed:

- malformed `transaction_data` values are rejected instead of normalized away
- JWT proof selection respects the issuer's advertised algorithms and falls back to `di_vp` when appropriate
- unsigned by-value DID client IDs are rejected for both canonical and legacy schemes
- the wallet call sites consume the current issuer-metadata return shape
- invalid `request_uri_method` values fail before any fetch

Verification: 612/612 plugin tests passed, 12/12 in-process OpenID e2e tests passed, the OpenID4VC package build passed, and the current-head GitHub workflows (including OpenID4VC Interop) are green.

*Non-blocking asterisk:* I did not personally rerun the separate live LER / full third-party Docker vendor matrix; the repository interop workflow and local in-process round trips are green and cover the reviewed fixes.

Looks good — approving.

Custard7:
Addressed all three items from the gitStream AI review in 97c7e417d (plugin suite 613 pass, in-process e2e 12 pass):

- **Misleading nonce error on unsigned Request Objects** — valid. `buildRequestFromClaims` now throws `VpError('missing_nonce')` with a transport-neutral message instead of `RequestObjectError('invalid_request_object', 'Signed Request Object …')`. This also aligns the unsigned-JSON path with the URL-parameter parser (same code, same friendly copy), matching the existing precedent of `VpError('invalid_transaction_data')` from the same function. One note: rejecting a nonce-less unsigned request is itself correct and stays — `nonce` is required in both draft 22 and 1.0 (§5.2) for replay protection, and the by-value path has always enforced it; only the error type/message changed. Regression tests added for both the signed and unsigned paths.
- **Bare `crypto.subtle` in `base64UrlSha256`** — valid. Now guarded the same way as `randomWalletNonce` and throws a typed `RequestObjectError` when Web Crypto is unavailable, instead of an untyped `ReferenceError` escaping `verifyWithX509`.
- **Unvalidated `x-forwarded-proto` in the playground EUDI proxy** — valid. Whitelisted to `https`/`http` (anything else falls back to `http`).

PR #1423, opened 2026-07-17
build: shrink Docker build context (4.5GB → ~200MB)

goblincore:
## Problem

Local Docker builds (`docker compose -f compose-local.yaml up --build`) were spending **400s+ on `load build context`** before any build step ran — the context had grown to **~4.5GB** on my local machine....

The culprit was `.dockerignore` coverage. It excluded `node_modules`/`dist`/`.git`, but several large trees that accumulated over time were still being tarred up and shipped to the daemon on every build.

## Fix

Add exclusions to the root `.dockerignore`:

| Added | Why |
|-------|-----|
| `.claude`, `.worktrees` | Git worktrees (Claude Code + manual) hold full source checkouts — these had grown to ~24GB across ~10 worktrees |
| `apps/*/ios`, `apps/*/android` | Capacitor native projects. Their untracked `DerivedData`/`Pods`/`.gradle` caches are multiple GB and are **never used by the server/web Docker builds** (no Dockerfile references ios/android/capacitor) |
| `~` | Defensive guard against a stray literal `~/` dir in the repo root — a classic unquoted-`~` scripting accident. The Docker context is **not** git-aware, so such a dir would otherwise be shipped. (The one found in this repo has been deleted.) |

Also removes the now-unused **`.vexp/`** entry from `.gitignore` — a local tool workspace that's no longer used (folder deleted).

## Impact (measured locally)

| | Before | After |
|---|--------|-------|
| Build context | ~4.5 GB | **~200 MB** |
| `load build context` | ~424 s | **~73 s** |

No functional change — build inputs are unchanged; only unused files are kept out of the context.


smurflo2 (approved):
Confirmed builds are still working for me!

TaylorBeeston (approved):
Pulled and re-verified the current head (`2bd5f4c`). This looks good.

What I checked:

- audited every Dockerfile / root-context workflow for native iOS or Android build inputs
- validated the new patterns with a Docker-compatible `.dockerignore` parser: they exclude the intended root worktree/native paths without swallowing tenant assets or similarly named nested paths
- recreated the effective context with all four `apps/*/{ios,android}` trees absent
- ran LearnCard's tenant-config preparation successfully from that filtered context
- built LearnCard successfully (12,213 modules) and Scouts successfully (7,996 modules) from the filtered context
- confirmed the current-head GitHub lint, tests, CodeQL, preview, and package-contract workflows are green

The automated concern about unanchored `.claude` / `.worktrees` patterns matching at every depth does not reproduce under Docker's matching semantics; these no-slash patterns are root-scoped.

*Non-blocking asterisk:* this environment has no Docker CLI/daemon, so I could not independently repeat the full `docker compose up --build` or the reported 4.5 GB → 200 MB timing measurement. I verified the actual filtered inputs and both affected inner web builds instead.

Approved.

PR #1430, opened 2026-07-21
feat: [LC-2053] Feedback Prompts

Custard7:
# Overview

#### 📚 What is the context and goal of this PR?

We have no in-app way to learn how key flows feel to users. This adds a lightweight feedback pilot to the LearnCard app: a one-tap "How was that?" prompt (frown / meh / smile) that appears at three success moments — issuing a credential, claiming via a boost/claim link, and claiming via OIDC. Answers land in our existing analytics, so we can see per-flow sentiment and response rates and use them to drive product decisions.

#### 🥴 TL; RL:

Tiny inline sentiment prompt on issue/claim success screens. One tap to answer, negative answers get an optional "what went wrong?" follow-up. Heavily rate-limited so nobody gets nagged.

Screenshot 2026-07-23 at 3 01 21 PM

Screenshot 2026-07-23 at 3 01 28 PM

Screenshot 2026-07-23 at 3 01 47 PM

#### 💡 Feature Breakdown (screenshots & videos encouraged!)

Everything lives in `apps/learn-card-app/src/feedback/`:

- **SentimentStrip** — the inline prompt. One tap records the answer and morphs to "Thanks!". Rendered on the issue success screen (all three variants: link / send / self), the OIDC claim success card (`OfferFinished`), and inside the claim success toast for flows that navigate away on success (`ClaimBoost`, `ClaimFromDashboard`, `ClaimFromRequest`).
- **FeedbackFollowUpSheet** — bottom sheet shown after a frown/meh answer. Reason chips (Confusing / Slow / Something broke / Other) plus an optional free-text field. "Something broke" or a note also files to Sentry via `captureFeedback`, so bug context reaches engineering.
- **feedbackGovernor** — persisted rate-limiter. Max 1 prompt per session, 3 per week, 30-day cooldown per surface after an answer, 90-day mute after 2 ignores. Also holds a ledger reserved for a future native app-store review prompt (not in this PR).
- **useFeedbackEligibility** — privacy gate. Child profiles, minors (via `getMinorAgeThreshold`), and users with `analyticsEnabled: false` never see prompts. Free text is only collected when `bugReportsEnabled` allows it.
- **Analytics** — four new typed events in `src/analytics/events.ts`: `feedback_prompt_shown`, `feedback_sentiment_given`, `feedback_followup_submitted`, `feedback_followup_dismissed`. Shown vs. given gives per-surface response rate — the metric that tells us if a prompt is annoying.
[...]

PR #1442, opened 2026-07-30
feat: [LC-2064] App Review Prompt

Custard7:
# Overview

#### 🎟 Relevant Jira Issues

N/A

Screenshot 2026-07-30 at 2 16 04 PM

Follow-up to the in-app feedback PR (#1430). That one collects sentiment; this one acts on it.

Users who have already told us a flow felt good now get asked to advocate: on iOS/Android they get the native OS review prompt, on web a small dismissible "star us on GitHub" card. We had no in-app path to store reviews at all before this.

The reason to gate on sentiment is practical, not just polite. Apple hard-caps review prompts at 3 per user per 365 days and gives us no way to know whether a prompt was even shown. Those asks are scarce and non-refundable, so spending one on someone who just hit a bug is a real waste. The sentiment data we already collect tells us who to skip.

#### 🥴 TL; RL:

If you've had a good experience, used the app a few times, and it's been at least a week, you might see the OS review prompt after your next successful claim or issue. On web you get a GitHub star card instead. Max twice a year.

**The compliance rule that shapes the whole design.** Google explicitly forbids asking the user any question before or while the review UI is shown, and Apple disallows custom review prompts (guideline 5.6.1). So the obvious version of this feature — tap 🙂, immediately show the review prompt — is not allowed.

What is allowed: use previously recorded sentiment as invisible eligibility, then fire the plain OS prompt cold at a later value moment. No question, no "Rate us" button, nothing gated. That's what this implements, and the 7-day settle window in the eligibility rules is what guarantees the ask lands in a different session from the sentiment tap. If someone later "optimises" that window to zero, we're in violation — there's a comment saying so.

**Eligibility** (`resolveAdvocacyDecision` in `apps/learn-card-app/src/feedback/feedbackGovernor.ts`) — returns a typed reason rather than a boolean so it's debuggable. All of: 3+ sessions, at least one recorded positive, 7+ days since that positive, no negative in the last 30 days, 90-day cooldown, max 2 per year.

**Why 2 per year and not Apple's 3** — the OS never tells us whether a dialog appeared, so our count can drift from theirs. One ask of headroom absorbs that.

**Native path** (`useAdvocacyPrompt.ts`) — dynamic import + try/catch inside a `Capacitor.isNativePlatform()` guard, matching the existing Haptics pattern. Fires 1.8s after the success screen so it doesn't land on top of the success animation. No UI of our own.

**Web path** (`GitHubStarCard.tsx`) — dismissible card, only for the `learncard` tenant. VetPass users shouldn't be asked to star someone else's repo.

**One prompt at a time** (`FeedbackMoment.tsx`) — the sentiment strip's 30-day surface cooldown and advocacy's 7-day window aren't mutually exclusive, so a user could have hit "How was that?" *and* an OS review dialog on the same screen. All call sites now go through `FeedbackMoment`, which renders exactly one of the three states.

**Analytics** — `store_review_requested` plus three GitHub card events. The event catalog says in a comment that this records the *attempt* and can never be read as a conversion, because neither platform reports the outcome. Worth knowing before someone builds a "reviews driven" chart out of it.

**Deployment targets** — checked, no changes needed. Plugin needs iOS 15 / minSdk 24; we're on iOS 15.6 and minSdk 26.

- **We cannot measure whether this works.** No conversion signal exists on either platform. The only telemetry is that we asked.
- **Only wired to the two success screens that persist** (issue success, OIDC claim success). The boost/claim-link flows toast and navigate away, so a delayed native prompt would be killed by the unmount. Those keep the plain sentiment strip.
- **Fixed a latent crash while here.** The governor store is persisted, and anyone who already has it in localStorage from #1430 has a `review` object without the new `requestLog` field. Persist merges shallowly, so that field hydrates as `undefined` and `.filter()` on it would have thrown for every existing user. All reads go through a `readRequestLog()` helper now. Worth remembering for future schema changes to this store.

#### 🔍 Types of Changes

-   [x] No
-   [ ] Yes

# Testing

**Use the Feedback tab in the debug panel** (bug icon, bottom corner). It exists because this feature is otherwise close to untestable by hand: the eligibility windows are measured in days, the per-session cap lives in memory, and on native the OS decides whether a dialog appears and never tells us. Available in dev and on the staging deploy preview.

The tab shows live governor state plus the `resolveAdvocacyDecision()` reason, and has buttons for **Reset**, **Make advocacy-eligible**, **Recent negative**, **Fill yearly cap** and **Expire cooldowns**. No reload needed — the prompt reads the governor when a success screen mounts.

Be aware of the platform limits before you start:

- **iOS dev builds always show the review dialog**, whatever the eligibility says. Proves wiring, nothing about gating.
- **TestFlight shows nothing at all.** Not a bug.
- **Android internal test ignores quotas.**
- `requestReview()` resolves the same whether a dialog appeared or not, so there is no success signal anywhere.

Web is therefore the only place the logic is observable.

Screenshot 2026-07-30 at 2 14 44 PM

**Sentiment strip**
1. Feedback tab → *Reset all feedback state*
2. Issue a credential to yourself → "How was that?" appears under "Issue Another"
3. Tap 🙂 → morphs to "Thanks!"
4. Issue another in the same session → no strip (1-per-session cap)
5. Tap 😞 instead on a fresh reset → follow-up sheet with reason chips; ticking "Something broke" or leaving a note also files a Sentry feedback event

**Advocacy / GitHub card**
1. *Make advocacy-eligible* → decision badge should read `eligible`
2. Hit a success screen → the **GitHub card** renders *instead of* the sentiment strip (they're mutually exclusive)
3. Dismiss it → trigger success again → no card (90-day cooldown)
4. *Recent negative* → card stays away (`recent_negative`)
5. *Fill yearly cap* then *Expire cooldowns* → decision should read `yearly_cap`, not `cooldown`
6. Navigate away within ~1s of the card appearing → budget is **not** spent

**RTL** — switch to العربية with the card showing: the dismiss X must sit top-**left** and must not overlap the body text.

**Native** — install to a simulator and check only two things: the dialog lands *after* the success animation rather than on top of it, and if it fails to appear the sentiment strip comes back rather than leaving a dead screen.

**PostHog** — `feedback_prompt_shown` vs `feedback_sentiment_given` gives the per-surface response rate. On `store_review_requested`, if `asksThisYear` ever exceeds 2 the governor has a bug.

iOS Native, Android Native, Chromium, Safari

#### 🧪 Code Coverage

No new tests. The eligibility resolver is pure and would unit-test well — happy to add that if you'd rather not take it on trust. E2E isn't viable here since the whole feature is built to not fire.

Verified with `tsc` (no new errors; repo total went from 2039 to 2006 after merging main) and clean LSP diagnostics on all changed files.

# Documentation

**Internal/AI Docs**

-   [x] **Code comments/JSDoc** — Complex logic that needs inline explanation

**Visual Documentation**

#### 💭 Documentation Notes

No API or SDK surface changes. The store-policy constraints and the persisted-store migration hazard are documented inline in `src/feedback/`, which is where anyone changing this code will be looking.

# ✅ PR Checklist

### ✨ PR Description
Purpose: Implement app store review prompts and GitHub star advocacy asks gated by sentiment and session thresholds to drive user advocacy.

Main changes:
- Added advocacy decision engine with sentiment eligibility rules, session requirements, and yearly cap enforcement in feedbackGovernor store
- Created native StoreKit/Play review prompt integration via Capacitor with web fallback GitHub star card component
- Enhanced feedback governor to track positive/negative sentiment, session counts, and advocacy request logs with time-based cooldowns
- Replaced SentimentStrip with FeedbackMoment wrapper component that conditionally shows advocacy asks or sentiment prompts across success screens


Custard7:
Thanks — two of the three were real, fixed in ff6cd74.

**Issue 2 (`asksThisYear` bypasses the year filter) — fixed.** Now reads `readRequestLog(feedbackGovernorStore.get.review()).length`, so the 365-day filter is applied at the read site. This also removed the `?? 1` fallback, which was dead code that would have quietly under-reported if it ever were reachable.

**Issue 3 (web budget consumed before the card is seen) — fixed, with one deliberate deviation.** The budget is now committed once the card has been mounted for a moment, or immediately on click/dismiss, instead of on mount. Navigating away before it renders long enough to be seen no longer costs one of the two yearly asks.

I did *not* move to recording only on click/dismiss, though. A card the user simply ignores has to consume the ask — otherwise it reappears on every success screen forever, which is a worse outcome than occasionally losing one ask to a fast navigation. The commit helper documents this so it doesn't get "fixed" later.

**Issue 1 (migration drops advocacy history, resetting the yearly cap) — not applicable, no change made.**

The reasoning is sound in general, but the premise doesn't hold here: the `review` ledger shipped in #1430 was never written to. On `main` it appears only in the type, the initial value, and a comment saying it's reserved for future use — there was no code path that incremented `requestCount`. So every existing user has `requestCount: 0` and `lastRequestedAt: undefined`.

That means there's no history to lose and no cap to reset, because nobody has ever been prompted — the feature being capped didn't exist until this PR. Apple's 3-per-year limit can't be breached on upgrade for the same reason.

The `?? []` guard in `readRequestLog` is still load-bearing, but for a different reason than the cap: without it, `undefined.filter()` throws for anyone carrying the old persisted `review` object, which would have crashed the success screen for every existing user. Adding changeset-style migration machinery for a field that has only ever held its initial value would be dead code.

Also added the missing changeset.

Custard7:
@claude please review

Custard7:
All six addressed in 41561f4. Thanks — #2 and #3 were both real and I'd missed them.

**#1 Capgo channel — bumped to `1.0.10`**, no exemption taken. You framed the exemption as arguable, but I don't think it holds: the guard's own wording carves out "a plugin patch release with no native API change," and this adds an entirely new plugin with new native code. `UNIMPLEMENTED` being caught means it degrades rather than crashes, which is a reason the bump is *safe to defer* in an emergency, not a reason it isn't needed.

**#2 Native failure left the user with nothing — fixed.** `setAdvocacyActive(false)` in the catch, so the screen falls back to the sentiment strip. You're right this wasn't hypothetical; it was the visible half of #1.

**#3 Destructive effect re-run — fixed.** `track` is out of the dependency array behind a ref, and the timer now lives in `timerRef` with cleanup moved to a mount-only effect. A provider swap mid-delay no longer clears a timer it won't re-arm. Good catch on the inconsistency with `SentimentStrip` — it defends against exactly this and I didn't carry the pattern across.

**#4 `sessionCount` counted runtimes, not visits — fixed.** This was the most consequential one: as written, `ADVOCACY_MIN_SESSIONS = 3` was close to unreachable on native, so the feature was largely dead on the platform it exists for. `recordSession` is now gap-gated (4h) in the store and `useRecordFeedbackSession` also fires on `appStateChange` → active. Native resumes count; web reloads collapse. I used a 4h gap rather than the 24h "returning" notion so a genuine morning-and-evening user isn't stuck.

**#5 RTL — fixed.** `pe-9` / `text-start` / `end-2`. Straightforward miss against a documented convention.

**#6:**
- Dead `?.` on `useTenantConfig()` — dropped.
- `FeedbackMoment` docstring corrected; it now says it covers success *screens* and names `ClaimFeedbackToast` as the deliberate exception.
- Advocacy now calls `consumeSessionPrompt()`, so the OS dialog and a claim-toast strip can't both land in one session. Worth closing — it was a guarantee I'd claimed in the PR body.
- **Tests added** — 16 cases in `feedbackGovernor.test.ts`: all seven `resolveAdvocacyDecision` outcomes, the cooldown-before-yearly-cap ordering, entries ageing out of the rolling year, the legacy `review` object without `requestLog`, and the session/sentiment ledgers. Ordering assertions are commented as load-bearing so they don't get "simplified."
- **Interop CI is unrelated.** All 3 failures are `EUDI /ui/presentations returned 400: MissingRegistrationCertificate` from the external EUDI reference verifier via `tests/openid4vc-interop-e2e/setup/eudi-client.ts` — a missing registration certificate on their side, not a protocol regression. 11 of 13 test files pass, and the only OID4VCI file this PR touches is `OfferFinished.tsx`, where the change is one additive JSX line.

`tsc` back to the 2006 baseline, unit suite green.

One note on #4 for the record: the gap threshold is now the thing that decides how quickly a user becomes advocacy-eligible, and it's unobservable in production like the rest of this feature. If 4h turns out to be too generous we'd only find out via `asksThisYear` climbing faster than expected.

Custard7:
Added a **Feedback tab** to the debug panel in f2e72a8, and rewrote the QA section of the description to use it.

The console-surgery steps I'd written were fragile and nobody would remember them in three months. This repo already solved that problem — the Pathways, Offline and InAppMessages tabs exist for the same reason — so this follows `PathwaysDebugTab`'s pattern and is gated by the same `WIDGET_ENABLED` (dev, or `VITE_ENABLE_AUTH_DEBUG_WIDGET=true`), meaning it works on the staging preview but never ships to production.

It shows live governor state, the `resolveAdvocacyDecision()` reason string, and per-surface prompt eligibility, with buttons for Reset / Make advocacy-eligible / Recent negative / Fill yearly cap / Expire cooldowns. Verifying the advocacy path is now two clicks instead of six console steps.

Making the decision function return a typed enum rather than a boolean paid off here — the tab renders the reason directly, so "why am I not seeing the prompt" is answerable at a glance.

One thing that came out of building it: `resetForDebug()` also clears the module-scope `sessionPromptCount`. Clearing localStorage by hand leaves that counter set, which is the most likely reason someone concludes the prompt is broken when it's working exactly as designed. Worth knowing if you QA'd this before now.

Custard7:
Found a real timing bug while double-checking the native path. Fixed in f6d01a0.

The advocacy ask on native is an **interruptive OS modal** (StoreKit sheet / Play bottom sheet), and "1800ms after the success screen mounts" was too loose a trigger for that. Four ways it could have landed badly:

1. **It could fire over the wrong page.** Ionic keeps visited pages mounted in the router outlet, so the unmount cleanup never ran on forward navigation — `useHeaderScrollSync` already documents exactly this ("the unmount cleanup below does NOT fire on forward navigation"). A user tapping "View in Wallet" 1.2s into the delay would have got the review modal over the **wallet page**. Now cancelled on `useIonViewWillLeave`, following that same precedent.

2. **It could land mid-celebration.** Confetti runs to ~2050ms (250ms stagger + 1300–1800ms fall) and the card reveal takes 700ms; the delay was 1800ms. Now 2400ms, so the ask arrives after the reward, not on top of it.

3. **Backgrounding left it armed.** Now cancelled on `appStateChange` when the app goes inactive.

4. **The claim-link branch hands off to a native share sheet** and leaves the user mid-task copying a link. It no longer arms the ask at all — the sentiment strip still renders there.

Cancelling records nothing, so an abandoned ask stays available instead of burning one of only two per year. A module-scope latch also stops two cached success screens arming duplicate prompts.

The net effect: the OS prompt now only fires if the user lands on a success screen and **sits there** past the celebration, still foregrounded, without moving on. That's a genuinely idle post-value moment rather than "some time after render."

Worth noting why this mattered more than it looks: because neither platform reports whether the dialog was actually shown, a mis-timed ask is **silently wasted** — fired while the user is looking at a different page, or suppressed behind another presenter, and we'd never know. With a cap of 2/year the cost of a loose trigger is real. Stricter firing conditions were the right trade even though they mean the ask fires less often.

Eligibility rules are unchanged.

Custard7:
@claude please do a final re-review, and approve for merge

Custard7:
All four fixed in ba38a1a. #1 and #2 were both real and both mine — thanks for catching them.

**#1 VetPass web losing the strip permanently — fixed.** I verified the premise before fixing: `environments/tenant-registry.json` maps `vetpass.app`, `alpha.vetpass.app` and `staging.vetpass.app` to `tenantId: vetpass`, so this was hitting real users. The tenant check now runs before any arming side effect:

```ts
if (!allowAdvocacy || firedRef.current || advocacyArmed || !privacyEligible) return;
if (!isNative && !isLearnCardTenant) return;   // before arming
if (resolveAdvocacyDecision() !== 'eligible') return;
```

The permanence was the nasty part — because nothing is recorded on that path, the user stays `eligible` forever, so it isn't a one-off, it recurs every session.

**#2 The fallback was inert — fixed.** You're right that last round's `setAdvocacyActive(false)` achieved nothing while the session budget stayed spent; the strip just rendered `null` instead. Added `releaseSessionPrompt()` and called it from both the `requestReview()` catch and `cancelPending`. That makes the invariant explicit: **arming spends the prompt, and anything that fails to present an ask hands it back.**

**#3 Expire cooldowns — fixed.** It now backdates the recorded asks rather than replacing them (`readRequestLog(review).map(() => stale)`), so the count survives and `Fill yearly cap` → `Expire cooldowns` reports `yearly_cap`. Worth noting the QA step in the description was only wrong *because* the button was wrong — fixing the button makes the documented step correct, so I've left the description as-is.

**#4 Reset couldn't clear the latch — fixed.** Two parts: `commitCardAsk` now releases the latch (once the ask is recorded, `resolveAdvocacyDecision` returns `cooldown`, so the latch has no further job), and `resetAdvocacyLatchForDebug()` is exported and called from the reset handler. "No reload needed" is true again.

**#5 Test gap — closed.** Six cases on `consumeSessionPrompt` ↔ `canPromptForFeedback`, including cross-surface suppression, release restoring the prompt, the clamp at zero, and the debug reset clearing it. 22 tests total, all passing.

You were right that a three-line case would have caught both. The lesson I'm taking: I tested the decision resolver thoroughly because it was the part that *looked* intricate, while both actual bugs were in the boring counter interaction between two modules. Worth remembering that seams between components are where the risk concentrates, not the component with the most branches.

`tsc` at the 2006 baseline, LSP clean.

gerardopar (approved):
LGTM 🚀

Screenshot 2026-08-04 at 10 12 15 AM

PR #1444, opened 2026-07-30
[LC-1958] Fix EdDSA Data Integrity interoperability

TaylorBeeston:
# Overview

#### Relevant Jira Issues

Fixes: LC-1958

#### What is the context and goal of this PR?

LearnCard's vendored SSI/DIDKit stack was missing four pieces of W3C EdDSA Data Integrity behavior: URL issuers triggered DID resolution, undefined JSON-LD terms were silently dropped, `cryptosuite` did not expand as `cryptosuiteString`, and minimal VC 2.0 documents failed expansion. The current verifier also did not enforce dependency-aware `previousProof` chains.

This updates the merged SSI fix from [learningeconomy/ssi#3](https://github.com/learningeconomy/ssi/pull/3), adds native regression coverage, rebuilds the native and WASM integrations, and makes the HTTP bridge usable as the local W3C VC-API endpoint.

#### TL;DR

The current W3C EdDSA suite now passes all 44 issuer, verifier, interop, data-model, and algorithm assertions against LearnCard's HTTP bridge.

#### Feature Breakdown

- Rejects JSON-LD data loss instead of issuing a credential after dropping undefined terms.
- Embeds the current Data Integrity v2 context, including the typed `cryptosuite` mapping.
- Verifies URL issuers against the proof's verification method without trying to resolve the issuer as a DID.
- Validates missing, self-referential, cyclic, and transitive `previousProof` dependencies.
- Issues minimal `https://www.w3.org/ns/credentials/v2` credentials.
- Adds six native regressions covering the four Jira failures and proof-chain behavior.
- Rebuilds the checked-in WASM bindings and publishes the optimized binary at a content-addressed, immutable CDN URL. The prior unversioned object remains available for older clients.
- Adds a local bridge runner and W3C suite configuration, while keeping serverless deployment on the package-provided WASM by default.
- Updates existing credential fixtures with explicit extension contexts now required by strict JSON-LD processing.

#### Important tradeoffs made

Strict expansion intentionally rejects credentials that rely on undefined terms. Existing fixtures now include the defining Open Badges, LearnCard Boost, or W3C examples context instead of depending on the old silent-drop behavior.

For URL issuers, there is no DID document that can authorize the proof verification method. The verifier still validates the signature, skips only the DID relationship check, and returns an explicit warning that issuer authorization was not checked.

For proof chains, normal proof-free verification remains the first path. The verifier only retries the predecessor-set document form used by the W3C fixtures after the standard form fails, so ordinary single-proof credentials keep their existing verification behavior.

The public WASM URL includes the complete SHA-256 and uses `Cache-Control: public,max-age=31536000,immutable`. New payloads can be published alongside old ones without changing what existing package versions load.

#### Types of Changes

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Chore

#### Does This Create Any New Technical Debt?

- [x] No
- [ ] Yes

# Testing

#### How Can Someone QA This?

1. Initialize the updated SSI submodule: `git submodule update --init --recursive`.
2. From the repository root, run the six native regressions with `(cd lib/ssi && cargo test --manifest-path ../../packages/plugins/didkit-plugin-node/native/Cargo.toml)`.
3. Force the native release build with `env -u CI BUILD_DIDKIT_NAPI=1 bun --cwd packages/plugins/didkit-plugin-node run build`, then run `bun --cwd packages/plugins/didkit run build` and `bun --cwd packages/learn-card-bridge-http run build`.
4. Follow `packages/learn-card-bridge-http/README.md` to start the local bridge and point the W3C `vc-di-eddsa-test-suite` at it.
5. Run `npx mocha tests/ --timeout 15000`; all 44 assertions should pass.

Observed locally:

- Native Rust regressions: 6 passed.
- Native N-API release compilation: passed.
- DIDKit WASM package build: passed.
- HTTP bridge build: passed.
- W3C EdDSA conformance suite: 44 passed.
- Focused managed-Docker Data Integrity E2E: 8 passed, 11 unrelated tests skipped.
- Browser smoke test loaded the content-addressed CDN WASM with HTTP 200 and generated an Ed25519 key.

Observed in GitHub Actions on commit `f8ac13924`:

- Test Runner: passed, including the full affected unit-test graph.
- DIDKit native prebuilds: passed for Linux x64, Linux arm64, macOS x64, and macOS arm64.
- Mock E2E, lint, CodeQL, and published-package contract checks: passed.
- OpenID4VC interop: 20 passed and 3 EUDI-only assertions failed because the upstream EUDI verifier now returns `MissingRegistrationCertificate`; the Walt.id and Sphereon paths passed.

Focused follow-up regressions:

- Credential library: 144 passed.
- AI session credentials: 9 passed.

The local repository-wide commands were also exercised, but are not green for failures outside this change:

- `CI=true NX_DAEMON=false bunx nx affected --target=test --base=origin/main --head=HEAD` reran after the strict-context fixes and failed only in `network-brain-service:test`: 55 `consentflow.spec.ts` setup-hook timeouts plus 9 `profiles.spec.ts` invite failures (two stale `Challenge already in use` errors and seven 5-second timeouts). That suite uses a reusable local Neo4j container; the same affected Test Runner passed on a fresh GitHub Actions runner.
- `bunx nx test:run e2e` completed 633 tests successfully, with 13 skipped, but exited nonzero because `holder-continuity-export.spec.ts` exceeded its existing 5-second timeout and Vitest reported 19 asynchronous `notFound`/DID resolution errors from other suites. The changed credentials and bitstring suites passed in that run.

#### Which devices would you like help testing on?

No device coverage needed; this is Rust, WASM, Node native, and HTTP API behavior.

#### Code Coverage

The native wrapper now has one regression for each Jira sub-task plus W3C and transitive proof-chain fixtures. Existing E2E coverage was updated for strict context expansion, and the external W3C suite exercises the HTTP behavior end to end.

# Documentation

#### Documentation Checklist

**User-Facing Docs**

No user-facing API or app flow changed.

**Internal/AI Docs**

- [x] AGENTS.md
- [x] Code comments/JSDoc

**Visual Documentation**

- [ ] Mermaid diagram

#### Documentation Notes

`AGENTS.md` now contains the reproducible native-to-WASM rebuild command. The bridge README documents local conformance setup, the package-provided WASM default, and the optional local artifact override.

# PR Checklist

- [x] Related to a Jira issue
- [x] Code follows style guidelines
- [x] Common end-to-end cases were manually tested by OMP
- [x] Code was reviewed by OMP
- [x] Ambiguous proof-chain compatibility behavior is commented
- [ ] All existing unit and E2E tests pass locally; infrastructure and unrelated async failures are documented above
- [x] Documentation checklist is complete
- [x] Product analytics considered; not applicable to this non-UI change

### Ready to squash-and-merge?

- [x] Valid credential behavior is backwards compatible, and the previous WASM artifact remains available
- [ ] There is no Do Not Merge label on this PR
- [x] Security implications were considered
- [x] This does not expose a new unauthenticated production endpoint

— OMP

#### Numeric Bitstring Status List compatibility follow-up

- Accepts the spec-defined string index and the non-standard unsigned JSON number emitted by 1EdTech's conformance credentials.
- Keeps SSI/DIDKit serialization canonical as a decimal string and rejects negative or fractional numbers.
- Makes `credentialStatus.id` optional, updates LearnCard's validator and helper detection, and adds focused native and TypeScript regressions.
- The rebuilt local WASM validated 1EdTech `ob-434` and `ob-436` as active and `ob-435` as revoked.

#### Rebased onto #1220 (DIDKit publishing automation)

`main` now carries #1220, which extracted the default WASM location into `DEFAULT_DIDKIT_WASM_URL` so the publisher can rewrite it independently of how Prettier wraps the `init` signature. The merge resolved `src/didkit/index.ts` in favor of that constant, carrying this branch's URL as its value — `set-default-wasm-url.mjs` matches this file again, so the automated publisher can update it.

`check-wasm-url.mjs` now reads `DEFAULT_DIDKIT_WASM_URL` directly and accepts both content-addressed layouts: the nested `/didkit/sha256-<sha>/didkit_wasm_bg.wasm` form the publish workflow writes, and the flat `didkit_wasm_bg-<sha>.wasm` form already serving this branch's artifact. Both patterns are anchored so a stray 64-hex string elsewhere in the URL cannot satisfy the guard; the binary-vs-URL sha equality check is unchanged.

The artifact is already published and verified: the URL returns HTTP 200 with `content-type: application/wasm` and 9,027,278 bytes, and its sha256 matches the checked-in `pkg/didkit_wasm_bg.wasm` (`7fec99f6…dee3c7`). It is deliberately **not** republished under the nested key — both layouts are immutable and content-addressed, the invariant that matters is "default URL sha == checked-in wasm sha", and the path converges to the nested form on the next publisher run.


TaylorBeeston:
@claude pls review

TaylorBeeston:
Addressed the review follow-up in `fce0e5728`.

- Kept the W3C-required URL-issuer behavior, but documented that `errors: []` proves signature integrity—not issuer authorization—and mapped the warning to the stable `Issuer Authorization` check label.
- Verification now checks every proof, retries proof-chain documents with direct predecessors only, and rejects bad proofs instead of returning the first success.
- Replaced recursive dependency traversal and panic-prone context parsing; scoped nested-context compatibility to the proof-chain retry path.
- Added the `w3.org` VC 1.1 context alias, removed the duplicate `did:key` recovery VM, and replaced the CLR examples catch-all with explicit vocabularies.
- Memoized bridge initialization, pinned the rebuilt WASM by SHA-256, packaged it as an explicit Serverless asset, and rejected unknown bridge proof options.

SSI follow-up: learningeconomy/ssi#5

Verified:
- native DIDKit: 8 / 8
- focused SSI proof-chain / proof-property regressions: 3 / 3
- credential library: 145 / 145
- LearnCard verification adapter: 5 / 5
- bridge API TypeScript check and Serverless package: passed
- packaged Lambda archive contains `src/didkit_wasm_bg.wasm`; invoking the extracted handler returned the expected DID
- local bridge smoke: HTTPS issuer verifies with `checks: ["proof"]`, both VC 1.1 context URLs issue, minimal VC 2.0 issues, and undefined terms return `DATA_LOSS_DETECTION_ERROR`

The repository-wide affected run still reaches the existing `network-brain-service:test` failures: 55 consent-flow setup-hook timeouts plus stale invite challenges/timeouts. Managed E2E startup is blocked locally because the existing `lcn-neo4j` container owns port 7474; unmanaged E2E has no signing service on port 4200. I left the existing local service stack untouched.

-- OMP

Custard7:
### 1. The default WASM URL is stale, and the glue this PR ships cannot load it

`fce0e5728` rebuilt **both** the binary and the wasm-bindgen glue, but `didkit/index.ts` was last touched in `0eefb20a3` and still points at the previous build:

| | sha256 | bytes |
|---|---|---|
| Checked-in binary @ HEAD | `4d256eb6…6277` | 9,025,492 |
| What the default URL serves | `3fca0a34…c52a` | 9,084,001 |

wasm-bindgen closure-wrapper import names are generated per build. I downloaded the CDN object and link-tested it against the glue in this PR:

```
OLD CDN binary (what default init() downloads):
  LinkError: Import #109 module="wbg" function="__wbindgen_closure_wrapper12353":
  function import requires a callable
NEW local binary: instantiates OK
```

The old binary imports `__wbindgen_closure_wrapper12353`; the new glue only defines `__wbindgen_closure_wrapper12304`. Missing import means a hard `LinkError`, not degraded behavior.

**This is why the `Test` check is red** — `init:test` fails on all 3 retries with the byte-identical error:

```
init:test > LearnCard SDK > emptyLearnCard > should work
Caused by: WebAssembly.instantiate(): Import #109 module="wbg"
  function="__wbindgen_closure_wrapper12353": function import requires a callable
```

Blast radius, stated precisely: the first-party apps are **fine** — `learn-card-app` and `scouts` (via `learn-card-base`) import the binary from `dist/` with Vite `?url`, so glue and binary stay paired. What breaks is **external SDK consumers in the browser** calling `initLearnCard()` without a `didkit` argument, which is the documented quick-start path, shipped here as a patch to `@learncard/didkit-plugin`.

I also confirmed `4d256eb6…` returns **HTTP 403** — it was never published — so this can't be fixed by just repointing the URL. It needs the protected publisher, exactly as the PR description says. Note the internal contradiction that makes this unambiguous: `scripts/sync-didkit.ts:4` pins `4d256eb6…` while `didkit/index.ts:11` still serves `3fca0a34…`.

### 2. `lib/ssi` is pinned to an unmerged PR branch

```
lib/ssi: cf8928ea -> 246e896e
learningeconomy/ssi: main...246e896e = ahead_by=2, behind_by=0
ssi PR #5 -> state=OPEN, head=246e896e
```

The pin *is* the head of an open PR, and those 2 commits carry all the hardening the checked-in WASM and native prebuilds were compiled from. Since this repo squash-merges, merging ssi#5 will orphan `246e896e`, at which point `git submodule update --init` breaks for everyone and the shipped binary is unreproducible from any reachable source.

### 3. Genuine silent-accept regression — in 2 places, not 4

An attacker fully controls the `issuer` field of a credential they mint, so `{"issuer": "https://real-org.example/issuer"}` signed with their own key now returns `checks: ["proof"], errors: []` plus a warning.

- **Exploitable — `ler-rs.ts:423,450`**: `verified: errors.length === 0` flowed into `VerifySharedResume`, so a forged employment/education credential rendered as verified. **Fixed below.**
- **Exploitable — bridge `app.ts:125`**: `/credentials/verify` returns HTTP 200; integrators gating on status or `errors.length === 0` accept a self-minted credential. **Left to you** — see below.
- **Overstated, not exploitable — `workflows.ts:351` and `contact-methods.ts:106`**: both authenticate a *presentation holder* (challenge/domain-bound VP proof; `contact-methods` additionally requires `isTrustedLoginProviderDID()`). Neither consults a credential `issuer`, and the non-DID relaxation doesn't propagate to VP proofs. I'd push back on the earlier review flagging these.

The type-level docs added in `VerificationCheckValidator` are good, but documentation isn't sufficient mitigation when in-repo consumers already machine-interpret `errors.length === 0` as verified.

---

## What I pushed

**`77af5c78` — Reject unauthorized issuers in LER-RS verification**
Treats warnings as disqualifying and surfaces them alongside errors, matching the convention already used by `learn-cloud credential.helpers`, brain-service DID-auth, and chapi. Adds `warnings?: string[]` to `VerificationResult`, threads the reason through `shared-resume.helpers.ts` (otherwise the UI showed a bare "verification failed." with the explanation stranded in `warnings`), and adds 4 regression tests. Confirmed meaningful: **2 of the 4 fail against the pre-fix code**, while the valid-credential and error-credential controls pass under both, so there's no over-rejection.

**`df073be5` — Guard the DIDKit WASM artifact against URL drift**
The sha-pin assertion asked for in the earlier review. **It currently fails, by design** — it converts the opaque `init:test` LinkError into:

```
The checked-in DIDKit WASM does not match the default init() URL.
  checked-in pkg/didkit_wasm_bg.wasm: 4d256eb6...
  URL in src/didkit/index.ts:          3fca0a34...
To resolve: publish the checked-in binary as
  https://assets.learncard.ai/didkit_wasm_bg-4d256eb6....wasm
then update the default URL in src/didkit/index.ts to that sha.
```

It goes green the moment the binary is published and the URL updated, and prevents this recurring.

One heads-up: `ler-rs.ts` has unrelated reformatting. That's unavoidable, not carelessness — `lint.yml:48` runs Prettier only over files changed by the PR, and this file was already non-compliant before it entered the changed set.

Custard7:
Follow-up — correcting my own fix, and confirming CI signal is restored.

My first cut at the WASM guard was a jest test inside `didkit-plugin`, which was the wrong home for it. Because `didkit-plugin` sits early in the dependency graph, failing there bailed the nx test run and **tripled the suppressed suites from 10 to 30** — including `credential-library`, `learn-card-base`, `learn-cloud-*`, `lca-api-*`, and the ler-rs regression test I'd just added. A stale release artifact shouldn't cost the review its test signal.

Relocated in `1cc43cb7` + `b637a965` to a standalone pure-Node lint job, mirroring the existing `check-i18n-keys` scan. Verified both directions: exits 1 with the actionable message today, and exits 0 once the URL names the checked-in binary.

CI state now, on `b637a965`:

| Check | Result |
|---|---|
| `Check DIDKit WASM matches its published URL` | **fail** in 11s — names the exact URL to publish |
| `Test` | fail, and the only failed task is `init:test` — the pre-existing LinkError |
| Skipped suites | back to **10** (was 30) |
| `ler-rs-plugin:test` | **pass** — the URL-issuer regression runs in CI |
| `Prettier` | pass |

So both blockers are each surfaced by one precise, independent check, and nothing unrelated is masked. `Test` goes green with the WASM publish; the guard job goes green with the URL update in the same step.

Custard7:
Retracting part of my own change — I overstepped, and the fix was wrong.

**Reverted in `24c1d4cf`.** `ler-rs.ts`, `types.ts`, `shared-resume.helpers.ts`, and the changeset are now byte-identical to `b3fd1ff21`. My remaining footprint on this branch is only the two WASM-guard files.

Two separate problems with what I pushed:

**1. It wasn't this PR's code.** This PR never touches `ler-rs` or the resume-builder app files. My reasoning was that the PR *causes* the regression through `ler-rs` without editing it — the SSI change makes URL issuers return `errors: []`, and `ler-rs` decided validity on `errors.length === 0`. That causal chain is real. But patching a file outside the diff also dragged in unrelated Prettier reformatting (`lint.yml` only checks changed files, and `ler-rs.ts` was already non-compliant), expanded the release surface via the changeset, and touched an app file in a PR that states no app flow changed. It also contradicted my own reasoning: I declined to patch `/credentials/verify` because it needs the `nonDidIssuerPolicy` design decision, then patched a different consumer of that same default.

**2. More importantly, the fix was incorrect.** Blanket "any warning disqualifies" over-rejects badly here. `learn-card-network/src/plugin.ts:2741` pushes a **warning** for any boost credential whose issuer is outside the trusted registry — and `DEFAULT_TRUSTED_BOOST_REGISTRY` (`:2645`) contains exactly one entry, `did:web:network.learncard.com`. So my change would have flipped essentially every boost credential in a LER resume to `verified: false`. `:2707` does the same for missing Boost ID metadata.

My tests didn't catch this because they mocked only the non-DID-issuer warning and a hard error. My claim that "the controls pass under both, so there's no over-rejection" was overconfident — I never exercised a benign-warning case. Thanks for pushing on it.

**The underlying issue still stands** and still needs fixing before merge, just correctly:

- Gate on the **specific** non-DID-issuer warning, not on `warnings.length`, since warnings here are a mixed-severity channel; or
- better, fix it at the verification API layer (`nonDidIssuerPolicy: 'error' | 'warn'`, default `'error'`) so `ler-rs`, the bridge, and external SDK callers are all correct by default and no call-site patching is needed.

The second is cleaner and is your design call. Happy to implement it in a separate PR against whichever shape you pick.

The WASM URL guard (`1cc43cb7`, `b637a965`) stays — that one is squarely about artifacts this PR does modify, and it's still correctly red.

TaylorBeeston:
Reviewed the numeric Bitstring Status List follow-up and fixed the missing generated-artifact step in `2ffb848ed`.

- The SSI compatibility change is scoped to `BitstringStatusListEntry`: canonical decimal strings still serialize unchanged; unsigned JSON integers are accepted; negative and fractional values remain rejected; older status-list formats remain string-only.
- Rebuilt and optimized the WASM from SSI `246e896`, copied the matching glue/binary into `didkit-plugin`, synchronized the bridge copy/hash, and added a direct checked-in-payload smoke check.
- Focused verification passed: SSI deserializer (1/1), native DIDKit regression (1/1), LearnCard helper regressions (4/4), local WASM payload smoke, and the new CI `Check DIDKit WASM payload behavior` job.

One release blocker remains intentionally visible: `Check DIDKit WASM matches its published URL` is red because `7fec99f61754f80762b56b9a4a6bd1ed2a36d26266b9f7c44f735d26b4dee3c7` is not published yet. The protected publisher needs to upload:

`https://assets.learncard.ai/didkit_wasm_bg-7fec99f61754f80762b56b9a4a6bd1ed2a36d26266b9f7c44f735d26b4dee3c7.wasm`

Then `packages/plugins/didkit/src/didkit/index.ts` can move to that content-addressed URL. Updating the URL before the object exists would turn the current explicit CI failure into a production HTTP 403, so I left it unchanged.

-- OMP

TaylorBeeston:
Artifact blocker is closed in `c924c6d3f`.

- Published the committed 9,027,278-byte WASM to `learncard-didkit-wasm` with `Content-Type: application/wasm` and `Cache-Control: public,max-age=31536000,immutable`.
- Public URL is live: `https://assets.learncard.ai/didkit_wasm_bg-7fec99f61754f80762b56b9a4a6bd1ed2a36d26266b9f7c44f735d26b4dee3c7.wasm`.
- Updated the default `init()` URL to that content address.
- Instantiated DIDKit through the public default URL and exercised numeric `BitstringStatusListEntry.statusListIndex` handling against the downloaded payload.
- All GitHub workflow runs for `c924c6d3f` passed, including the URL/payload guards, package tests, native prebuilds, OpenID4VC interop, CodeQL, Mock E2E, and package-contract validation.

The remaining merge-order item is SSI PR #5: merge it first, then repoint `lib/ssi` to the resulting `main` commit if the squash merge changes the pinned SHA.

-- OMP

Custard7:
### Pre-merge action item: re-pin `lib/ssi` once learningeconomy/ssi#5 merges

This branch pins `lib/ssi` to `246e896`, which currently exists **only on ssi#5's branch** — `ahead_by=2` of ssi `main`. Once ssi#5 merges it will produce a new SHA, and if the branch is then deleted, `246e896` stops being reachable from `main` and `git submodule update --init` starts failing repo-wide. So the last step before merging this PR is moving the pointer to a commit on ssi `main`.

**It's a pure pointer move — no WASM rebuild, no republish, no URL change.** I verified the unmerged remainder of ssi#5 (`246e896 → 5472079`) contains zero release code:

| File | Why it can't affect the build |
| --- | --- |
| `ssi-json-ld/src/lib.rs` | All three changed hunks (862, 897–910, 924) sit inside `mod test`, which is `#[cfg(test)]`-gated at line 696 |
| `ssi-json-ld/Cargo.toml` | Sole addition is in `[dev-dependencies]` (line 34; section starts at 28) — dev-deps don't apply to a path dependency's release build |
| `.github/workflows/build.yml`, `.gitignore`, `Cargo.lock` | CI/tooling only. ssi's own workspace lock is ignored when consumed via `path = "../../ssi"` |

So the checked-in `pkg/didkit_wasm_bg.wasm` (`7fec99f6…dee3c7`) stays valid, `DEFAULT_DIDKIT_WASM_URL` stays as-is, and both WASM guards keep passing.

**Checklist**

- [ ] Merge learningeconomy/ssi#5
- [ ] Re-pin: `git -C lib/ssi fetch origin main && git -C lib/ssi checkout <new-ssi-main-sha>`, then commit the submodule pointer
- [ ] Confirm `Check DIDKit WASM matches its published URL` and `Validate DIDKit WASM build` stay green (they should be unaffected)

<sub>Separately, and not blocking on this: `packages/plugins/ler-rs/src/ler-rs.ts:423`/`:450` still derive `verified` from `errors.length === 0` while ignoring warnings, so a credential with a URL `issuer` signed by an unrelated key now reports `verified: true`; and the changeset ships as `patch` even though strict JSON-LD expansion turns previously-succeeding issuance into a hard `DATA_LOSS_DETECTION_ERROR` for third parties on the HTTP bridge. Both are discussed in review.</sub>

Custard7 (approved):

TaylorBeeston:
Final DIDKit/SSI refresh is now on this branch:

- Ran the #1220 publisher with DIDKit `d418013f6b9b6d22de6bdeb02df753b9c38517ca` and merged SSI #5 (`3ea772a3d0a3dbda51b3ca94211e1b44bcb46515`).
- Published and verified WASM SHA-256 `60efb6027699fda135db0d817f7a541bbd660ae88bec9cbec9796feeb3740b96` at `https://assets.learncard.ai/didkit/sha256-60efb6027699fda135db0d817f7a541bbd660ae88bec9cbec9796feeb3740b96/didkit_wasm_bg.wasm`.
- Applied the generated submodule/bindings/WASM output from auto PR #1455 to this branch in `7be22e2a38e4c2db2a2cdf6ee2c9c6311a8aba51`.
- Closed #1455 without merging it; `main` was not changed. The temporary automation branch and helper workflows were removed.

The older `7fec99f6…` artifact paragraph in the description is superseded by this update. Fresh CI is running on clean head `f65a1fe619cb9860fc483315ef3a12dd5c83286f`.

PR #1451, opened 2026-08-04
[LC-1958] Add explicit issuer authorization verification

TaylorBeeston:
## What

Separates cryptographic proof verification from issuer authorization for credentials with HTTP(S) issuers.

- `checks: ["proof"]` continues to verify the W3C VC-API vector and returns the existing issuer-authorization warning.
- `checks: ["proof", "issuerAuthorization"]` resolves the issuer's controlled identifier document and requires the proof key to be authorized for its proof purpose. Missing or unauthorized documents fail closed.
- The option is exposed through LearnCard's verification types and the HTTP bridge validator.
- Native and WASM bridges now pass caller-supplied issuer and verification-method documents into SSI's resolver chain. Remote resolution remains opt-in through `allowRemoteContexts`; otherwise only statically supplied documents can satisfy the check.

The rebuilt 9,076,341-byte WASM is published at `assets.learncard.ai` under SHA-256 `da82d7c815e8a3bd14ed9d8b866918a67cc3f03d34daca22ef41e4b9febe4be7`, and the bridge copy plus default `init()` URL are pinned to that payload.

## Verification

- SSI URL-issuer authorization regression
- Native DIDKit URL-issuer regression
- WASM authorized, warning-only, unauthorized, and W3C URL-issuer + `did:key` smoke cases
- HTTP bridge: proof-only returns 200 with the warning; explicit authorization fails closed when no issuer document is available
- DIDKit plugin and bridge builds/tests
- Public asset returns 200 with the expected hash, length, content type, cache policy, and instantiates as DIDKit `0.4.0`
- CI runs the full affected and e2e suites

## Stack

- Targets #1444 so the new issuer policy stays separate from the current Data Integrity and numeric status-list fixes.
- SSI implementation: learningeconomy/ssi#6
- DIDKit WASM resolver wiring: learningeconomy/didkit#8 (draft until SSI #6 lands, because DIDKit CI intentionally builds against SSI `main`)

## Security note

The W3C VC-API assertion only requires successful proof verification. It does not establish that an HTTP(S) issuer authorized the signing key. Callers that need that trust decision must explicitly request `issuerAuthorization` and handle failures; the proof-only warning is not an authorization result.


TaylorBeeston:
Local verification update:\n\n- SSI and native/WASM issuer-authorization regressions pass, including the exact W3C URL-issuer + `did:key` vector.\n- `didkit-plugin`, `didkit-plugin-node`, and `create-http-bridge` plus their build dependencies pass uncached Nx builds.\n- Public WASM fetch/instantiate passes with the committed SHA, length, MIME type, and cache policy.\n- The full managed e2e run passed 633 tests (13 skipped). Its only failure was the existing 5-second timeout in `holder-continuity-export.spec.ts`; the test completed in 6.2 seconds. Rerunning that exact test with `--testTimeout=10000` passed in 7.61 seconds. The two API-key timeout flakes from the first, concurrent run passed on the isolated run.\n- The repo-wide affected run made it through the package tests shown in its log but stalled in `network-brain-service:test`; I cancelled the orphaned Vitest/Neo4j run rather than presenting it as green. The directly affected package builds and behavioral checks are green.\n\nLeaving this draft until the stacked SSI/DIDKit dependencies and CI are green.

PR #1452, opened 2026-08-04
fix: [LC-1969] use solid category icons in the passport activity feed

goblincore:
# Overview

#### 🎟 Relevant Jira Issues

Fixes: [LC-1969](https://welibrary.atlassian.net/browse/LC-1969)

**Basically the issue here is that the current design changes based on theme - but it seems according to the ticket and figma they should always use the "solid" icons from the "neutral" mode theme:**

LC-1969 asks for the passport activity feed to use the **solid** version of each wallet category icon. Checking prod, it still doesn't — and Katie flagged the same thing on iPhone in the ticket comments. Nothing for LC-1969 had actually been committed (`git log --grep=LC-1969` comes back empty), so the feed has been rendering the pre-redesign icons all along.

The interesting part is *why* it looked half-fixed: which icon you get depends on the theme selector.

`ActivityCredentialIcon` resolved `IconWithShape ?? Icon`. The `colorful` set defines `IconWithShape` — the illustrated art (star-on-a-podium, rosette-on-a-square) — so that won. The `formal`/neutral set has no shaped variant, so it fell through to `Icon`, which *is* the flat solid glyph design specced. Same feed, two different icon families depending on a setting unrelated to this surface.

#### 🥴 TL; RL:

The activity feed picked the illustrated category art in the colorful theme and the solid glyph in the neutral theme. Now it always picks the solid glyph.

https://www.loom.com/share/10a8584d53274585aa46901af15f6d89

Adds an explicit `IconSolid` slot to the category icon table and points the feed at `IconSolid ?? Icon`, with `IconWithShape` deliberately out of the chain:

| Icon set | Declares `IconSolid`? | Feed renders |
| --- | --- | --- |
| `colorful` | yes — the existing `*IconFormal` components | solid glyph ✅ |
| `formal` | no — its `Icon` already *is* that glyph | solid glyph ✅ (unchanged) |
| `vetpass` | no — inherits `formal` | its own branded `Vetpass*Icon` |

The solid components already exist in `packages/learn-card-base/src/svgs/walletsIconsFormal/`; this PR wires them up rather than adding new assets. Verified against the two Figma nodes on the ticket — Boosts is `#3B82F6` solid rosette with a white heart, Achievements is `#F43F5E`, both exact matches.

`IconSolid` is also registered on `CategoryIconsSchema`. The icon table passes through `validateThemeData`, and Zod **strips undeclared keys**, so without that line the slot would silently be `undefined` at runtime — the same failure mode that dropped `navbar.dashboard` in LC-1921. There's a regression test pinning it.

**Two drive-bys, both surfaced by this work:**

1. **`ACTIVITY` heading casing.** LC-1898 swapped the literal `ACTIVITY` for `m['passport.activity.heading']()`, which returns `"Activity"`. The `h3` kept its `tracking-[1px]` eyebrow styling but lost the casing, so it stopped matching the month group labels right beneath it. Uppercased in JS via `toLocaleUpperCase(getLocale())` rather than a CSS `uppercase` class, matching how `monthLabel` in `activityFeed.helpers.ts` already does it — CSS `text-transform` isn't locale-aware and this string ships in es/fr/ar.
2. That casing change is also why **`PassportActivityFeed.test.tsx` has been red on `main`** since LC-1898. It's green again.

- **Added a 4th icon slot instead of hardcoding the solid set in the component.** A direct `import { FormalWalletIcons }` in `ActivityCredentialIcon` would have been a 2-line diff, but it bypasses the theme system and would have handed VetPass generic icons in place of its branded ones. Going through the icon table keeps tenant theming intact and gives future surfaces that want a flat glyph somewhere to hang it.
- **Only `colorful` declares `IconSolid`.** `formal` and `vetpass` fall through to `Icon`, which is already correct for both. Declaring it redundantly on every set would have tripled the diff for no behavior change — at the cost that if someone later gives `formal` an `IconWithShape`, this stays correct only because `IconWithShape` isn't in the resolution chain. That's called out in a comment at the call site.
- **VetPass's feed changes** from `Vetpass*WithShape` to `Vetpass*Icon` (its single-glyph variant). That's the consistent reading of the ticket — flat glyph everywhere — but it is a visual change for that tenant. The feed is behind `enablePassportActivityFeed`, so blast radius is small.
- **The ticket's second ask — enlarging the arrow between the avatar and the category icon — is skipped**, per Donny's own comment on LC-1969: that arrow no longer exists in the shipped feed.

#### 🔍 Types of Changes

-   [ ] No
-   [x] Yes

Minor, and pre-existing rather than introduced: the Assistance/IDs hex mismatch above wants a design decision, and `PortfolioIconFormal` has a `0 0 43 44` viewBox where every sibling is `0 0 51 50`, so it renders slightly larger relative to the others at a fixed box size. Both already affect the neutral theme today; happy to file follow-ups if you'd rather not fold them in here.

# Testing

1. Open the passport page with `enablePassportActivityFeed` on.
2. Confirm every activity row shows a **flat, single-colour** category glyph — no illustrated podium/square backgrounds.
3. Toggle the theme selector between **colorful** and **neutral**. The icons should now be identical in both; before this change they differed.
4. Confirm the section heading above the feed reads **ACTIVITY**, matching the casing of the `AUGUST 2026` month labels below it.
5. Worth a pass across a few categories — Badge, Achievement, Course, ID — since each maps to a different component.

iOS Native — Katie hit the original report on iPhone 15 / iOS 26.2.1, so a confirm on device would be ideal. Chromium is enough for the theme-toggle check.

#### 🧪 Code Coverage

12 new unit tests, all green, plus the 86-test `src/pages/wallet` + `src/theme` suites:

- `ActivityCredentialIcon.test.tsx` (new, 5 tests) — solid glyph wins over shaped art; falls back to `Icon` and **never** to `IconWithShape`; generic-credential and no-icons paths still render the fallback glyph; sizing `className` passes through.
- `icons.validators.test.ts` (new, 3 tests) — `CategoryIconsSchema` preserves `IconSolid` through Zod. This is the guard against the LC-1921 silent-strip failure, which is invisible at compile time.
- `PassportActivityFeed.test.tsx` — no longer failing.

No e2e added: this is a presentational swap with no new flow. I have **not** run it on a device — the verification above is unit tests plus a static render of all 12 solid glyphs at the feed's 38px size, checked against the Figma nodes. Worth a real-device look before merge.

`tsc --noEmit` error count is unchanged from `main` (2003 → 2003, all pre-existing); none reference `IconSolid`.

# Documentation

**User-Facing Docs**

-   [ ] **Tutorial**
-   [ ] **How-To Guide**
-   [ ] **Reference**
-   [ ] **Concept**
-   [ ] **App Flows**

**Internal/AI Docs**

-   [ ] **AGENTS.md**
-   [x] **Code comments/JSDoc**

**Visual Documentation**

-   [ ] **Mermaid diagram**

#### 💭 Documentation Notes

No user-facing docs — this changes which existing icon renders, not any flow or API. The two non-obvious bits are documented inline instead: why `IconWithShape` is excluded from the resolution chain (at the call site in `ActivityCredentialIcon`), and why `IconSolid` must be listed on `CategoryIconsSchema` or Zod drops it (at the schema).

# ✅ PR Checklist

-   [ ] I have **manually tested** common end-2-end cases — see Code Coverage; not yet verified on device
-   [x] I have considered **product analytics** for user-facing features — none needed, no new user action

[LC-1969]: https://welibrary.atlassian.net/browse/LC-1969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### ✨ PR Description
Purpose: Replace illustrated category icons with solid flat glyphs in passport activity feed for improved readability at small sizes (20-38px).

Main changes:
- Modified `ActivityCredentialIcon` to prioritize `IconSolid` glyph over `IconWithShape` with fallback to `Icon`
- Added `IconSolid` property to `CategoryIcons` type and theme icon definitions with formal glyph reuse
- Updated activity feed heading to use locale-aware uppercase conversion via `toLocaleUpperCase(getLocale())`
- Added comprehensive unit tests validating icon selection logic and schema validation


gerardopar (approved):
LGTM! 🚀

Image

PR #1454, opened 2026-08-04
fix: [LC-1964] YouTube embeds fail with Error 153 on iOS native

goblincore:
# Overview

#### 🎟 Relevant Jira Issues

Fixes: [LC-1964](https://welibrary.atlassian.net/browse/LC-1964)

YouTube videos attached to credentials fail to play in the native iOS app, showing **"Error 153: Video player configuration error"** instead of the player.

Capacitor serves the iOS app from `capacitor://localhost`. WKWebView will not send a custom-scheme URL as an HTTP `Referer`, and YouTube's embed endpoint rejects requests without a valid `http(s)` referrer — it returns an `UNPLAYABLE` player that renders as Error 153.

Verified directly against `https://www.youtube.com/embed/<id>`:

| `Referer` sent | Result | Context |
| --- | --- | --- |
| *(none)* | error screen | |
| `capacitor://localhost` | **error screen** | **iOS native** |
| `https://localhost` | plays | Android native |
| `https://learncard.app` | plays | web |

So this was never specific to the demo-sync portfolio badge — **every YouTube embed in the native iOS app was broken**. Android and web were unaffected.

#### 🥴 TL; RL:

iOS's `capacitor://` origin means YouTube embeds get no valid referrer and refuse to play. Where an embed can't work, we now show the poster frame with a play button that opens the video in the in-app browser instead of a broken iframe.

**AI Pathways already used this external-open approach (`AiPathwayContentPreview.tsx`); this brings the rest of the app in line with it.**

#### 💡 Feature Breakdown

https://github.com/user-attachments/assets/60224148-086b-4e3d-b95d-6938b075f665

- **`canEmbedVideoIframe()`** (`@learncard/react`) — tests the actual condition YouTube cares about (is this document served over `http(s)`?) rather than sniffing for Capacitor, which keeps a native dependency out of the SDK package. Also means it degrades correctly in any future non-`http(s)` host.
- **`ExternalVideoFallback`** — poster frame + play button, rendered in place of a doomed iframe. Hands playback off via a pluggable opener.
- **`setExternalUrlOpener()`** — seam so the SDK stays dependency-free (`window.open` default). `learn-card-base` registers Capacitor's `Browser.open` at startup, so native opens the in-app browser rather than bouncing the user to Safari.
- Applied at all three YouTube embed sites: `Lightbox` (credential views), `MessageMedia` (in-app messages), `BoostMediaPreview` (boost CMS preview).

- **Playback leaves the inline player on iOS.** The user taps play and lands in the in-app browser rather than watching inside the lightbox. The alternative that preserves inline playback — proxying the embed through a page hosted on an `https` LearnCard domain so the nested iframe sends a valid referrer — adds a hosted asset, a runtime dependency on the web domain, and per-tenant setup. Deferred; happy to file a follow-up if we want inline playback back.
- **Not** changing Capacitor's `iosScheme` to `https`. That would fix the origin globally but changes the WebView origin for every install, dropping existing localStorage/IndexedDB and invalidating auth/CORS assumptions. Far too invasive for this bug.

#### 🔍 Types of Changes

-   [ ] No
-   [x] Yes

Inline iOS playback is not restored — only handed off to an external browser. See the tradeoff note above; worth a follow-up ticket if design wants inline playback on iOS.

# Testing

1. Run the app in the iOS Simulator (or on device).
2. Open a credential with a YouTube video attachment — e.g. the video badge from the demo sync, in Portfolio.
3. Tap the video. **Before:** black lightbox with "Error 153: Video player configuration error". **After:** poster frame with a play button; tapping it opens the video in the in-app browser.
4. Confirm the web app is unchanged — video still plays inline in the lightbox.
5. Android should also be unchanged (it serves from `https://localhost`, which YouTube accepts).

iOS Simulator / iOS Native (iPhone 15, iOS 26.x — the reporter's device) / Android Native / Chromium

#### 🧪 Code Coverage

Unit tests added in `packages/react-learn-card/src/components/Lightbox/__tests__/Lightbox.video.test.tsx`:

- `canEmbedVideoIframe()` returns `true` for `http:`/`https:` and `false` for `capacitor:`
- `getExternalVideoUrl()` builds a watch URL from YouTube metadata, falls back for other platforms
- `Lightbox` renders the inline iframe on the web
- `Lightbox` renders no iframe and routes to the registered external opener when embeds can't work

Full `@learncard/react` suite passes (98 tests). `learn-card-base` passes 527/528 — the one failure (`dateHelpers` `calculateAge`) is a pre-existing timezone-dependent test that fails identically on `main`. `tsc --noEmit` on `learn-card-app` produces an identical error set before and after this change (no new type errors).

# Documentation

**User-Facing Docs**

-   [ ] Tutorial
-   [ ] How-To Guide
-   [ ] Reference
-   [ ] Concept
-   [ ] App Flows

**Internal/AI Docs**

-   [ ] AGENTS.md
-   [x] Code comments/JSDoc

**Visual Documentation**

-   [ ] Mermaid diagram

#### 💭 Documentation Notes

Bug fix with no API or user-flow change. The non-obvious part — *why* a custom-scheme origin breaks YouTube embeds — is documented as JSDoc on `canEmbedVideoIframe()` and `setExternalUrlOpener()`, which is where someone would actually look.

# ✅ PR Checklist

-   [ ] I have **manually tested** common end-2-end cases
-   [x] I have considered **product analytics** for user-facing features

---

[LC-1964]: https://welibrary.atlassian.net/browse/LC-1964?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### ✨ PR Description
Purpose: Fix YouTube embeds returning Error 153 on iOS native by detecting incompatible document schemes and falling back to external browser playback.

Main changes:
- Added `canEmbedVideoIframe()` helper detecting document protocol to identify when YouTube embeds fail due to missing HTTP referrer on Capacitor's custom schemes
- Implemented `ExternalVideoFallback` component with play button overlay that opens videos in external browser instead of iframe when embeds cannot work
- Registered Capacitor's in-app browser as external URL opener during app startup to keep users in native app during playback


goblincore:
Thanks — went through the three findings. Two were real and are fixed in 2b18c80; the third doesn't hold up.

### 🔒 `javascript:` scheme in `openExternalUrl` — fixed

Agreed, and worth taking seriously: these URLs come from credential attachments issued by arbitrary third parties, so they're untrusted input. `openExternalUrl` now resolves the URL and drops anything that isn't `http(s)` before it reaches the opener. `noopener` would not have prevented execution.

Covered by a test asserting `javascript:`, `data:` and empty URLs never reach the registered opener.

### 🐞 Empty external URL for non-YouTube video types — fixed, but at the root

The suggested patch guarded the empty-string symptom. The actual problem was that my gate was **too broad**: I applied `canEmbedVideoIframe()` to every video platform, when only YouTube rejects embeds over a missing/non-http referrer.

Checked the other players' endpoints with a `capacitor://localhost` referrer:

```
https://player.vimeo.com/video/76979871    capacitor=200   https=200
```

Vimeo serves the player regardless of referrer. So gating it — and Drive and Loom — was an unnecessary regression that would have pushed working inline embeds out to an external browser.

`canEmbedVideoIframe` now takes an optional `platform` and returns `true` immediately for anything that isn't YouTube. That also removes the empty-URL path entirely: the fallback only renders for YouTube, where `getExternalVideoUrl` always has a `videoId` to build a watch URL from. Callers that know the platform pass it; the no-arg form stays conservative.

### 🐞 `height: '100vh'` on the fallback wrapper — not applying

This one is based on a misreading. The claim is that "the iframe path does not apply an explicit height override of this kind", but it does — [`BoostMediaPreview.tsx:200`](https://github.com/learningeconomy/LearnCard/blob/2b18c8070/apps/learn-card-app/src/components/boost/boostCMS/BoostPreview/BoostMediaPreview.tsx#L200) sets `height: '100vh'` on the iframe, along with the same `width: '100%'` and `backgroundColor: '#353E64'`. The wrapper was written to match its sibling deliberately, so the two paths render at the same size. Changing only the fallback is what would make them inconsistent.

Happy to revisit if the 100vh treatment is wrong for this panel generally — but that would be a pre-existing issue with the iframe path, not something this PR introduces.

apps/learn-card-app/ios/App/Podfile
```diff
@@ -11,6 +11,7 @@ install! 'cocoapods', :disable_input_output_paths => true, :deterministic_uuids
 def capacitor_pods
   pod 'Capacitor', :path => '../../../../node_modules/@capacitor/ios'
   pod 'CapacitorCordova', :path => '../../../../node_modules/@capacitor/ios'
+  pod 'CapacitorCommunityInAppReview', :path => '../../../../node_modules/@capacitor-community/in-app-review'
```
TaylorBeeston: Can we revert this and the corresponding generated Android entries in `capacitor.build.gradle`, `capacitor.plugins.json`, and `capacitor.settings.gradle`? These look like incidental output from running a Capacitor sync while testing. They are not part of LC-1964, and registering the in-app-review plugin changes the native surface and is what trips Capgo Channel Guard.

TaylorBeeston (changes requested):
Nice, the actual LC-1964 fix looks solid now. The fallback is scoped to YouTube, the URL opener rejects unsafe schemes, the changeset is in, and the added tests cover the important behavior.

I did catch one thing that looks accidental: the latest commit also checked in generated native registration changes for `@capacitor-community/in-app-review`. That dependency was already present before this branch, these files are unrelated to the YouTube fix, and they are what is tripping Capgo Channel Guard. I think the generated native changes should be reverted from this PR rather than adding a channel bump or compatibility exemption.

Once those are cleaned up, this looks good to me.

goblincore:
Agreed — that was my mistake and it's reverted in c736b88.

I ran `bunx cap sync` locally to build for the simulator, then used `git add -A` for the expand-button commit, which swept in the regenerated native registration for `@capacitor-community/in-app-review`. Confirmed it's unrelated drift and not something this branch introduced:

```
origin/main  apps/learn-card-app/package.json                    in-app-review: present
origin/main  android/capacitor.settings.gradle                   in-app-review: absent
```

So the dependency is on `main` but its generated native registration never was — `cap sync` just regenerated what `main` is already missing.

Your call on reverting rather than bumping the channel is the right one, and for a stronger reason than tidiness: a channel bump would have been actively harmful here. This fix is pure JS/TS, so it should ship as an OTA bundle to existing installs. Bumping the channel would have forced a native release for a change that doesn't need one — and quietly shipped a native dependency change nobody reviewed.

The PR diff is now JS/TS only:

```
.changeset/easy-states-tan.md
apps/learn-card-app/src/components/boost/boostCMS/BoostPreview/{BoostMediaPreview.tsx,BoostMediaPreview.test.tsx}
apps/learn-card-app/src/index.tsx
packages/learn-card-base/src/{helpers/externalUrlOpener.ts,in-app-messages/MessageMedia.tsx}
packages/react-learn-card/src/{components/Lightbox/*,helpers/*,index.ts}
```

Separately: the `in-app-review` native drift on `main` is real and probably worth its own ticket — right now anyone who runs `cap sync` will regenerate those files and hit the same guard.

TaylorBeeston (approved):
Looks good! The YouTube fallback is now correctly scoped so Vimeo, Drive, and Loom keep their existing inline behavior; external URLs are validated; the published packages have an appropriate changeset; and the externally-played Boost preview no longer offers a dead full-screen action. The incidental Capacitor sync changes have also been fully reverted, and all current checks are passing.

- ChatGPT Work

PR #1459, opened 2026-08-05
chore: Switch latest image to weekly interop canary

Custard7:
## What

The interop suite started failing on every PR on 2026-08-05 with `Could not resolve issuer key for 'did:jwk:...'` from the walt.id verifier (`roundtrip.spec.ts` and `multi-credential.spec.ts`).

Root cause: the compose file tracked `waltid/*:latest`, and that day's `latest` push (walt-id/waltid-identity PR #1954, "Crypto2") made the verifier resolve issuer keys strictly by the JWT `kid` header. Our test harness handed walt.id an issuer key with a random `kid` (`kid-x7f3a2`-style) that matches nothing in the `did:jwk` document, which only has one verification method: `<did>#0`. Older builds ignored the mismatch and fell back to the first key; the new one rejects.

Three changes:

1. **Fix our side**: `createIssuerKey()` in `tests/openid4vc-interop-e2e/setup/walt-id-client.ts` now sets `kid` to `<did>#0`. This is the spec-correct value and works with both old and new walt.id resolvers. (The plugin itself was never affected — it already uses `didToVerificationMethod`.)
2. **Pin PR runs**: the walt.id image tag in `compose.yaml` is now `${WALTID_IMAGE_TAG:-0.23.0}`, so PR failures always mean a regression on our side, never upstream drift.
3. **Weekly canary**: the interop workflow gains a Monday 06:00 UTC schedule that runs the same suite with `WALTID_IMAGE_TAG=latest`, restoring the catch-drift-early property without blocking PRs. Canary failures on main already Slack-notify via the existing failure-notification workflow. A `waltid_tag` dispatch input lets you test any candidate tag manually (useful when bumping the pin).

## How to review

- `walt-id-client.ts`: the `kid` change is the only functional edit (plus two prettier-only reflows).
- `compose.yaml`: check the tag interpolation. `docker compose config` resolves to `0.23.0` with no env var, `latest` with `WALTID_IMAGE_TAG=latest`.
- `.github/workflows/openid4vc-interop.yml`: schedule trigger + job-level env. PR-triggered runs get an empty `WALTID_IMAGE_TAG`, which compose treats as unset.
- README documents the pin/canary split and records the incident as walt.id quirk #8.

To verify locally:

```bash
# pinned (what PRs run)
bun --filter @workspace/openid4vc-interop-e2e-tests run test:interop:e2e

# canary (what the weekly run does)
WALTID_IMAGE_TAG=latest bun --filter @workspace/openid4vc-interop-e2e-tests run test:interop:e2e
```

Both pass locally: 23 passed / 3 skipped against `latest`.

### DIFF
--- a/.github/workflows/openid4vc-interop.yml
+++ b/.github/workflows/openid4vc-interop.yml
@@ -68,6 +68,17 @@ jobs:
                   # Surface real-vendor failures rather than silently skipping.
                   CI: 'true'
 
+            - name: Record vendor image digests
+              if: always()
+              working-directory: tests/openid4vc-interop-e2e
+              # Ties a red canary run to the exact upstream build it
+              # tested — `latest` is mutable, digests are not.
+              run: |
+                  docker compose config --images | sort -u | while read -r img; do
+                      digest=$(docker image inspect --format '{{index .RepoDigests 0}}' "$img" 2>/dev/null || echo 'not pulled')
+                      echo "$img -> $digest"
+                  done
+
             - name: Dump docker compose logs on failure
               if: failure()
               working-directory: tests/openid4vc-interop-e2e
--- a/.github/workflows/openid4vc-interop.yml
+++ b/.github/workflows/openid4vc-interop.yml
@@ -8,6 +8,13 @@ name: OpenID4VC Interop Tests
 # The in-process suite (tests/openid4vc-e2e/) continues to run on
 # every PR via the normal `Test Runner` (nx affected) workflow — this
 # file is purely for the heavyweight, real-vendor tier.
+#
+# walt.id image strategy: PR runs use the known-good pin baked into
+# tests/openid4vc-interop-e2e/compose.yaml (WALTID_IMAGE_TAG default),
+# so PRs only fail on regressions on our side. The weekly scheduled
+# run overrides WALTID_IMAGE_TAG=latest as a canary for upstream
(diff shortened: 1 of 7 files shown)


Custard7:
@claude please review

Custard7:
Addressed the review findings in 5d55b9a:

- **A** — added the kid-invariant regression guard to `tests/issue.spec.ts`: `expect(issuerKey.jwk.kid).toBe(`${issuerKey.did}#0`)`, with a comment explaining why the pinned image can't catch it.
- **B** — pinned all three `alpine/socat` sidecars to `1.8.1.3`.
- **C** — fixed the stale `waltid/*:latest` references in `examples/openid4vc-playground/README.md` and `packages/plugins/openid4vc/README.md`; both now point at the `WALTID_IMAGE_TAG` pin.
- **D** — the README canary-repro snippet now includes `docker compose pull` with a note about stale caches.
- **E** — added a "Record vendor image digests" step (`if: always()`) to the workflow, so a red canary is traceable to an immutable digest.
- **F** — fixed "quirk 8 below" → "above". Skipped the two subjective rewordings (quirks preamble, pin phrasing) as pre-existing/cosmetic.

Skipped changeset-bot: test-only workspace package, no version bump warranted.

Custard7:
@claude please re-review for final approval

PR #1470, opened 2026-08-10
feat: [LC-2071] - add accessibility tooling and core-flow improvements

gerardopar:
# Overview

## Summary

This PR introduces automated accessibility checks and improves accessibility across the LearnCard app's core user journeys without intentionally changing the default layout or interaction model.

## Scope covered

-   Sign-in and onboarding
-   Dashboard home
-   Wallet home and category tabs
-   Credential detail
-   Claim and accept credential
-   Share credential
-   Settings and Privacy Settings

## Tooling and CI

-   Added `eslint-plugin-jsx-a11y` with the recommended ruleset for `apps/learn-card-app/src/**`.
-   Added the repository `lint-a11y` command and a dedicated accessibility-lint job in the pull-request lint workflow.
-   Added `@axe-core/playwright` and an Axe journey spec at `apps/learn-card-app/tests/accessibility.spec.ts`.
-   The Axe suite gates on zero serious or critical violations and attaches full-page screenshots at each journey checkpoint.
-   Added the Axe suite to the existing Docker-backed, real-backend Playwright E2E workflow. It runs after the existing Playwright runner; it is not a separate test environment.

## Accessibility improvements

-   Added a skip-to-content link, visible only for keyboard focus.
-   Added shared focus-management and reduced-motion helpers.
-   Added accessible dialog behavior, including dialog semantics, focus trapping, focus restoration, and Escape dismissal where applicable.
-   Replaced non-semantic interactive elements with native buttons or links where touched in scope.
-   Added meaningful accessible names to icon-only controls.
-   Associated form labels, validation messages, and verification-code inputs with their controls.
-   Added keyboard-visible focus rings and verified Enter/Space activation in core flows.
-   Improved tab, toggle, menu, credential-card, claim, share, dashboard, wallet, account, onboarding, and privacy-settings semantics.
-   Updated animations to respect `prefers-reduced-motion`, keeping opacity fades while disabling transform and rotation motion.

## Login design and contrast

-   Centered the sign-in prompt beneath the logo and removed its panel treatment.
-   Matched the social-login `or` divider lines to the text color.
-   Removed the background behind the shared/public-computer control while retaining a keyboard-only focus ring.
-   Set accessible midpoint login colors between each tenant's existing 600 and 700 palette values:
    -   LearnCard / Colorful: `#058760` — white text contrast ratio: 4.525:1.
    -   VetPass: `#0061B8` — white text contrast ratio: 6.173:1.

## Verification completed

[...]

PR #1486, opened 2026-08-19
fix: serve desktop login background as responsive WebP from CDN (3.5MB → 108KB)

goblincore:
# Overview

#### 🎟 Relevant Jira Issues

None — found while testing on staging.

The desktop login background is a **3.5MB PNG**, downloaded on every desktop login page view.

[#1470](https://github.com/learningeconomy/LearnCard/pull/1470) replaced `environments/learncard/assets/branding/desktop-login-bg.png` with a raw **5632×6078** export, taking it from 218KB to 3,707,574 bytes. `prepare-native-config.ts` copies that directory into `public/branding/`, so the full 3.5MB shipped in the web build.

This serves it from the CDN as a responsive WebP set instead.

| | Bytes |
|---|---|
| Before #1470 | 218,243 |
| **On staging today** | **3,707,574** |
| After this PR (1440px @2x) | **108,146** |
| After this PR (1920px @2x) | 139,476 |

#### 🥴 TL; RL:

3.5MB PNG → ~108KB WebP. **34× smaller.**

#### 💡 Feature Breakdown

- **`optimizeUrl` / `generateOptimizedSrcSet`** (`filestack.helpers.ts`) — resize + convert to WebP at an explicit quality.
- **`DesktopLoginBackground`** (new) — emits `srcSet` at 600/900/1200/1600/2000w with `sizes="(min-width: 1024px) 50vw, 100vw"`. Also de-duplicates the identical 11-line `` block that `LoginPage` and `ClaimLoginPage` each carried.
- **`config.json`** — `desktopLoginBgUrl` now points at the CDN handle; the 3.5MB PNG is deleted from the repo.

**The non-obvious part:** the natural Filestack incantation does not work. The standalone `quality=value:N` task **only applies to JPEG output** and is silently ignored for PNG and WebP:

| URL | Result |
|---|---|
| `resize=width:1600/quality=value:75` | 792,112 B — still PNG, quality ignored |
| `resize=width:1600/output=format:webp,quality:75` | **108,146 B** |

Quality has to ride on the `output` task itself. The existing `generateSrcSet` / `resizeAndChangeQuality` helpers emit `quality=value:` and so cannot express this — hence a new pair rather than a change to those. The reason is documented in the JSDoc so the next person doesn't rediscover it.

#### 🛠 Important tradeoffs made

- **The login page now depends on the Filestack CDN.** If it is unreachable, the existing `onError` handler hides the image and the login form renders on the plain green background — it degrades rather than breaking, but the local fallback is gone. Given the login page already cannot function without network access to the auth APIs, this seemed an acceptable trade for 3.4MB.
- **Non-Filestack URLs pass through untouched**, so `vetpass` and script-generated tenants keep rendering their local `/branding` files with no change in behaviour. Only the `learncard` tenant moves to the CDN.
- **Quality 75** was chosen by inspection against 60/85. The asset is a screenshot composite with fine UI text; at 75 it is clean, and it renders into a ~940px slot so downscaling hides the rest.
- `desktop-login-bg-alt.png` (45KB, used by the endorsement modals) is left alone — small enough not to matter.

#### 🔍 Types of Changes

-   [ ] No
-   [x] Yes

Two small ones, both called out rather than hidden:

1. **No unit tests on the new helpers.** `optimizeUrl` and `generateOptimizedSrcSet` are pure and trivially testable; I verified them by hand (see QA below) but did not add specs. Happy to add them to this PR if you'd rather not merge without.
2. **The 3.5MB asset is only removed for the `learncard` tenant.** `snapshot-learncard-defaults.ts` still maps `src/assets/images/desktop-login-bg.png` (the older 218KB / 700×924 version) as a seed default for newly created tenants, and `generate-tenant-assets.ts` still auto-generates a gradient `desktop-login-bg.png`. That path is untouched and works, but new tenants still get a local PNG rather than a CDN asset.

# Testing

1. `bun scripts/prepare-native-config.ts learncard --stage staging` then `bun run serve`
2. Open `http://localhost:3000/login` at a desktop width (≥1024px)
3. The phones artwork should render as before, with the green showing through — WebP alpha is preserved
4. In DevTools → Network, the image should be `image/webp` at ~108KB, **not** a 3.5MB PNG
5. Resize to a mobile width and reload — the background should not render at all (unchanged behaviour)

Verified locally against staging:

- All 5 srcset renditions return `200 image/webp`: 31 / 53 / 76 / 108 / 139 KB
- Browser picks **1600w** at 1440px @2x and **2000w** at 1920px @2x — responsive selection works
- Slot measures 940px against the 960px `sizes` hint, so the hint is well calibrated
- `validate-tenant-configs` — all 7 configs valid
- `prepare-native-config` — `public/branding/` no longer contains the PNG
- `tsc` — zero errors in the new/changed files (the 4 remaining in `LoginPage.tsx` are pre-existing: `approved` / `dob` / `country` / `showCloseButton`)

Chromium / Safari / Firefox at desktop widths. Safari's WebP support is fine on any version we target, but a second pair of eyes on the alpha compositing over the green would be welcome.

Not applicable to native — this renders only when `isDesktop`.

#### 🧪 Code Coverage

No new tests. The existing suites covering the touched surfaces pass:

- `apps/learn-card-app` `src/components/boost-endorsements` — 10 passed (covers `useTenantBrandingAssets` consumers)
- `packages/learn-card-base` `src/config` — 91 passed (covers `brandingHelpers` / `tenantConfigSchema` asset URL resolution)

As noted above, the two new pure helpers deserve specs and I'm glad to add them.

# Documentation

**Internal/AI Docs**

-   [x] **Code comments/JSDoc** — the `quality=value:` vs `output=format:webp,quality:` gotcha is documented on `optimizeUrl`, since it is the kind of thing that silently costs 700KB

#### 💭 Documentation Notes

No user-facing docs needed — this is an asset delivery change with no API surface. The `branding.desktopLoginBgUrl` schema already documented that absolute `https://` URLs are used as-is, so no schema or config docs change either.

# ✅ PR Checklist

-   [ ] Related to a Jira issue — no ticket; found while testing staging
-   [ ] I have considered **product analytics** — N/A, no user-facing interaction added

### ✨ PR Description
Purpose: Optimize desktop login background from 3.5MB PNG to responsive WebP renditions via Filestack CDN, reducing payload to ~108KB with adaptive image delivery.

Main changes:
- Created DesktopLoginBackground component with responsive srcSet generation using configurable widths (600-2000px) and WebP format conversion
- Implemented Filestack URL optimization helpers: `isOptimizableFilestackUrl`, `optimizeUrl`, and `generateOptimizedSrcSet` with quality control
- Migrated desktopLoginBgUrl to Filestack CDN URL and replaced inline img elements in ClaimLoginPage and LoginPage components

---


goblincore:
Both findings reproduce. Fixed in b45883843.

**Issue 1 — handle-less CDN URL.** Confirmed exactly as described. `getUrlParams('https://cdn.filestackcontent.com/')` returns `['']`, and `splice(-1, 0, ...)` on a length-1 array inserts at index 0 rather than before the last element:

```
optimizeUrl('https://cdn.filestackcontent.com/', { width: 1200 })
// was: 'https://cdn.filestackcontent.com/resize=width:1200/output=format:webp,quality:75/'
```

A task-only URL with no handle, as you said. Added `isOptimizableFilestackUrl`, which requires the CDN host *and* a trailing handle, and gated `optimizeUrl` on that instead of `isFilestackUrl`. The `-1` splice is still correct for every URL that reaches it — it's what preserves a leading security policy — so the guard is on reachability rather than the splice itself. There's now a test pinning that too.

**Issue 2 — identical srcset candidates.** Taking the fix, though not quite for the stated reason. Browsers won't "unnecessarily re-request the same resource" — identical URLs collapse to one request, and layout here comes from CSS (`w-full h-full object-contain`), so the descriptors don't affect rendering. The real problem is the one in your last sentence: the contract is misleading. It's an exported helper, and a caller who bypasses the `isOptimizable` guard gets a srcset claiming a 2000w rendition that doesn't exist. Returning `''` says that honestly and lets `src` stand on its own.

Applied your suggestion, extended to cover the issue-1 case as well — a handle-less CDN URL passes `isFilestackUrl` and would otherwise have produced the same identical-candidate list:

```ts
if (!isOptimizableFilestackUrl(url)) return '';
```

Also added the specs this PR flagged as missing (13 tests), covering both edge cases plus task replacement and security-policy preservation. `tsc` clean on the changed files; `learn-card-base` suites 104 passed.

Thanks — issue 1 was a genuine latent bug in a newly exported helper.

PR #1461, opened 2026-08-06
fix: [LC-1962] Centralize safe-area insets for all modals (phase 1)

goblincore:
# Overview

#### 🎟 Relevant Jira Issues

Fixes: [LC-1962](https://welibrary.atlassian.net/browse/LC-1962)

#### 📚 Goal

Native modals kept getting cut off by the status bar / home indicator because every modal had to individually remember safe-area handling. This PR ends the whack-a-mole by making the **modal surface layer own device insets** — modal content never mentions safe area again, and a CI gate makes that rule permanent.

#### 🆚 Why this approach over #1438

#1438 had two structural problems. First, despite being scoped as a modal fix it effectively touched *everything* safe-area-related at once (158 files across modals, per-component classes, footers, scss) with no way to bound the blast radius — every fix risked regressing an unrelated surface. Second, it centralized insets in structural CSS that *guesses* which element paints the surface (`:first-child` selectors, transparent borders, `data-modal-*` escape hatches), so the contract lived in documentation and reviewer vigilance — 12 review rounds and it still showed device regressions. This PR instead:

- **Owns structure instead of guessing it** — one `AppModal` primitive renders the modal DOM for all 7 variants; the six shells wrap it (`useModal` API unchanged, zero call-site churn). Insets are plain padding on the element the primitive itself renders and paints.
- 
What this means is the creation of a new component (learncardbase/components/modals/surfaces/AppModal.tsx) that handles the logic:
Screenshot 2026-08-06 at 7 55 16 PM

Then for all the modal types in useModal they have been swapped to use this base primitive component:
Screenshot 2026-08-06 at 7 55 20 PM

- **Enforces instead of documents** — `check-safe-area.mjs` fails CI on any safe-area primitive outside the surface layer (shrink-only allowlist: 163 → 66 in this PR, → 0 in phase 2).
- **Makes bugs measurable** — `?insets=47,34` simulates device insets in any desktop browser; every geometry claim in this PR was verified against computed styles and on the iOS simulator.
- **Kills the root causes found along the way** — one canonical token pair reconciling iOS `env()` and Android SystemBars; raw `<IonModal>` count 32 → 0 (finishing the old IonModal→useModal migration, deleting verified-dead code); legacy rules that only looked right because two bugs cancelled per-device (the actual source of the historical whack-a-mole).

#### 🗺 Phasing — this PR is Phase 1 (modals only)

Loom of this branch running in iOS Simulator:

[...]

PR #1488, opened 2026-08-20
fix[LC-2114]: Restore ScoutPass BoostCMS network search

goblincore:
### Context and goal

https://welibrary.atlassian.net/browse/LC-2114

In ScoutPass BoostCMS, the fullscreen recipient picker could show an empty contact state but fail to display ScoutPass network results after the user typed a profile ID. The network request still ran and no console error was emitted; the mounted picker simply continued rendering the empty, open-time result snapshot.

This is a separate follow-up to the contact-selection fix in #1487.

### Root cause

The shared modal stack stores an already-created React element. The list-view BoostCMS component owned `search`, `searchResults`, and `isLoading`, then passed their current values into the fullscreen picker when opening it.

Typing still updated state in the hidden parent and triggered `useGetSearchProfiles`, but the stored modal element never received the parent's new props. The visible picker therefore remained on `search=""` and `searchResults=[]`.

The same frozen query also affected the existing troop-leader and network-admin scoped filtering paths.

### Regression history

This behavior regressed on **January 14, 2026**, when #918 replaced ScoutPass `useIonModal` usage with the shared `useModal` stack. Ionic's former overlay refreshed mounted component props; the shared modal stack stores the React element created at modal-open time.

This predates and was not introduced by the later LC-1962 modal safe-area work in #1461.

### What changed

- Give the mounted fullscreen picker ownership of its live search query.
- Run global profile search only for non-empty queries in the mounted regular-user picker.
- Preserve troop-leader and network-admin scoped search without enabling a redundant global query.
- Disable global-search observers in the hidden BoostCMS and troop-ID list shells.
- Pass the derived live results consistently between nested picker views.
- Replace the positional query flag with an `{ enabled }` options object.
- Add regression coverage for zero contacts, empty queries, and scoped searches.
- Add the required ScoutPass and `learn-card-base` patch changesets.

### CI follow-up

The initial CI run exposed two test-only issues:

- `BoostAddressBook.test.tsx` fully mocked `learn-card-base` but omitted the newly used `useGetSearchProfiles` export.
- `boostHelpers.test.ts` dynamically imported a large barrel dependency inside `beforeAll`, exceeding Vitest's 10-second hook timeout under CI load.

The mock now includes the hook. The slow async setup was removed rather than given a larger timeout: `boostHelpers` uses direct module imports and the test uses static imports, reducing its 11-test execution time to about 22 ms in the full suite.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactor / maintenance

## QA steps

1. Run ScoutPass locally against staging services.
2. Open a BoostCMS credential and choose **Send To**.
3. Open the fullscreen contact picker with an account that has no contacts.
4. Type a known ScoutPass profile ID.
5. Confirm the matching ScoutPass network profile appears.
6. For troop-leader or network-admin roles, confirm search remains scoped to the troop or network.

## Verification

- `bun run test` — 3 Bun tests plus 37 Vitest files / 157 tests passed.
- `bun run build` — production build completed.
- `git diff --check` and the repository's staged-file Prettier task — passed.
- Manual localhost verification against staging services — confirmed by Donny.

## Related work

- #1487 fixes the sibling stale-modal issue where clicking a contact's **+** button did not update the mounted picker's selection.
- Both symptoms are tracked in LC-2114.

---

### ✨ PR Description
Purpose: Restore network profile search functionality in BoostCMS recipient picker by adding conditional query enabling based on user role and view mode.

Main changes:
- Added `enabled` option parameter to `useGetSearchProfiles` hook with conditional logic for troop leaders and network admins
- Updated BoostSearch, BoostCMS, BoostAddressBook components to conditionally enable profile search queries based on user restrictions
- Refactored BoostAddressBook to support dual search modes with separate modal and parent search state management

### DIFF
--- a/.changeset/calm-scouts-search.md
+++ b/.changeset/calm-scouts-search.md
@@ -1,0 +1,6 @@
+---
+'scoutpass-app': patch
+'learn-card-base': patch
+---
+
+Restore live ScoutPass network search in the BoostCMS recipient picker.
--- a/apps/scouts/package.json
+++ b/apps/scouts/package.json
@@ -120,7 +120,7 @@
         "docker-build": "VITE_DOCKER_SOURCE=true LCN_URL=http://localhost:4000/trpc CLOUD_URL=http://localhost:4100/trpc API_URL=http://localhost:5100/trpc NODE_OPTIONS=\"--max-old-space-size=8192\" vite build -d",
         "test-playwright": "playwright test",
         "test-ui": "playwright test --ui",
-        "test": "bun test tests/vite-config.test.ts tests/boost-cms-date-picker.test.ts",
+        "test": "bun test tests/vite-config.test.ts tests/boost-cms-date-picker.test.ts && NODE_OPTIONS=--no-experimental-webstorage vitest run --config vitest.config.ts",
         "test-debug": "PLAYWRIGHT_SLOW_MO=1000 playwright test --headed --debug",
         "test-report": "playwright show-report",
         "write-test": "playwright codegen --ignore-https-errors --load-storage=demoState.json https://localhost:3000",
--- a/apps/scouts/src/components/boost/boost-search/BoostSearch.test.tsx
+++ b/apps/scouts/src/components/boost/boost-search/BoostSearch.test.tsx
@@ -1,0 +1,133 @@
+// @vitest-environment happy-dom
+
+import React from 'react';
+import { fireEvent, render, screen } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import BoostSearch from './BoostSearch';
+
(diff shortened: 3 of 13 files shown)


goblincore:
@claude plz review

goblincore:
Addressed the review and CI findings in `4aa0a4679`:

- The mounted picker now owns its live query and consistently passes derived results.
- Empty queries, scoped troop/network searches, and hidden list shells no longer enable irrelevant global searches.
- `useGetSearchProfiles` now uses an `{ enabled }` options object.
- Fixed the missing `useGetSearchProfiles` test mock.
- Removed the timeout-prone dynamic `beforeAll` from `boostHelpers.test.ts` instead of increasing its timeout; its 11 tests execute in about 22 ms in the full suite.

Local verification: 157/157 ScoutPass tests pass and the production build completes. The push has started a fresh CI run.

PR #947, opened 2026-01-21
fix: [LC-1388] Save full name once creating an account

rhen92:
# Overview
This PR removes some old code to properly update the firebase profile on Scouts. 

#### 🎟 Relevant Jira Issues
https://welibrary.atlassian.net/browse/LC-1388

#### 📚 What is the context and goal of this PR?
Fixes an issue where the full name displays on account creation but is missing after logging out and back in.

#### 💡 Feature Breakdown (screenshots & videos encouraged!)
**Before**

https://github.com/user-attachments/assets/8253caae-db8c-483e-955f-06eb55d2654c

**After**

https://github.com/user-attachments/assets/1963ff35-b3b5-45d6-bb7b-9c7d972f6ea1

#### 🔍 Types of Changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Chore (refactor, documentation update, etc)

#### 💳 Does This Create Any New Technical Debt? ( If yes, please describe and [add JIRA TODOs](https://welibrary.atlassian.net/jira/software/projects/WE/boards/2) )
- [x] No
- [ ] Yes

# Testing

#### 🧪 Code Coverage

# Documentation

#### 📝 Documentation Checklist

**User-Facing Docs** (`docs/` → [docs.learncard.com](https://docs.learncard.com))
- [ ] **Tutorial** — New capability that users need to learn (`docs/tutorials/`)
- [ ] **How-To Guide** — New workflow or integration (`docs/how-to-guides/`)
- [ ] **Reference** — New/changed API, config, or SDK method (`docs/sdks/`)
- [ ] **Concept** — New mental model or architecture explanation (`docs/core-concepts/`)
- [ ] **App Flows** — Changes to LearnCard App or ScoutPass user flows (`docs/apps/`)

**Internal/AI Docs**
- [ ] **CLAUDE.md** — New pattern, flow, or context that AI assistants need
- [ ] **Code comments/JSDoc** — Complex logic that needs inline explanation

**Visual Documentation**
- [ ] **Mermaid diagram** — Complex flow, state machine, or architecture

#### 💭 Documentation Notes

# ✅ PR Checklist
- [x] Related to a Jira issue ([create one if not](https://welibrary.atlassian.net/jira/software/projects/WE/boards/2))
- [ ] My code follows **style guidelines** (eslint / prettier)
[...]

PR #1500, opened 2026-08-25
fix: bump python-client deps for Dependabot alerts (urllib3, pytest, filelock)

Custard7:
## Why
Clears 5 open Dependabot alerts on `packages/open-api-lcn-clients/python-client`:

- **HIGH** urllib3 < 2.7.0 — sensitive headers forwarded across origins in proxied redirects (#946)
- **HIGH** urllib3 >= 2.6.0, < 2.7.0 — decompression-bomb safeguard bypass in streaming API (#947)
- **MODERATE** pytest < 9.0.3 — vulnerable tmpdir handling (#945)
- **MODERATE** filelock < 3.20.1 / < 3.20.3 — TOCTOU symlink attacks during lock file creation (#943, #944)

## What
- `urllib3` floor raised `>=2.1.0` → `>=2.7.0` (runtime dep, patch-compatible within the existing `<3.0.0` cap)
- `pytest` floor raised `>=7.2.1` → `>=9.0.3` (dev group only)
- `filelock >= 3.20.3` added explicitly to the dev group / test requirements — it was previously an undeclared transitive (via tox), so this pins the patched floor

Applied consistently to `pyproject.toml`, `requirements.txt`, and `test-requirements.txt`.

## Notes
- This is a generated OpenAPI client; regeneration templates may need the same floors upstream if the client is regenerated (see open PR #979).
- Part of the SOC2 Dependabot cleanup.

---

### ✨ PR Description
Purpose: Bump python-client security dependencies and drop Python 3.9 support to resolve Dependabot vulnerability alerts.

Main changes:
- Upgraded urllib3 to >=2.7.0, pytest to >=9.0.3, filelock to >=3.20.3; bumped minimum Python to 3.10
- Added apply-security-floors.sh script to restore security floors after openapi-generator regeneration
- Updated CI workflows to remove Python 3.9 testing across GitHub Actions, Travis CI, and GitLab CI


Custard7:
Follow-up pushed addressing two gaps:

1. **`requires-python` bumped `>=3.9` → `>=3.10`** — urllib3 2.7.0, pytest 9.0.3, and filelock 3.20.3 all require Python ≥3.10 (verified via PyPI metadata), so the old floor would have made resolution unsolvable on 3.9. Python 3.9 is EOL. `setup.py` (which duplicates the floors) is now updated too.

2. **Regeneration persistence** — the `open-api-generator.yml` workflow does `rm -rf python-client/` and regenerates from stock templates, which would clobber these edits on the next run (there are no in-repo templates, and `.openapi-generator-ignore` doesn't survive the `rm -rf`). Added `packages/open-api-lcn-clients/apply-security-floors.sh` (checked in *outside* the regenerated dir) and a workflow step that re-applies the floors after generation. Script is idempotent.

Custard7:
Addressed review feedback:

1. **tox/pytest incompatibility (valid)** — tox floor raised `>= 3.9.0` → `>= 4.11.0` in `pyproject.toml` and `test-requirements.txt` (tox 3.x is EOL and incompatible with pytest 9; tox ≥4.11 requires Python ≥3.10, consistent with the `requires-python >= 3.10` floor in this PR). `apply-security-floors.sh` re-applies the tox floor after client regeneration too, and remains idempotent (verified by double-run).

2. **urllib3 in requirements.txt (stale finding)** — `requirements.txt` was already updated to `urllib3 >= 2.7.0` in the first commit of this PR; the bot appears to have reviewed only the pyproject hunk. No action needed.

Custard7:
Addressed remaining review feedback:

- **BSD sed `\n` portability (valid)** — the `filelock` insertion in `apply-security-floors.sh` now uses `awk` instead of a sed replacement containing `\n` (a GNU-only extension that BSD/macOS sed silently ignores). Verified on macOS: insertion works on a regenerated pyproject without `filelock`, and the script remains idempotent (double-run produces no diff).
- **tox/pytest incompatibility (already fixed)** — `tox >= 4.11.0` is already in `pyproject.toml`, `test-requirements.txt`, and the floors script.
- **`requirements.txt` urllib3 (stale)** — already pins `urllib3 >= 2.7.0` since the first commit; the bot reviewed only the pyproject hunk.

gerardopar:
@claude review please

Custard7:
Addressed both review items in 61a6977:

1. **Silent no-op on template drift** — `apply-security-floors.sh` now asserts its own post-conditions: after the substitutions, it greps every patched file for the expected final floor (urllib3 ≥2.7.0, pytest ≥9.0.3, filelock ≥3.20.3, tox ≥4.11.0, requires-python ≥3.10) and for the *absence* of Python 3.9 in the client CI configs, exiting nonzero with a "template drift" message on any miss. Reproduced the reviewer's drift scenario (stock floor moved to `>=2.2.0`) — the script now fails loudly with exit 1 instead of printing success.

2. **Python 3.9 in generated client CI configs** — removed the 3.9 entries from `python.yml` (matrix), `.travis.yml`, and `.gitlab-ci.yml`, and added matching substitutions to the floors script so regeneration re-strips them (covered by the post-condition checks above).

Verified: script is idempotent on the committed tree (double-run, zero diff) and correctly re-applies everything against a simulated freshly-regenerated client with stock floors and 3.9 entries restored.

Custard7:
Latest review round, addressed in 166ff50:

1. **Stray `.tmp` on abort (partially valid)** — added an EXIT trap that removes `pyproject.toml.tmp` if the script dies mid-write. Note the review's core premise is incorrect: the redirect writes to a *separate* `.tmp` file, so the original `pyproject.toml` is never truncated or absent — it's untouched until the atomic same-directory `mv`, and `set -euo pipefail` already aborts before the `mv` if awk fails. The only real residue was a leftover `.tmp`, which the trap now cleans up. Verified: idempotent double-run, insertion still works, no `.tmp` left behind.

2. **`requirements.txt` urllib3 (stale, third repetition)** — `packages/open-api-lcn-clients/python-client/requirements.txt` line 1 has pinned `urllib3 >= 2.7.0, < 3.0.0` since the first commit of this PR (a1c66a-era commit), and `test-requirements.txt` is likewise updated. The reviewer keeps evaluating only the pyproject.toml hunk. No action needed.

/gs feedback The requirements.txt urllib3 finding has now been raised three times and is factually wrong each time — the file was updated in this PR's first commit. Please check the committed file contents, not just the diff hunk under review.

gerardopar (approved):
LGTM 🚀

PR #1497, opened 2026-08-24
chore: [LC-1985] - Standardize testing on Vitest

gerardopar:
# Overview

#### 🎟 Relevant Jira Issues

Fixes: [LC-1985](https://welibrary.atlassian.net/browse/LC-1985)

Related subtasks: [LC-2005](https://welibrary.atlassian.net/browse/LC-2005), [LC-2007](https://welibrary.atlassian.net/browse/LC-2007), [LC-2008](https://welibrary.atlassian.net/browse/LC-2008), [LC-2009](https://welibrary.atlassian.net/browse/LC-2009), [LC-2010](https://welibrary.atlassian.net/browse/LC-2010), [LC-2011](https://welibrary.atlassian.net/browse/LC-2011)

#### 📚 What is the context and goal of this PR?

Standardize repository-wide unit and integration testing on Vitest 4.1.10, remove active Jest usage, deduplicate configuration, and make test/build/Nx targets accurately represent the work they perform.

#### 🥴 TL; RL:

- Adds shared Node, happy-dom, and service/integration Vitest presets.
- Migrates active Node, library, browser, React, service, and integration suites.
- Separates CI-safe unit tests from Docker, interoperability, UI, and E2E targets.
- Removes Jest-only dependencies, configs, setup files, and runtime APIs.
- Adds a CI guard that rejects accidental active Jest usage.

#### 💡 Feature Breakdown (screenshots & videos encouraged!)

- Pins `vitest`, `@vitest/coverage-v8`, and `@vitest/ui` to 4.1.10.
- Adds `vitest.shared.ts` with UTC execution, globals, V8 coverage reporters, and project-type presets.
- Adds thin package-local Vitest configs across 47 unit-test projects.
- Uses happy-dom for browser/React packages; the LearnCard app retains a documented jsdom exception for DOMPurify SVG sanitization and computed CSS serialization.
- Adds explicit service integration configs for Brain Service, LCA API, LearnCloud Service, and credential issuance.
- Normalizes `test`, `test:watch`, `test:coverage`, `test:ui`, `test:e2e`, Docker, and interoperability targets.
- Adds `scripts/check-no-jest.mjs` and the `NoJest` CI job.

#### 🛠 Important tradeoffs made:

- Vitest globals are retained to preserve existing tests without unnecessary explicit-import churn.
- No new coverage thresholds are introduced; coverage output is standardized first.
- Service unit targets remain CI-safe, while Docker-backed and external-context suites stay on explicit integration targets.
[...]

PR #1496, opened 2026-08-24
fix: move serverless-offline + @sentry/esbuild-plugin to devDependencies

Custard7:
## Why
`@learncard/learn-cloud-service` (published to npm) declares **serverless-offline** as a runtime dependency. Every downstream consumer (e.g. `app-store-apps` via `@learncard/init`) therefore installs the entire Serverless Framework - including the abandoned `decompress` package, which carries a **critical path-traversal CVE with no patch**. Same story for `@sentry/esbuild-plugin` (a build-time plugin) in three services.

## What
| package | moved to devDependencies |
|---|---|
| @learncard/learn-cloud-service | serverless-offline, @sentry/esbuild-plugin |
| @learncard/network-brain-service | @sentry/esbuild-plugin |
| @learncard/lca-api-service | @sentry/esbuild-plugin |

`@sentry/serverless` stays in dependencies (it wraps handlers at runtime).

## Impact
- Consumers stop inheriting the decompress critical alert + a massively smaller install
- No runtime behavior change: these are only used by `serverless offline` local dev and esbuild bundling, both of which run from the monorepo where devDependencies are installed
- After merge: publish patched versions, then bump `@learncard/*` in app-store-apps to clear its last critical alert

## Related (separate issue recommended)
`@learncard/core`'s dependency `isomorphic-webcrypto` drags in expo/react-native/metro for all consumers (source of the `image-size` high alerts downstream). Replacing it with native WebCrypto is bigger surgery but would slim every consumer install dramatically.

Part of the SOC2 Dependabot cleanup.

### ✨ PR Description
Purpose: Relocate build tooling dependencies (@sentry/esbuild-plugin and serverless-offline) from dependencies to devDependencies across three serverless services to reduce production bundle size.

Main changes:
- Move @sentry/esbuild-plugin to devDependencies in brain-service, lca-api, and learn-cloud-service packages
- Move serverless-offline to devDependencies in learn-cloud-service package for local development only

_Generated by LinearB AI and added by gitStream._
<sub>AI-generated content may contain inaccuracies. Please verify before using.
💡 **Tip:** You can customize your AI Description using **Guidelines** [Learn how](https://docs.gitstream.cm/automation-actions/#describe-changes)</sub>

---
Rovo Dev code review: <strong>Rovo Dev has reviewed this pull request</strong>
Any suggestions or improvements have been posted as pull request comments.

PR #1503, opened 2026-08-25
fix: bump esbuild to ^0.25.0 in ler-rs + render-method plugins (2 Dependabot alerts)

Custard7:
## Why
Clears 2 open MODERATE Dependabot alerts (DEVELOPMENT scope):

- esbuild <= 0.24.2 lets any website send requests to the local dev server and read the response (#493 `plugins/ler-rs`, #746 `plugins/render-method`)

## What
`esbuild ^0.14.38` → `^0.25.0` (devDependency) in both plugins.

## Verification
- `bunx nx build ler-rs-plugin` and `bunx nx build render-method-plugin` both pass with esbuild 0.25 — the build scripts don't use any of the APIs removed between 0.14 and 0.25.

## Overlap note
Dependabot's grouped PR #1344 also bumps esbuild in these packages, but it's entangled with a vitest 1.6 → 4.1 major bump (being handled separately in #1497), so this extracts just the security-relevant esbuild change. #1344 can be closed or rebased once vitest standardization lands.

Part of the SOC2 Dependabot cleanup (follows #1496).

### ✨ PR Description
Purpose: Upgrade esbuild from ^0.14.38 to ^0.25.0 across two plugin packages to resolve Dependabot security alerts.

Main changes:
- Bumped esbuild dependency to ^0.25.0 in ler-rs and render-method plugin packages
- Removed obsolete "aqu" 0.4.3 dependency from both package.json files

---

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1291,7 +1291,7 @@
         "@types/jest": "^29.2.2",
         "@types/node": "^17.0.31",
         "aqu": "0.4.3",
-        "esbuild": "^0.14.38",
+        "esbuild": "^0.25.0",
         "esbuild-jest": "^0.5.0",
         "esbuild-plugin-copy": "^1.3.0",
         "jest": "^29.3.0",
@@ -1381,7 +1381,7 @@
         "@types/jest": "^29.2.2",
         "@types/node": "^17.0.31",
         "aqu": "0.4.3",
-        "esbuild": "^0.14.38",
+        "esbuild": "^0.25.0",
         "esbuild-jest": "^0.5.0",
         "esbuild-plugin-copy": "^1.3.0",
         "jest": "^29.3.0",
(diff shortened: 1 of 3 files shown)


TaylorBeeston (changes requested):
Found one thing we need to fix before merging: the direct bump works, but aqu@0.4.3 still pulls aqu/esbuild@0.15.18, so the vulnerable version remains in both affected package graphs. I left the exact path inline; the builds/tests pass otherwise.

— OMP

TaylorBeeston (changes requested):
One more concrete detail on the finding: the vulnerable path is the aqu devDependency in both package.json files, and bun.lock resolves aqu/esbuild to 0.15.18. Since the package build scripts invoke esbuild directly and don’t use aqu, removing that unused dependency looks like the cleanest fix.

— OMP

Custard7:
@TaylorBeeston Addressed — `aqu@0.4.3` removed from both `ler-rs` and `render-method`. Confirmed unused: both build via `scripts/build.mjs` invoking esbuild directly, and no script/config references aqu. `bun.lock` no longer lists aqu in either package's section, so `aqu/esbuild@0.15.18` is out of both affected graphs (aqu itself remains in the lockfile only for the ~25 *other* plugins that still build with it — out of scope for these two alerts).

Verification: `nx build` passes for both plugins (fresh, cache skipped) and `jest` runs clean. One honest note on the esbuild-jest concern from the bot review: both packages have zero test files (`--passWithNoTests`), so the esbuild-jest transform is never exercised at runtime — nothing to break today, but if tests are ever added, esbuild-jest 0.5.0's compatibility with esbuild 0.25 would need a real check then.

TaylorBeeston (approved):
LGTM! The follow-up removes the unused aqu dependency from both affected packages, and the lockfile no longer puts aqu/esbuild in either graph. I reran both Nx builds and test targets fresh; everything passes. The e2e job was canceled by the shared runner queue.

— OMP

PR #1505, opened 2026-08-25
chore: [LC-1984] standardize runtime and tooling without Tailwind 4

TaylorBeeston:
# Overview

#### 🎟 Relevant Jira Issues

Completes: [LC-1998](https://welibrary.atlassian.net/browse/LC-1998), [LC-2001](https://welibrary.atlassian.net/browse/LC-2001), [LC-2002](https://welibrary.atlassian.net/browse/LC-2002), [LC-2003](https://welibrary.atlassian.net/browse/LC-2003), and [LC-2004](https://welibrary.atlassian.net/browse/LC-2004)

Part of: [LC-1984](https://welibrary.atlassian.net/browse/LC-1984)

Not included: [LC-2006](https://welibrary.atlassian.net/browse/LC-2006)

The monorepo had several competing runtime, TypeScript, ESLint, environment-validation, and Nx conventions. That made local behavior differ from CI and allowed package-level tooling to drift.

This PR establishes one supported path for Node 24.12.0, Bun 1.3.14, TypeScript 5.9.3, ESLint 10 flat config, layered TypeScript presets, typed environment contracts, and package-owned scripts composed through Nx.

I removed the Tailwind 4.3 migration from this PR after the populated dashboard exposed app-wide visual drift that the build and login smoke checks did not catch. LearnCard App, ScoutPass, and `@learncard/react` are back on their existing Tailwind 3 configs, source classes, and generated CSS pipeline. LC-2006 is back in To Do so that migration can get focused visual-regression coverage instead of increasing the risk of the rest of this tooling work.

#### 🥴 TL; RL:

- Pins Node 24.12.0 and Bun 1.3.14 across local setup, CI, Docker, and package metadata.
- Makes TypeScript 5.9.3 the actual installed compiler.
- Replaces legacy ESLint config with ESLint 10 flat config and an explicit legacy-error baseline; lint-staged blocks new errors in touched files.
- Adds shared base, browser, Node, and test TypeScript presets and migrates their consumers.
- Adds typed, fail-fast environment contracts for LearnCard App, ScoutPass, Brain Service, LCA API, and LearnCloud Service.
- Moves target ownership to package scripts while keeping Nx as the dependency-aware orchestrator.
- Adds `bun run verify:lc-1984` as a streaming, fail-fast regression gate for runtime pins, config ownership, actual package commands, canonical lint/typecheck, and dependency-aware builds.
- Deliberately keeps the three active Tailwind surfaces on Tailwind 3; there is no intended UI change in this PR.

#### 💡 Feature Breakdown

**Runtime and package management**

- Synchronizes `.nvmrc`, Mise, root/package `engines`, `packageManager`, GitHub Actions, Docker images/installers, and Netlify configuration.
- Removes stale package-local CI workflows that were still testing Node 10/12/14.

**TypeScript and ESLint**

- Introduces layered root presets for shared, browser, Node, and test projects.
- Upgrades the installed compiler to TypeScript 5.9.3 and fixes the WebCrypto `ArrayBuffer`/`SharedArrayBuffer` typing exposed by that upgrade.
- Replaces `.eslintrc` with flat config and makes the legacy UI baseline explicit rather than silently skipping those sources.

**Typed configuration**

- Adds schema-backed environment parsing and startup validation to five runtime surfaces.
- Removes direct untyped environment reads from the migrated paths and documents the environment contract.

**Nx and package scripts**

- Gives packages direct `build`, `lint`, `test`, `test:watch`, and `typecheck` ownership where applicable.
- Uses Nx to compose those package commands and verifies that every referenced script exists.
- Makes mocked E2E explicitly build `@learncard/helpers` before Vite starts, so clean checkouts cannot rely on stale `dist` output.

**Tailwind scope**

- Restores the known-good Tailwind 3 JavaScript configs, PostCSS integration, source utilities, browser targets, and shared React CSS build.
- Removes the Tailwind 4 CSS-first files, Vite/PostCSS adapters, source codemod, browser-floor document, changeset, and verifier claims.
- Keeps LC-2006 separate until populated dashboard/shared-component visual contracts are covered.

**Clean-checkout hardening**

- Restores JavaScript emission for `@learncard/email-templates` while retaining declarations.
- Pins `@inlang/paraglide-js` to `2.20.2` in both apps; the lockfile resolves `@inlang/sdk` 2.10.2 and Lix 0.4.10 to avoid the initial-compilation stall in newer Lix releases.
- Streams each verifier phase directly and exits on the first failed command instead of buffering output and continuing.

#### 🛠 Important tradeoffs made

- Node 24.12.0 and Bun 1.3.14 are intentional runtime floors.
- LearnCard App and ScoutPass still contain known legacy lint debt. It is now recorded and bounded instead of hidden; expanding this work into a legacy-UI rewrite would make the migration unreviewable.
- The canonical typecheck remains scoped to the existing typed projects. Test projects consume the shared test preset, but this PR does not broaden into unrelated historical end-to-end diagnostics.
- Deferring LC-2006 means the LC-1984 parent remains in progress, but it avoids shipping a systemic styling regression with otherwise independent tooling work.
- Paraglide remains on 2.20.2 until the newer Lix startup behavior is fixed and validated against LearnCard's translation volume.

#### 🔍 Types of Changes

- [ ] Bug fix
- [ ] New feature
- [x] Breaking change
- [x] Chore

The breaking scope is limited to runtime and tooling contracts; supported application behavior and public APIs are unchanged.

- [x] No
- [ ] Yes

The legacy lint debt is pre-existing and now measurable. Tailwind 4 remains tracked by LC-2006 rather than being represented as complete here.

# Testing

From the repository root:

1. Confirm `node --version` is `v24.12.0` and `bun --version` is `1.3.14`.
2. Run `bun install --frozen-lockfile`.
3. Run `bun run verify:lc-1984`.
4. Run `bunx nx run learn-card-app:test-mock --skip-nx-cache`; all five mocked E2E cases should pass after Nx builds helpers.
5. Run `bun run lc dev learncard local app`; Vite should become ready without stalling during Paraglide compilation.
6. Run `bunx lint-staged` with the PR files staged.
7. Start LearnCard App against the local backend stack, sign in through `/hidden/seed`, and compare `/dashboard` with `main` at 390px. The header, quick actions, Activity typography/card density, and bottom nav should retain the existing Tailwind 3 rendering.

A populated mobile dashboard account is the highest-value review surface because that is where the Tailwind 4 regression became obvious. ScoutPass and shared credential-card states are also useful checks even though their CSS pipeline now matches `main`.

#### 🧪 Code Coverage

No coverage percentage target applies to this tooling migration. Post-rebase verification passed canonical lint/typecheck and all dependency-aware builds; mocked E2E passed 5/5; and the local app-only flow compiled Paraglide, reached Vite in 9.8 seconds, and returned HTTP 200. Staged-file linting and the real local LearnCard dashboard at 390px were also checked. The Activity/dashboard source and Tailwind integration match the known-good `main` implementation; the empty dashboard Activity state renders with compact 11–14px typography and no clipping.

# Documentation

- [ ] User-facing GitBook docs
- [ ] New public API reference
- [ ] App-flow documentation
- [x] Internal environment/tooling documentation and executable verifier

This is internal tooling work with no new user workflow or public SDK API. `environment-variables.md`, app environment docs, inline compatibility notes, and the executable verifier cover the changed contracts. The Tailwind-specific browser-floor document was removed with LC-2006.

# ✅ PR Checklist

- [x] Related Jira issues are linked and accurately scoped.
- [x] Code follows ESLint/Prettier conventions.
- [x] Common end-to-end surfaces were manually checked.
- [x] The diff was reviewed after removing LC-2006.
- [x] Relevant build and verification commands pass locally.
- [x] No product analytics change is needed because this PR does not add a user interaction.

- [ ] Code is backwards compatible
- [x] There is no “Do Not Merge” label.
- [x] Security implications were considered.
- [x] No new public endpoint is exposed.

The backwards-compatibility box remains unchecked because older runtime versions are outside the new support contract.

— OMP

[LC-1998]: https://welibrary.atlassian.net/browse/LC-1998?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LC-2001]: https://welibrary.atlassian.net/browse/LC-2001?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LC-2002]: https://welibrary.atlassian.net/browse/LC-2002?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LC-2003]: https://welibrary.atlassian.net/browse/LC-2003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LC-2004]: https://welibrary.atlassian.net/browse/LC-2004?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LC-1984]: https://welibrary.atlassian.net/browse/LC-1984?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LC-2006]: https://welibrary.atlassian.net/browse/LC-2006?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### ✨ PR Description
Purpose: Migrate from build-time environment variables to standardized runtime configuration objects without Tailwind 4 dependency.

Main changes:
- Created `buildEnvironment.ts` and `environment.ts` modules providing typed, frozen environment configuration accessible throughout apps
- Replaced `import.meta.env.*` and `process.env.*` calls with `environment` object references for consistent centralized config
- Standardized Storybook and test build environment definitions using validated schema patterns


goblincore (changes requested):
codex agent: Requesting changes because the current branch is not verifiable from a clean checkout:

- @learncard/email-templates now inherits emitDeclarationOnly: true, but still exports dist/index.js. This causes five CI jobs to fail; stale local dist output can mask it.
- Mock E2E fails because @learncard/helpers is not built before Vite loads the new environment module.
- Canonical lint fails with React Hooks errors outside the committed baseline.
- verify:lc-1984 buffers all progress, continues after failures, and appears hung.


Please fix the clean-build path, make the verifier observable, rebase, and rerun the complete check suite.
Screenshot 2026-08-27 at 12 39 02 PM

goblincore:
@TaylorBeeston — additional blocking local-development regression found while following the testing instructions on commit `e7c212d`.

### Reproduction

After a successful `bun install`, start LearnCard App in local/app-only mode (`bun run lc dev learncard app`). Startup reaches:

```text
✓ i18n guard: no duplicate paraglide/messages.js imports
```

It then fails to become ready. The live Vite process remained at roughly 98–100% of one CPU core for more than 9 minutes. No Paraglide output timestamps changed and no `project.inlang/cache/paraglide-js` cache was written. A macOS stack sample isolated the work to the native Lix thread executing read SQL/DataFusion hot-state materialization and repeatedly sorting `resolve_visible_batch` rows.

The regenerated lockfile upgrades this stack even though the manifests still specified `^2.18.1`:

- `@inlang/paraglide-js`: `2.20.2` → `2.24.1`
- `@inlang/sdk`: `2.10.2` → `3.0.2`
- `@lix-js/sdk`: `0.4.10` → `0.12.3`

LearnCard currently has about 27,245 scalar translations across four locale files, making the blocking full compile especially expensive. This matches the upstream report that the Vite plugin recompiles all messages on the server-readiness critical path: https://github.com/opral/paraglide-js/issues/741

### Controlled downgrade result

I pinned `@inlang/paraglide-js` to exact `2.20.2` in both app manifests and ran `bun install`, restoring Inlang SDK `2.10.2` and Lix SDK `0.4.10`. With no other source change, the same app-only startup:

- completed Paraglide compilation in seconds;
- reported `VITE v6.4.2 ready in 22297 ms`;
- returned HTTP 200 from `http://localhost:3000/`;
- dropped to 0% Vite CPU after first dependency optimization.

Please restore/pin the previous Paraglide resolution, or otherwise address and validate the newer Lix performance behavior before merge. A time-bounded smoke check for the documented app-only `bun lc` flow would also prevent this from recurring. The upstream cache fix only helps after an initial compilation completes: https://github.com/opral/paraglide-js/pull/749

TaylorBeeston:
Thanks for digging into this! I fixed the clean-checkout failures and the Paraglide regression, then rebased onto `6618bd588`. Email templates now emit JavaScript, mock E2E builds helpers first, latest-main lint diagnostics are baselined, and `verify:lc-1984` streams and fails fast. Paraglide is pinned to `2.20.2` in both apps (`@inlang/sdk` 2.10.2 / Lix 0.4.10 in the lockfile).

Post-rebase, the verifier passed; mocked E2E passed 5/5; and the local app-only flow reached Vite in 9.8s and returned HTTP 200. Could you take another look?

— OMP

goblincore:
@TaylorBeeston — I rereviewed the current head (`f959864`) from a clean isolated checkout. The earlier requested fixes look addressed: email-template JS emission, the mock-E2E helper build, streamed/fail-fast verifier output, and the exact Paraglide pin.

I can still reproduce three clean-checkout blockers:

1. `node scripts/check-no-jest.mjs` exits 1 because the root `package.json` still declares `@types/jest`, `esbuild-jest`, and `ts-jest`. This matches the red **Reject active Jest usage** job.
2. From `packages/plugins/didkit-plugin-node`, `bun run build` exits 0 but emits only `.d.ts` files. `dist/index.js` is missing even though `package.json` exports it. The package inherits `emitDeclarationOnly: true` from `tsconfig.library.json` without overriding it. This is the underlying clean-build/package-resolution problem affecting the downstream build jobs.
3. `tests/smoketests` now exposes `playwright test` as the generic Nx `test` target, so `.github/workflows/test.yml` picks it up through `nx affected --target=test`. That workflow has no `playwright install` step. With an empty `PLAYWRIGHT_BROWSERS_PATH`, `bun run smoke:browser` fails all 7 browser checks with “Executable does not exist,” matching CI.

The PR currently also reports merge conflicts with `main`, and 11 checks remain red. Four of those are statuses from one Netlify docs deploy; I could not inspect those logs, so I am not assigning a cause to them here.

Screenshot 2026-09-02 at 12 56 07 PM

TaylorBeeston:
Thanks! I fixed the three clean-checkout blockers and rebased onto `092b1d36e`: removed the active Jest dependencies, restored DIDKit Node’s JavaScript emission, and removed smoketests from the generic `test` target. I also fixed the DEV/CORS feedback and made Docker queue delivery share Lambda’s durable-storage/acknowledgement path.

The clean DIDKit build, focused app tests, 62 queue/credential integration tests, and full verifier all pass now. Could you take another look? =)

— OMP

TaylorBeeston:
Found and fixed the separate analytics bootstrap-race note too. Both environment detection and shared event context now tolerate unresolved tenant config, and Vitest defines the injected build environment/version constants. The analytics hygiene, build-environment, and CORS tests pass (41 tests total). Pushed `e6e95f2c0`. Thanks for flagging this!

— OMP

TaylorBeeston:
One more analytics edge case is fixed in `dbb838bf9`: native staging builds now key off `MODE` instead of Vite’s build-wide `PROD` flag, so they stay tagged as staging. The environment mock is mutable and typed, with a native-staging regression test.

I reran the complete `verify:lc-1984` workflow on these changes, then reran the focused 42-test set after the commit hook. Everything passes.

— OMP

goblincore:
@TaylorBeeston — one more blocking CI issue confirmed on the current head (`dbb838b`): the E2E suite is now being launched by the normal unit-test job even though this workflow already has a separate `e2e-tests` job.

This PR adds both:

- `"test": "vitest run"` to `tests/e2e/package.json`
- a generic `test` target to `tests/e2e/project.json`

That makes the normal workflow command, `bunx nx affected --target=test --base=origin/main --head=HEAD`, include `nx run e2e:test`.

In the failed Test job, that target prints:

```text
Skipping docker compose up (set E2E_MANAGE_DOCKER=true to enable)
Waiting for health check...
```

The unit-test runner never starts the required backend, so E2E waits until the outer 10-minute timeout. Because the workflow retries three times, this unnecessarily stretches the normal Test job to roughly 31 minutes while the dedicated EC2-backed `e2e-tests` action runs separately anyway.

Please remove the generic `test` script/target from `tests/e2e` and keep `test:e2e` / `test:run` as explicitly invoked targets. That restores the separation present on `main` and matches the cleanup already made for `tests/smoketests`.

The LearnCard Base logger assertion is a separate real unit-test failure and still needs fixing, but it should not be coupled to this accidental E2E invocation.

goblincore:
@TaylorBeeston — correction: the four Netlify failures on #1505 all come from one `reference/` docs build failure, not four separate problems.

The PR lockfile resolves Webpack `5.110.1`, while the docs use Docusaurus `2.1.0` and WebpackBar `5.0.2`. That combination fails Webpack's `ProgressPlugin` validation because WebpackBar passes unsupported `name`, `color`, `reporters`, and `reporter` options. This matches the known incompatibility documented in https://github.com/facebook/docusaurus/issues/11974.

Release PR #1524 does **not** reproduce the failure: its Netlify deploy was canceled and recorded as non-failing, and it changes neither `bun.lock` nor `reference/`. So this should be scoped to #1505's dependency changes, not current `main`.

Could you pin Webpack to a compatible version or upgrade the docs dependencies so the Netlify preview can build?

goblincore:
@TaylorBeeston — the `chromatic-learn-card-app` failure is caused by the new build-environment global not being defined in Storybook.

The Storybook build and upload both succeed, but Chromatic fails while loading the published Storybook to extract stories:

```text
Error: __APP_BUILD_ENV__ is not defined
```

`src/config/environment.ts` now evaluates `Object.freeze(__APP_BUILD_ENV__)`. The regular app defines that constant in `vite.config.mts`, and Vitest provides a test value, but `.storybook/main.ts` has its own Vite `define` block and does not provide `__APP_BUILD_ENV__`. The unresolved identifier therefore survives the Storybook build and throws in the browser.

Could you add a Storybook-safe `__APP_BUILD_ENV__` value to the Storybook Vite config, or share the definition logic across the app, tests, and Storybook?

goblincore:
@TaylorBeeston — the failing **Run accessibility lint** check is not currently reporting accessibility violations.

#1505 changes the root `lint` script to run:

```text
bun scripts/lint-workspace.ts && node scripts/lint-a11y.cjs
```

The new workspace lint runs first with `react-hooks` `recommended-latest` enabled and exits before `lint-a11y.cjs` can start. CI reports 59 React Hooks errors across 21 files, but none of those files were changed by this PR, all 21 already have entries in `scripts/eslint-baseline.json`, and the violation counts largely match. The committed baseline fingerprints therefore do not match the output from the clean CI install—likely because they were generated with a different/stale lint environment.

Could you regenerate the baseline from a clean checkout using the pinned Node/Bun versions and `bun install --frozen-lockfile`, then rerun this check?

Separately, the existing workflow check name is now misleading because it invokes the expanded root `lint` script. General ESLint and the accessibility-specific lint should ideally be separate steps/checks, although that naming/workflow cleanup is not necessarily part of this PR.

TaylorBeeston:
Addressed the newest CI and Chromatic feedback in `8cfaa1f7b`:

- removed the E2E package’s generic `test` script/Nx target, so the normal unit-test job can’t launch browser E2E
- fixed the logger production assertion to use the injected `IS_PRODUCTION` constant
- guarded analytics provider loading before tenant bootstrap and corrected native/unknown-host stage classification from tenant config
- fixed Vitest’s `DEV`/`PROD` invariant
- defined the app build/version constants in Storybook
- pinned Webpack `5.105.4` at the workspace override so Docusaurus and its nested dependencies resolve the compatible version
- regenerated the ESLint baseline from the frozen install

Verification:

- `bun install --frozen-lockfile`
- `bun run verify:lc-1984`
- LearnCard App focused tests: 44 passed
- LearnCard Base logger tests: 58 passed
- Docusaurus reference build passed
- final Storybook production build passed; generated bundles contain no unresolved `__APP_BUILD_ENV__` or `__APP_VERSION__` references

— OMP

TaylorBeeston:
One more direct check against the normal Test job path:

`bunx nx affected --target=test --base=origin/main --head=HEAD`

Passed across 49 affected projects: 179 test files passed, 1 skipped; 1,912 tests passed, 4 skipped. The resolved project list does not include `e2e`, and LearnCard Base completed through the same affected-test run.

— OMP

goblincore:
@TaylorBeeston I kept your latest fixes from `8cfaa1f7b` and pushed the remaining CI cleanup on top:

- Split workspace ESLint from jsx-a11y so the Accessibility check is accurately scoped.
- Normalized checkout-root paths before hashing ESLint baseline diagnostics, then regenerated the baseline.
- Ignored generated Playwright report/result directories.
- Fixed the two remaining Prettier failures.
- Raised the dashboard quick-action caption color from `grayscale-500` to `grayscale-600` to resolve the Firefox Axe contrast failures.

Local verification:

- `bun run lint:workspace`: passed (11,349 existing legacy errors baselined)
- `bun run lint:a11y`: passed (0 errors; 319 existing warnings)
- CI-style changed-file Prettier check: passed
- Workflow contract test: passed
- `bun run test:a11y`: 2 passed, 1 conditionally skipped; the two journeys that previously failed contrast now pass

Pushed as `f1e6ede53` and `5eb50a97e`.

goblincore (approved):
@TaylorBeeston looks like all checks green now, builds fine on my end! LGTM

PR #1510, opened 2026-08-27
[LC-2132] Return OAuth presentations in URL fragments

TaylorBeeston:
## Why

LearnCard currently returns signed DID Auth presentations in the consumer’s callback query string:

```text
https://consumer.example/callback?session=S&did=D&vp=<signed-presentation>
```

The query is part of the HTTP request URL, so the presentation can end up in proxy, CDN, server, access, and request-monitoring logs. HTTPS protects it in transit, but not from the systems processing the request, and a signed JWT is readable rather than encrypted. Multi-kilobyte presentations can also hit browser or proxy request-line limits.

## What “fragment” means here

The fragment is the part after `#`, sometimes called the hash. It is not a cryptographic hash:

```text
https://consumer.example/callback?session=S#did=D&vp=<signed-presentation>
```

The browser requests only `/callback?session=S`; it does not send the fragment in the HTTP request or `Referer`. The fragment is still visible to the browser, may briefly appear in the address bar, and may be retained in browser-side state. This is not encryption. The benefit is specifically keeping the presentation out of HTTP request URLs and the infrastructure logs that routinely record them.

The consumer is responsible for reading the fragment and moving the presentation into a request body. The paired AI Passport callback does that with a no-store, CSP-protected same-origin POST.

## What changed

- adds an explicit `response_mode=fragment` opt-in for consent return URLs
- keeps the existing query delivery as the default when the opt-in is absent
- centralizes query/fragment URL construction in `createConsentFlowReturnUrl`
- preserves the mode while users move through the different consent paths
- updates each path that can finish consent and return a DID/presentation

The number of touched components is intentional: LearnCard currently completes consent from several paths depending on contract and user state. If one path drops the response mode, fragment delivery would work for some users and silently fall back to query delivery for others.

## Compatibility and rollout

This pairs with [AI Passport #68](https://github.com/WeLibraryOS/AI-Passport/pull/68).

Fragment delivery is opt-in. Existing consumers that do not send `response_mode=fragment` continue receiving `did` and `vp` in the query exactly as before.

Deploy this producer support first, then AI Passport. Producer-first preserves every existing consumer while making fragment mode available. Deploying the AI Passport consumer first would break MCP OAuth login because LearnCard would still send the presentation in the query while the new callback expects the fragment-to-POST flow.

## Tradeoffs

- the browser and destination page can still read the fragment; this is not end-to-end encryption
- the fragment may be visible in the address bar or retained in browser-side state
- fragment consumers need JavaScript or equivalent client-side handling to move the response into a request body
- the eventual POST body remains sensitive and must not be logged
- every consent completion path must preserve the opt-in consistently

## QA

- without `response_mode`, confirm existing query delivery is unchanged
- with `response_mode=fragment`, confirm `did` and `vp` are after `#` and absent from the callback HTTP request URL
- exercise already-consented, new-consent/sync, full-screen, and game return paths
- confirm existing fragment parameters are preserved
- exercise a multi-kilobyte presentation with the paired AI Passport callback

## Verification

The return-URL helper tests cover default query behavior, fragment behavior, and existing fragment preservation. Targeted LearnCard build and consent callback checks pass.

— OMP

### ✨ PR Description
Purpose: Implement optional delivery of OAuth presentations via URL fragments to improve security by keeping credentials out of HTTP request logs.

Main changes:
- Created `createConsentFlowReturnUrl` utility to handle both query and fragment response modes
- Updated consent flow components to parse `response_mode` and use the new URL utility
- Added unit tests for URL generation and updated `ExternalConsentFlowDoor` integration tests

### DIFF
--- a/.changeset/calm-fragments-return.md
+++ b/.changeset/calm-fragments-return.md
@@ -1,0 +1,5 @@
+---
+'learn-card-app': patch
+---
+
+Add opt-in fragment delivery for consent callback presentations while preserving query delivery by default.
--- a/apps/learn-card-app/src/pages/consentFlow/ConsentFlowSyncCard.tsx
+++ b/apps/learn-card-app/src/pages/consentFlow/ConsentFlowSyncCard.tsx
@@ -26,6 +26,7 @@ import ConsentFlowEditAccess from '../launchPad/ConsentFlowEditAccess';
 
 import useTheme from '../../theme/hooks/useTheme';
 import * as m from '../../paraglide/messages.js';
+import { createConsentFlowReturnUrl } from './createConsentFlowReturnUrl';
 
 export type ConsentFlowWriteAccessType = {
     [BoostCategoryOptionsEnum.socialBadge]: boolean;
@@ -97,7 +98,7 @@ const ConsentFlowSyncCard: React.FC<ConsentFlowSyncCardProps> = ({
     const location = useLocation();
     const brandingConfig = useBrandingConfig();
 
-    const { returnTo } = queryString.parse(location.search);
+    const { returnTo, response_mode: responseMode } = queryString.parse(location.search);
 
     // state for handling - data share duration
     const [shareDuration, setShareDuration] = useState<{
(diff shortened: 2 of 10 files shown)


TaylorBeeston:
Closing this for now, along with WeLibraryOS/AI-Passport#68.

Keeping the DID-Auth presentation out of request URLs has a narrow exposure-reduction benefit, but it isn't necessary for the LC-2118 privacy gateway work. The existing challenge, browser binding, audience, and single-use checks already constrain replay, and this change adds cross-repo rollout coordination and another callback path to maintain.

That tradeoff isn't worth prioritizing right now. We can revisit it if production logging exposure or callback-size failures give us a concrete reason. The remaining consent/privacy work is being detached from this change.

— ChatGPT Work

PR #1511, opened 2026-08-27
[LC-2125] Return live AI consent metadata

TaylorBeeston:
## What changed
- returns contract URI, status, terms URI, grant/expiry timestamps, and terms needed for server-side AI consent enforcement
- limits results to live records and omits withdrawn or expired consent
- extends the shared consent response contract without exposing unrelated records

## Stack and rollout
Depends on #1510 and pairs with WeLibraryOS/AI-Passport#71.

Deploy this producer before the AI Passport active-consent consumer.

## Verification
- LearnCard types and brain-service builds pass
- withdrawn-consent omission is covered by the focused integration test; local execution still requires the service SEED fixture

— OMP

### DIFF
--- a/.changeset/live-consent-metadata.md
+++ b/.changeset/live-consent-metadata.md
@@ -1,0 +1,6 @@
+---
+'@learncard/types': patch
+'@learncard/network-brain-service': patch
+---
+
+Return live consent metadata required for server-side active-consent enforcement.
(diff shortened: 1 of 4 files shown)


TaylorBeeston:
Superseded by #1536 as part of the PR-stack cleanup.

This change now targets main independently, without the unrelated stack dependencies.

GitHub would not allow changing bases inside the native stack, so the existing changes were carried into a replacement PR. This is organization work, not a new implementation or an approval of the outstanding review concerns. The original branch and discussion are retained here for reference.

— ChatGPT Work

PR #1521, opened 2026-08-27
Docs migration part 2: new quickstart + surfacing orphaned pages

Custard7:
Second PR in the docs stack. No URLs change in this one.

The big change is the quickstart. The old "Your First Integration" was 240 lines teaching claimable boosts. The new one is 92 lines and does the thing most integrators actually want: send a credential to an email address with `sendCredentialViaInbox` (about 15 lines of code), with a curl variant using an API token for people who don't want the SDK. Also adds a "What do you want to build?" page with the five integration paths, a glossary (we say "SSS" 69 times across the docs and never once spell out what it means), and a "Which SDK do I need?" table since Partner Connect and Embed SDK sound interchangeable but point in opposite directions.

Second commit fixes a dumb problem we had: 48 doc files existed on disk but weren't in SUMMARY.md, so they were unreachable — including all 16 plugin docs and the AI API section. People were finding some of these through Google but not through our own nav. This puts 34 of them back in the nav (which fixes the 29 dead links the checker flags on the foundation branch — the allowlist shrinks 46→17 here), merges the three overlapping network prose pages into one, and expands the Storage API landing page. It also moves a few pages that read like internal engineering notes out of the public docs: tenant-branded emails, the resume builder LER-RS mapping, and credential activity tracking, all with redirects.

I also looked at merging consent-transactions and verifiable-data-in-consentflow like the audit suggested, but after actually reading them they cover different things, so they stay.

Before merging:
- [ ] anyone know of a partner actually implementing against the tenant-email or LER-RS pages? If so shout — they're being made internal in this PR
- [ ] actually run the quickstart against staging with a fresh seed — email should arrive, claim link should work
- [ ] run the curl variant too
- [ ] if the real flow turns out to need more than ~15 lines or hits some snag (signing authority, profile weirdness), that's a product bug — file it and fix the doc to match reality before merging
- [ ] skim network-architecture.md and make sure the merge of the three pages reads ok


Custard7:
Superseded by #1528, which collapses the three-PR stack into a single PR from the same branch history (all commits preserved). Reason: GitBook only generates previews for PRs targeting `main`, and for a docs restructure the rendered preview is the review. Review there; this PR's discussion is kept for history.

PR #1529, opened 2026-09-03
SSS hardening

Custard7:
# Overview

#### 🎟 Relevant Jira Issues

Related to: SSS prod cutover (Web3Auth → SSS key derivation)

We haven't switched production from `web3auth` to `sss` (Shamir Secret Sharing — a user's key is split into a device share and a server share) yet because the recovery story wasn't safe enough for students. Two problems in particular:

1. **Learners could get permanently locked out.** Recovery setup was skippable, and the fallback (a share emailed to the signup address) fails exactly when a student loses their school account — which also kills their login. Every recovery path used the original OAuth token, so even a verified backup email didn't help.
2. **We were briefly custodial.** The email backup route received the plaintext share in the same process that holds the server share and its encryption key. A compromised API could reconstruct any user's private key.

This PR closes the blockers we identified for flipping SSS on in prod. It does not flip it — prod stays on `auth.keyDerivation: "web3auth"` until staging soak is done.

#### 🥴 TL; RL:

Recovery is now mandatory and proven before an account is "active", a student who loses their school login can still get back in through a verified personal email, and the server can never see a plaintext share. Six workstreams plus the Dashboard prompt from #1530, all behind existing state machinery, nothing enabled by default.

#### 💡 Feature Breakdown

**Provisional → active state machine** — New `UserKey.sssActivationState`. Shares are stored as `provisional`; a new `/keys/activate` route commits to `active` only when at least one recovery method is confirmed at the current share version. Existing records without the field are treated as active, so nobody currently on SSS is affected. Migrations can be abandoned and are purged after 30 days. Rollout is gated by the existing `auth.keyDerivation` tenant setting: prod (`learncard/config.json`) is still `web3auth`, so no prod user enters this path; staging/local are `sss` and will migrate legacy web3auth users into provisional SSS on next login.

**Confirmed recovery enrollment** — `RecoveryMethod.confirmedAt`. Every method must prove it works before it counts: email needs a 6-digit code from the actual email, phrase requires re-entering challenge words, backup file requires decrypting it once, passkey needs a full create + assert round trip. The signup email is no longer auto-enrolled.

**Progressive prompting (folds in #1530)** — New signups are *not* gated; an empty account has nothing to lose. @gerardopar's Dashboard recovery prompt handles the nudging: calm and snoozable once the account holds a credential, urgent and un-snoozable for public-computer sessions. This PR extends that same urgent treatment to migrations awaiting a confirmed method (`activationPending`), so a migrating user with a full wallet can't leave it half-secured. His `onCompleted` callback now fires at the two-phase proof step rather than the earlier "I saved it" step.

**Isolated email relay** — New `services/escrow-relay`. The client encrypts the share to the relay's public key (ephemeral P-256 ECDH → HKDF → AES-GCM); `lca-api` forwards ciphertext only and rejects anything shaped like a plaintext share. The relay has no database, no `SEED`, no imports from `lca-api`. Deployment/IAM checklist in `services/escrow-relay/ISOLATION.md`.

**Lost-login recovery** — New `/keys/recovery-session/*` routes. A student who lost their school account verifies a personal recovery email by OTP, gets a single-use 15-minute session token, recovers with any confirmed method, then binds a new sign-in. Old provider ID is removed, shares rotate, share history is purged (so a stale old device correctly lands in recovery instead of silently working). Entry point is "I lost access to my school login" on the login page.

**Lifecycle and auth invariants** — `atomicShareUpdate`/`atomicRecovery` are now actually used on the production paths (previously only in tests), with next-login reconciliation for the "server wrote, ack lost" case. User records are keyed by immutable provider ID instead of email, so a recycled school address can't inherit someone else's key. Sensitive routes require a fresh single-use DID challenge (`/keys/challenge`); all DID mismatches are hard failures instead of warn-and-continue.

**Telemetry hygiene** — Sentry no longer captures tRPC inputs on `/keys/*` (they contained tokens and shares), plus a redaction pass as defense in depth. Provider tokens moved from query strings to an `X-Auth-Token` header.

Also: a support runbook (`packages/learn-card-base/src/auth-coordinator/LOCKOUT-TRIAGE.md`) for triaging "I'm locked out" tickets by layer, and a residual-risk note in `RECOVERY.md`.

- **The relay is still a transient share holder** (as is the recipient mailbox). This is a bridge, not the destination — the enclave-backed escrow design retires the email share entirely. It's strictly better than today, where the API itself held the plaintext.
- **Provisional new users can't be rolled back.** A brand-new user who loses their device before confirming a method is unrecoverable by design. That's why the enrollment gate is forced and not dismissible.
- **Legacy grandfathering.** Records without `sssActivationState` / methods without `confirmedAt` on already-active accounts are treated as valid. Tightening that would lock out existing users.
- **Commit history.** The six workstreams interleave in `keys.ts`, `UserKey.ts`, and `sss-strategy.ts`, so intermediate commits aren't individually build-guaranteed. Review at HEAD; squash-merge.
- **CodeQL #296 (biased random from modulo)** — addressed with rejection sampling in `selectRecoveryPhraseChallengeIndices`. It only picked which 3 of 25 words to re-type (bias was ~1e-8), but it's a crypto package and the fix is four lines.
- **`RecoveryKeyData.confirmationCode` is required** in `@learncard/email-templates` (minor changeset; the version bumps on release). brain-service's `postmark.adapter.ts` template switch was updated to pass it through; it never sends that template, the arm exists for exhaustiveness.
- **Incidental changes riding along:** `qr-login` in both `sss-key-manager` and `learn-card-base` (moved to the `X-Auth-Token` header), the shared logger (redaction), and a new `account-sign-in-changed` email template arm sent after a lost-login rebind.

#### 🔍 Types of Changes

-   [ ] No
-   [x] Yes — the email share via relay is interim until enclave escrow lands. `SEED` should move to KMS with a tested restore drill (P1). Support needs admin tooling to see `UserKey` fields referenced in the runbook. Provisional migrations that expire after 30 days do so silently.

# Testing

Prod tenant config is still `web3auth`, so nothing changes there. To exercise this, use a tenant env with `auth.keyDerivation: "sss"` (`config.staging.json` / `config.local.json` already are) and run the relay locally (`services/escrow-relay`, needs `ESCROW_RELAY_PRIVATE_KEYS_JSON`, `POSTMARK_*`; README has the key generation command). Point the app at it with `VITE_ESCROW_RELAY_PUBLIC_KEY` / `VITE_ESCROW_RELAY_KEY_ID`, and `lca-api` at it with `ESCROW_RELAY_URL` / `ESCROW_RELAY_AUTH_TOKEN`.

1. **Progressive enrollment** — Sign up fresh: no gate, "Skip for now" works, no Dashboard prompt on an empty account. Claim a credential: the calm prompt appears on the Dashboard; snooze it and confirm it stays hidden for 7 days. Open setup from the prompt and confirm each method's proof step (code entry, challenge words, file re-decrypt, passkey) before "You're covered" appears.
1b. **Migration gate** — Migrate a web3auth account: the urgent "Finish securing your account" prompt appears with no snooze button, and the setup modal has no Skip until a method is confirmed.
2. **Fail-closed email** — Kill the relay or Postmark token, try email backup. Should error visibly, not succeed.
3. **Lost school login** — Enroll a phrase and verify a personal recovery email. Delete the Firebase user. From login, use "I lost access to my school login" → OTP → phrase → sign in with a new account. Wallet should have the same DID. Reuse the session token or wait 15 minutes: both should fail.
4. **Stale device** — After step 3, open the old device. It should land in recovery, not silently log in.
5. **Recycled email** — Delete a Firebase user, create a new one with the same email. It must get a fresh key record, not the old one.
6. **Mid-rotation failure** — Go offline (devtools) during recovery setup, then reload. Login should still work via reconciliation.
7. **Telemetry** — Run a flow with recognizable sentinel values and check the Sentry project and request logs for them. Should be zero.

Chromium + Safari for the web flows. iOS/Android native for passkey enrollment (PRF support varies — the passkey path is treated as secondary until we've verified it across platforms).

#### 🧪 Code Coverage

New Vitest suites: `lca-api/test/keys-{secret-leakage,lifecycle-invariants,recovery-confirmation,identity-rebind}.spec.ts` (fault injection, replay/expiry, recycled-email, DID-mismatch), `sss-key-manager/src/{api-client,email-relay-crypto}.test.ts`, `escrow-relay/src/relay.test.ts`. Run with `bun run test -- run` in each package (`bun run test:integration -- run` for the lca-api integration suites). Browser e2e for the full rebind flow still needs to run against staging.

# Documentation

**User-Facing Docs**

-   [ ] Tutorial
-   [ ] How-To Guide
-   [ ] Reference
-   [ ] Concept
-   [ ] App Flows

**Internal/AI Docs**

-   [x] **AGENTS.md** — `services/learn-card-network/lca-api/AGENTS.md` updated for the relay
-   [x] **Code comments/JSDoc**

**Visual Documentation**

-   [ ] Mermaid diagram

#### 💭 Documentation Notes

Internal-facing change; no user docs until `keyDerivation: "sss"` is on for real tenants. `RECOVERY.md` gained the threat note, `LOCKOUT-TRIAGE.md` and `escrow-relay/ISOLATION.md` are new. Full design and rollout plan live in `.sisyphus/plans/` (not committed).

# ✅ PR Checklist

-   [ ] I have **manually tested** common end-2-end cases — automated coverage yes; browser flows pending staging
-   [ ] I have considered **product analytics** for user-facing features — recovery-adoption telemetry is a P1 follow-up

-   [x] Code is backwards compatible — legacy records grandfathered, all schema changes additive, prod tenant remains on `web3auth`
-   [x] This change does not expose new public facing endpoints that do not have authentication — new `/keys/*` routes require a DID challenge or a verified-OTP recovery session; the relay requires `ESCROW_RELAY_AUTH_TOKEN`

---
Rovo Dev code review: <strong>Rovo Dev couldn't review this pull request</strong>
Something went wrong while reviewing this pull request.
[Troubleshoot code review errors](h

### ✨ PR Description
Purpose: Harden the SSS key management system by implementing confirmed recovery enrollment, lost-login identity rebind, and an isolated email relay.

Main changes:
- Added a multi-phase "identity recovery" flow to bind accounts to new login identities
- Implemented mandatory confirmation for recovery methods (phrase challenges, backup verification, email OTPs)
- Introduced an isolated email relay using ephemeral P-256 ECDH and AES-256-GCM encryption


packages/sss-key-manager/src/atomic-operations.ts
```diff
@@ -150,22 +151,35 @@ export async function atomicShareUpdate(
...
+        if (deviceStored) {
             try {
-                await storage.storeDevice(options.previousDeviceShare);
-                options.onRollback?.('Server storage failed, restored previous device share');
+                if (options.previousDeviceShare !== undefined) {
+                    await storage.storeDevice(options.previousDeviceShare);
+                    options.onRollback?.('Server storage failed, restored previous device share');
+                    rolledBack = true;
+                } else if (storage.clearDevice) {
+                    await storage.clearDevice();
```
TaylorBeeston: I reproduced this through the strategy with a simulated lost response: start a fresh account, let `/keys/auth-share` commit, then drop its reply. This rollback deletes the only saved device share, and reconciliation returns nothing. If no cached key survives a reload, the new account is stranded before recovery setup. Could we preserve the pending share and check whether the server committed before clearing it? A lost reply isn't the same as a rejected write.

— OMP

packages/learn-card-base/src/auth-coordinator/AuthCoordinator.ts
```diff
@@ -551,13 +665,160 @@ export class AuthCoordinator {
...
+                    error: errorMessage,
+                });
+            } else {
+                this.setState({
+                    status: 'identity_recovery',
+                    phase: 'choose_method',
+                    email,
+                    recoverySessionToken: this.state.recoverySessionToken,
+                    recoveryMethods,
+                    error: errorMessage,
```
TaylorBeeston: I reproduced this in a rendered-modal check: choose lost-login recovery, verify the email code, then submit 25 invalid recovery words. The coordinator records the error but resolves this call, while the modal only stops loading when it rejects. The error stays hidden and Recovering stays disabled; at the stack tip, Back is disabled too. Could we propagate the failure or have the modal handle the returned error state so the user can correct their phrase?

— OMP

apps/learn-card-app/src/components/learncard/MyLearnCardModal.tsx
```diff
@@ -401,6 +412,39 @@ const MyLearnCardModal: React.FC<MyLearnCardModalProps> = ({
...
+                                  });
+
+                                  if (!jwt || typeof jwt !== 'string') {
+                                      throw new Error('Failed to sign DID-Auth VP');
+                                  }
+
+                                  return jwt;
+                              };
+
+                              await keyDerivation.confirmRecoveryMethod!({
```
TaylorBeeston: On a provisional account, open My LearnCard → Account Recovery and confirm a phrase or backup. This path reports success without activating the account, unlike the provider-owned setup flow. Its passkey callback has the same gap, as does ScoutPass's profile-menu passkey callback. For a Web3Auth-backed provisional migration, the 30-day cleanup can then remove successfully configured recovery material. Could all setup entry points share the same confirmation, activation, and retry handling?

— OMP

services/learn-card-network/lca-api/src/routes/keys.ts
```diff
@@ -498,79 +1231,69 @@ export const keysRouter = t.router({
...
+            // lca-api deliberately cannot decrypt this payload. The isolated
+            // relay verifies that its encrypted recipient matches this
+            // server-verified address before sending.
+            await sendRecoveryKeyToEscrowRelay(input.relayPayload, targetEmail);
+
+            await addRecoveryMethodToUserKeyByAuthProvider(authProvider, {
+                type: 'email',
+                createdAt: now,
+                confirmationStatus: 'pending',
+                shareVersion: userKey.shareVersion ?? 1,
```
TaylorBeeston: To hit this, prepare an email recovery key on device A, rotate shares on device B, then let A's email-backup request arrive. The email contains the old version, but this records the new one. After code confirmation, lost-login recovery combines that old recovery share with the new server share and fails. Could the client send the encrypted share's version alongside the envelope, and the server validate it instead of assuming the latest version?

— OMP

apps/learn-card-app/src/providers/AuthCoordinatorProvider.tsx
```diff
@@ -1312,9 +1390,22 @@ const AuthSessionManager: React.FC<{
                             );
                         }
 
-                        await coordinator.initialize();
+                        if (coordinator.state.status === 'identity_recovery') {
+                            await coordinator.prepareIdentityRecovery({ method: 'device' });
```
TaylorBeeston: The lost-login chooser offers Sign In from Another Device, but approving its QR code reaches an unsupported recovery method here. To reproduce, verify the recovery email, choose that option, and approve from a signed-in device. The SDK has no `device` recovery branch and the recovery-session endpoint rejects it, so the received share never advances to replacement sign-in. Could we hide this option during lost-login recovery until it has a supported handoff? Ordinary device linking can stay available.

— OMP

services/learn-card-network/lca-api/src/models/UserKey.ts
```diff
@@ -403,3 +916,94 @@ export const deleteUserKey = async (contactMethod: ContactMethod): Promise<void>
...
+            $set: {
+                authProviders,
+                contactMethod: input.newContactMethod,
+                authShare: input.authShare,
+                shareVersion,
+                shareUpdatedAt: now,
+                previousAuthShares: [],
+                recoveryMethods,
+                keyProvider: 'sss',
+                sssActivationState: 'active',
```
TaylorBeeston: Is confirmed recovery intentionally waived after rebind? With escrow off, recover a lost login and bind a replacement sign-in: this purges old share history and leaves every method unconfirmed, but marks the account active. Losing the replacement device before reenrollment leaves none of those methods usable, and the provisional recovery prompt won't run. Could we keep enrollment pending until a new method is proved? Automatic escrow enrollment doesn't always cover this because it can be off or fail.

— OMP

TaylorBeeston (changes requested):
Rechecked this against the whole stack through #1565 (`36ca832`). The recovery proof steps are a good improvement, but interrupted setup can still discard the saved device share, and some setup paths report success without activation. I left the surviving findings inline, including reproduction steps; I'm not flagging the challenge/relay paths fixed later in the stack. The runtime checks used simulated network/storage boundaries, not a live staging recovery.

— OMP

services/escrow-relay/src/relay.ts
```diff
@@ -0,0 +1,185 @@
...
+});
+
+const tokenMatches = (authorization: string | undefined, expectedToken: string): boolean => {
+    const suppliedToken = authorization?.startsWith('Bearer ')
+        ? authorization.slice('Bearer '.length)
+        : '';
+    const supplied = Buffer.from(suppliedToken);
+    const expected = Buffer.from(expectedToken);
+
+    return supplied.length === expected.length && timingSafeEqual(supplied, expected);
```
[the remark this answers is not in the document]
Custard7: Fixed in 64777764d — both tokens are SHA-256 hashed to a fixed width before `timingSafeEqual`, so there's no length short-circuit.

services/escrow-relay/src/server.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    const chunks: Buffer[] = [];
+    let totalBytes = 0;
+
+    request.on('data', (chunk: Buffer) => {
+        totalBytes += chunk.length;
+
+        if (totalBytes > MAX_BODY_BYTES) request.destroy();
+        else chunks.push(chunk);
```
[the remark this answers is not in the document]
Custard7: Fixed in 64777764d. Note `destroy()` without an error doesn't emit `error`, so this wouldn't crash — but the client did get a reset instead of a response. It now replies 413 before dropping the socket, and returns 500 if the relay handler throws (previously an unhandled rejection).

packages/sss-key-manager/src/atomic-operations.ts
Custard7: Fixed in 1263adc24 / aa39f2483 / 9cc297250. `atomicShareUpdate` now distinguishes a definitive rejection (`ShareWriteRejectedError`) from an indeterminate outcome (network error, lost reply, 408/5xx, malformed success). On an indeterminate outcome the new device share is kept as a separate account-scoped pending entry alongside the previous one, and `reconcileShares` checks it against the current server share + expected DID and promotes it after reload — no server history needed. Persistence/reconcile run under a lock so overlapping ops can't clobber the pending entry. Regression tests cover fresh-account and rotation lost-reply cases (both failed before the fix) plus a real IndexedDB-backed case.

packages/learn-card-base/src/auth-coordinator/AuthCoordinator.ts
Custard7: Fixed in 42c2ae110 — `prepareIdentityRecovery()` now rethrows after recording the error state, so the modal's existing handlers show the friendly error, stop loading, and re-enable Back. All callers were checked; added coordinator + rendered-modal regressions (25 invalid words case).

apps/learn-card-app/src/components/learncard/MyLearnCardModal.tsx
Custard7: Fixed in 1a21405e4…785c47ad0. Every setup entry point (both providers' flows, My LearnCard → Account Recovery, ScoutPass profile menu, all passkey callbacks) now goes through one context operation, `runRecoverySetup(method, action)`, which only resolves after setup/confirmation **and** activation. If activation fails, the successful setup result is retained and a retry only re-runs activation (no re-created passkey or re-consumed phrase challenge). Method counts are now derived from server-confirmed types instead of optimistic increments (also fixes the ScoutPass over-count).

services/learn-card-network/lca-api/src/routes/keys.ts
Custard7: Fixed in 3b146b831. The client sends the envelope's `shareVersion`; the server validates it's the current version or still in retained history, records that version, and rejects unavailable versions with BAD_REQUEST before delivery. Older clients that omit it keep the current-version fallback (documented). Deploy order: server first.

apps/learn-card-app/src/providers/AuthCoordinatorProvider.tsx
Custard7: Fixed in a901b55f2 — the device option is hidden when the chooser is in the identity-recovery (lost-login) phase; ordinary recovery/device linking is unchanged. ScoutPass has no lost-login chooser.

services/learn-card-network/lca-api/src/models/UserKey.ts
Custard7: Not intentional — fixed in 7bceb5632 + 6ce55969f. Rebind now leaves the account `provisional` until a method for the new share version is confirmed, and the coordinator's `finishIdentityRecovery()` no longer hardcodes `active`, so the provisional recovery prompt runs right away. The 30-day cleanup only targets `keyProvider: 'web3auth'` records and rebind writes `sss` + clears `provisionalCreatedAt`; a regression verifies rebound key material survives cleanup past 30 days.

PR #1536, opened 2026-09-04
[LC-2125] Enforce live consent for AI Passport

TaylorBeeston:
# Overview

#### Relevant Jira issues

[LC-2125](https://welibrary.atlassian.net/browse/LC-2125), part of [LC-2118](https://welibrary.atlassian.net/browse/LC-2118). Companion gateway: WeLibraryOS/AI-Passport#81.

#### Context and goal

AI Passport needs current evidence of what a learner allowed, not a historical consent boolean. This exposes live terms, expiry, grant/contract revisions, and verified guardian state, then cuts the LearnCard clients over to the gateway’s selection-only request and session-based authentication.

#### Feature breakdown

- Record actual guardian approval with consent terms and audit transactions. Validate the presentation and current manager authority, including existing role-based managers; bind cached approval and headers to the child and refresh approval after credential preparation, immediately before the final mutation.
- Preserve original adult dedicated-AI grants. Legacy child grants need a genuine guardian-confirmed update. Contract changes or lost manager authority invalidate approval; removing a manager does not turn a historically child-approved grant into an adult grant. No fabricated notice-version acceptance.
- Enforce current app-owned terms when resolving learner context. Withdrawn/expired grants, explicit sharing exclusions, and revoked/suspended credentials do not become model context. An app with a contract cannot pull in unrelated app-issued credentials as a fallback.
- Send selected credential URIs and personal-field names, not raw credential JSON or an arbitrary endpoint. Refresh consented data for each formatter request; return server revision/cache metadata. Remove browser prompt caching and legacy DID-only AI authentication, and invalidate sessions/tickets across wallet or service changes, including in-flight negotiations.
- Make the admin benchmark issue real standards-valid signed credentials and select actual consented evidence, so it exercises the same proof boundary as a learner.

#### Important tradeoffs

Deploy the producer API first, then Passport#81, then these client changes. The formatter contract and legacy authentication cutover are intentionally breaking; there is no raw-context compatibility path. Guardian approval uses LearnCard’s existing authority rules, not a new legal-guardianship determination.

The full LC-2118 work, including Passport#82 lifecycle cleanup, remains a prerequisite for public monorepo import. LC-2119 provider/compliance approval is separate and still blocked.

#### Types of changes

- [x] Bug fix
- [x] New feature
- [x] Breaking change

# Testing

#### How to QA

1. Use a dedicated AI contract and a managed child. Continue with a currently authorized guardian, choose one credential, accept, and read the receipt through the contract owner’s SDK. It should report current guardian approval and exactly the selected URI.
2. Let approval expire while editing and submit. The real guardian flow must refresh approval; signing failure must not submit consent.
3. Withdraw the grant, remove sharing, expire a category, or remove guardian authority. Refresh learner context and confirm the affected data no longer authorizes processing. Existing adult grants should still work without a fabricated notice update.
4. Run the admin learner-context benchmark against the paired local gateway. Compare cold/warm requests and the returned consent revision; switch wallets or services while authentication is in flight and confirm the old session cannot return.

#### Verification

Review fixes in `194e08c36`:

- 291 consent, guardian-route, and app-store integration tests passed against isolated backing services.
- 79 shared auth/ingestion/chat-store tests and 47 guardian/formatter/consent-flow frontend tests passed.
- Chromium iframe regression passed: an authorized structured-context request succeeds, then withdrawal makes the next request return `FORBIDDEN`.
- Browser smoke check confirmed the admin loading spinner, disabled duplicate clicks, one selection request, and re-enabled controls after completion.
- Brain’s supported TypeScript checks, partner SDK checking, and the production LearnCard web build passed. Scoped lint passed with 13 existing errors explicitly baselined; no baseline expansion.
- Whole-app frontend `tsc` remains non-clean. A same-environment compiler comparison against the unchanged branch found no new diagnostics (2,017 baseline versus 2,014 after these fixes).

Earlier paired verification, before this review-fix commit: the real browser/HTTP guardian flow passed, including submission-time approval refresh. The paired formatter consumed one genuinely signed selected credential, kept learning facts, masked direct identifiers at the captured provider boundary, and restored the learner-facing output.

Provider transport for the formatter check was synthetic and local. This is not a claim that external provider access, DPA, or ZDR is approved. Chromium was exercised; native mobile and Safari were not.

# Documentation

Updated the Brain consent API README, public types, partner SDK migration guide, and release changeset. The partner SDK removal of `browser-hit`, `browser-miss`, `cacheLookupMs`, and `prewarmAgeMs` is documented and marked as a minor release on 0.x. No PR was merged.

— OMP

[LC-2125]: https://welibrary.atlassian.net/browse/LC-2125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LC-2118]: https://welibrary.atlassian.net/browse/LC-2118?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### ✨ PR Description
Purpose: Enforce live-consent and guardian-backed authorization for AI Passport data access and session management.

Main changes:
- Implemented server-side verification of live, unexpired consent for learner context requests
- Added guardian-signed VP requirements for child profile consent and contract updates
- Removed browser-side prompt caching and legacy DID-based AI authentication modes
- Integrated identity-aware session invalidation for wallets, profiles, and AI service changes


services/learn-card-network/brain-service/src/routes/index.ts
```diff
@@ -373,99 +373,90 @@ export const verifiedContactRoute = openRoute.use(async ({ ctx, next }) => {
...
-            return guardianClaims.iss === managerDidWeb || guardianClaims.iss === manager.did;
-        });
-
-        if (!guardianProfile) {
-            return next({
-                ctx: { ...ctx, isChildAccount: true, hasGuardianApproval: false },
-            });
+            if (
+                result.errors.length === 0 &&
+                result.warnings.length === 0 &&
```
Custard7: **Should-fix: no clock-skew tolerance.** Below, `claims.exp <= now + GUARDIAN_APPROVAL_MAX_TTL_SECONDS` requires `exp ≤ server_now + 300`. The client (`useGuardianGate.tsx:103-107`) signs `exp = floor((Date.now() + 300000)/1000)`, i.e. exactly `client_now + 300`. Any device clock ≥1s ahead of the server makes every guardian approval fail — and it fails silently into the bare `catch {}` below with no log line.

Suggest: add a skew margin server-side (e.g. `+ 60`), and/or clamp the client TTL to ~4 min. Add a test with a token whose `exp` is `now + 301`.

Minor: `result.warnings.length === 0` is new strictness on `verifyPresentation`; fine, but please `log.debug` the rejection reason in the catch so future didkit warnings don't become an undiagnosable outage.

packages/learn-card-base/src/helpers/aiPassportAuth.ts
```diff
@@ -27,47 +27,52 @@ const WEBSOCKET_TICKET_PATTERN = /^[A-Za-z0-9_-]{32,128}$/;
...
-    did: string
-): Promise<AiPassportAuthMode | undefined> =>
-    getAiPassportAuthMode(did) ?? (await authRequests.get(getAuthKey(did)));
-
-export const getAiPassportUrl = (path: string, did?: string): URL => {
-    const url = new URL(path, networkStore.get.aiServiceUrl());
-
-    if (did && getAiPassportAuthMode(did) === 'legacy') url.searchParams.set('did', did);
+// Invalidate even an away-and-back switch while a negotiation is suspended.
+walletStore.store.subscribe((state, previous) => {
```
Custard7, line 43: **Should-fix: WebSocket not invalidated.** These subscriptions clear HTTP auth state, but nothing bumps `connectionGeneration` or calls `disconnectWebSocket()`. `connectWebSocket()` (`chatStore.ts:563-565`) returns an existing OPEN/CONNECTING socket unconditionally, and the new identity checks (`chatStore.ts:600-606`) only run when *creating* a socket. Logout and chat-unmount do disconnect, so exposure is a wallet/profile or `aiServiceUrl` switch while chat stays mounted — but the changeset explicitly claims sockets are invalidated across wallet or service changes including in-flight negotiations.

Suggest: matching subscriptions in `chatStore.ts` that call `disconnectWebSocket()` (avoids the circular import), plus an OPEN- and CONNECTING-state switch test. The current chat mocks (`chatStore.test.ts:16-28`) hard-code identity so they can't catch this.

Alternatively (or additionally): bind socket reuse in `connectWebSocket()` (`chatStore.ts:563`) to the captured wallet DID + `getBackendUrl()`, since that early-return and the `onmessage` `ws === socket` guard never re-check identity.

services/learn-card-network/brain-service/src/accesslayer/consentflowcontract/relationships/read.ts
```diff
@@ -530,28 +561,38 @@ export const getConsentedDataBetweenProfiles = async (
...
+                    terms: flattenObject(convertDataQueryToNeo4jQuery(params)),
+                }),
+                cursor,
+                ...contractQueryParams,
+            })
+        ).match({
+            related: [
+                { model: Profile, where: { profileId: consenterProfileId } },
+                ConsentFlowTerms.getRelationshipByAlias('createdBy'),
+                { identifier: 'terms', model: ConsentFlowTerms, where: { status: 'live' } },
```
Custard7: **Should-fix / please confirm: drops one-time consents.** `oneTime` consents are written with `status: 'stale'` (`create.ts:166`, `update.ts:72,300`), so `getConsentedDataForDid` can no longer return them to the data provider after consent. Excluding `withdrawn` is a correct fix; excluding `stale` looks unintended. There's no `oneTime` test coverage anywhere in `brain-service/test/`, so this wouldn't have surfaced.

Either `status IN ['live','stale']` (or `<> 'withdrawn'`), or document the semantic change and how a provider is meant to consume a one-time grant in the changeset + README.

services/learn-card-network/brain-service/src/routes/contracts.ts
```diff
@@ -938,6 +939,19 @@ export const contractsRouter = t.router({
                 });
             }
 
+            if (!guardianIdentity) {
```
Custard7, line 941: **Design question (non-blocking):** once a term has any `guardianApproval` history, updating it requires `guardianIdentity` forever — even when `isChildAccount === false` because all managers were removed. README says this is intentional, but the practical effect is a formerly-managed adult can never modify or re-consent to that contract; the only path is withdraw + fresh consent. Is that the intended UX? If so, worth stating the escape hatch in the README and surfacing a clear message in the app.

apps/learn-card-app/src/pages/consentFlow/FullScreenConsentFlow.tsx
```diff
@@ -145,84 +145,19 @@ const FullScreenConsentFlow: React.FC<FullScreenConsentFlowProps> = ({
...
-            } else if (!successCallback || shouldDisableRedirect) {
-                closeAllModals();
-            }
-
-            if (!shouldDisableRedirect) {
-                const contractRedirectUrl = getConsentFlowContractRedirect({
-                    challenge,
-                    contractRedirectUrl: redirectUrl,
-                    domain,
+        await guardedAction(async () => {
```
Custard7: **Should-fix (x2).**

1. `useGuardianGate` now rejects on signing failure (good), but this `try/catch` lives *inside* the guarded callback, and `handleNextStep` awaits the gate without catching. Signing fails before the callback runs, so the error never reaches the toast; with PIN verify the modal is already closed and the user gets nothing. Consent correctly does not submit — it's just unrecoverable UX. Catch the outer `guardedAction` promise in both handlers and show a friendly retry message.

2. Refresh-at-submit happens *before* the slow part: `consentToContract` → `getTermsWithSharedUris` (`useConsentToContract.ts:23-33`) does sequential reads / encrypted uploads. A 5-min approval valid here can expire during that prep and the mutation then fails server-side rather than refreshing. Re-check/refresh approval immediately before the mutation, and add a submission-level test that advances time during prep.

apps/learn-card-app/src/hooks/useGuardianGate.tsx
```diff
@@ -88,67 +84,48 @@ export const useGuardianGate = (options: UseGuardianGateOptions = {}): GuardianG
...
+                throw new Error('Guardian signing identity is unavailable');
+            }
 
-                if (typeof vp === 'string') {
-                    guardianApprovalStore.set.setApproval(parentDid, vp, expiresAt);
-                }
-            } catch (error) {
-                log.error('Failed to create guardian approval VP:', error);
+            const parentWallet = await initWallet(parentPrivateKey, parentDid);
+            const expiresAt = Date.now() + Math.min(verificationTTL, DEFAULT_VERIFICATION_TTL);
```
Custard7: Cached expiry is stored in ms here, but the signed claim uses `floor(expiresAt/1000)` — so for up to ~1s the cache says valid while the token has already expired server-side. Align the cached value to the signed `exp` (and subtract a small margin so the client stops using it before the server would reject).

apps/learn-card-app/src/hooks/post-message/useLearnCardMessageHandlers.ts
```diff
@@ -431,205 +408,70 @@ export function useLearnCardMessageHandlers({
...
+                credentialUris: source.credentialUris,
+                personalFields: Object.keys(source.personalData ?? {}),
+                instructions: options.instructions,
+                detailLevel: options.detailLevel ?? 'compact',
+                includeStructuredContext: false,
+            }),
+        []
     );
 
     const prewarmLearnerContext = useCallback(
```
Custard7: **Should-fix: prewarm re-fires every render.** This depends on `initWallet`, which `useWallet` recreates each render, so the effect at ~L1448 reruns on every render. The removed cache-key suppression used to dedupe this; now each run issues `sendAppEvent` + an authenticated formatter call even when the server answers with a cache hit.

Stabilize the trigger on identity/app/service/selection changes and dedupe concurrent warmups. (Keep real SDK requests uncached — live authorization should stay mandatory.)

packages/learn-card-partner-connect-sdk/src/types.ts
```diff
@@ -520,21 +519,14 @@ export interface LearnerContextRawData {
     personalData?: Record<string, unknown>;
 }
 
-export type LearnerContextCacheStatus =
-    | 'browser-hit'
-    | 'browser-miss'
-    | 'backend-hit'
-    | 'backend-miss'
-    | 'structured';
+export type LearnerContextCacheStatus = 'backend-hit' | 'backend-miss' | 'structured';
```
Custard7, line 522: **Should-fix: undocumented public break.** Removing `'browser-hit' | 'browser-miss'`, `cacheLookupMs`, and `prewarmAgeMs` (exported via `src/index.ts`) will break external partners typing against them. No in-repo consumers remain, which is good, but the changeset marks `@learncard/partner-connect` as `patch` and doesn't enumerate the removals. Please list them with migration guidance and bump per the package's breaking-change policy (minor on 0.4.x).

Custard7 (changes requested):
Strong, well-reasoned PR — the guardian identity chain (`kid` → `iss` → `vp.holder` → `claims.iss` → current manager) and the contract-owner ↔ integration-owner check in `handleRequestLearnerContextEvent` close real holes, and the rewritten `useGuardianGate` tests are a big step up. Requesting changes for a handful of concrete gaps between what the description/changeset promise and what the code enforces; details inline.

**Should-fix**
1. Guardian approval TTL check has zero clock-skew tolerance (`routes/index.ts`).
2. Open/connecting WebSocket survives wallet / AI-service changes despite the changeset claim (`aiPassportAuth.ts` / `chatStore.ts`).
3. `getConsentedDataForDid` now drops one-time (`stale`) consents (`relationships/read.ts`).
4. Guardian signing failures never reach the consent UI's error handler (`FullScreenConsentFlow.tsx`).
5. Refresh-at-submit runs before the slow credential-prep step, so approval can still expire mid-submit (`FullScreenConsentFlow.tsx` / `useGuardianGate.tsx`).
6. Prewarm now re-fires on every render (`useLearnCardMessageHandlers.ts`).
7. Public partner-connect type removals shipped as `patch` with no migration notes (`partner-connect-sdk/src/types.ts`).

**Design question (non-blocking)**
- Permanent lockout for formerly-managed adults on any term with guardian history (`contracts.ts`).

**Nits**
- `getContractTermsForProfile` now hard-`parse`s legacy rows; consider `safeParse` + log on this hot path.
- Deleted `learner-context-perf.spec.ts` was the only iframe `REQUEST_LEARNER_CONTEXT` round-trip coverage; a small request → withdraw → FORBIDDEN test would cover the PR's core claim.
- `LearnerContextPromptTestPage` consent-selection load has no loading/disabled state (AGENTS.md loading-state rule); repeated clicks overlap.
- Missing explicit return types: `aiPassportAuth.ts` `guardCurrentIdentity` / `request` / `authFetch`, `chatStore.ts` `ensureChatAiPassportAuth`.
- No `oneTime` coverage anywhere in `brain-service/test/` — worth adding alongside fix #3.

**Also good:** revoked/suspended precedence fix in `app-store-listing/relationships/read.ts`, shared `transaction.date` for terms + transaction timestamps, and the fail-closed removal of legacy DID-only AI auth.

services/learn-card-network/brain-service/src/routes/contracts.ts
TaylorBeeston: Withdrawal is not an escape hatch here: fresh consent still sees the guardian history. Removing managers is not evidence of adulthood, so I kept that security boundary.

The README now explains the existing recovery path: `inbox.claimPendingGuardianLinks` can restore authority only when the verified guardian contact and a matching child-linked `GUARDIAN_APPROVED` credential satisfy that flow. Otherwise guardian-link/account recovery is a prerequisite; this PR does not introduce an adult-conversion bypass. The consent UI now surfaces a friendly approval-required message instead of failing silently.

— OMP

TaylorBeeston:
Addressed the review feedback in 194e08c36.

- Guardian validation allows 60s of issuance clock skew without accepting expired or overlong approvals. Advisory verifier warnings no longer reject a valid JWS; diagnostics use fixed categories, not raw warnings or claims. Cached expiry matches the signed claim and stops being reused 60s early.
- Signing failures reach the outer UI catch. Consent/update flows refresh approval after credential preparation, immediately before the mutation; the regression advances time during uploads.
- Wallet/profile/service changes invalidate OPEN, CONNECTING, and pending socket negotiations, clear private chat state, and prevent stale events/reconnects.
- Provider reads include authorized `stale && oneTime` grants, still exclude withdrawn grants, and skip malformed legacy rows without losing valid results.
- Prewarm uses a stable scope and only deduplicates in-flight work. Real SDK requests remain live-authorized. Non-JSON failures and partial structured credential reads have coverage.
- Restored the iframe request → withdraw → `FORBIDDEN` regression, added the admin loading/disabled state, and documented the public SDK removals with a minor release.

Verification: 291 backend + 79 shared-client + 47 frontend tests passed, plus the Chromium iframe regression and the actual admin pending-state/recovery smoke check. Brain/SDK typechecks, scoped lint, and the production web build passed. Whole-app frontend typechecking still has existing debt; the baseline comparison found no new diagnostics.

— OMP

Custard7 (approved):
Re-reviewed `194e08c36` against the previous round. All seven should-fixes are addressed, each with a test that targets the specific failure mode:

1. **Clock skew** — `GUARDIAN_APPROVAL_CLOCK_SKEW_SECONDS = 60` applied to `exp` and `iat` without extending the signed lifetime; `guardian-gated-routes.spec.ts` covers ±60s, legacy no-`iat`, and the 61s/361s rejection edges. Advisory verifier warnings are no longer fatal and rejections now `console.warn` a reason.
2. **WebSocket invalidation** — `connectionIdentity` snapshot + `isCurrentConnectionIdentity()` gates reuse, creation, and every `on*` handler; store subscriptions (wallet / switched profile / `aiServiceUrl` / `auth.did`) call `invalidateChatIdentity()`. `chatStore.test.ts` now covers OPEN-socket switches across all four axes and a CONNECTING-state switch, and the identity mocks are real stores instead of hardcoded values.
3. **One-time consents** — `terms.status = 'live' OR (terms.status = 'stale' AND terms.oneTime = true)`, with `consentflow.spec.ts` asserting stale one-time grants are still readable, withdrawn ones are not, and other stale terms don't leak through. Also switched to `safeParse` + skip on malformed stored grants.
4. **Signing failures reach the UI** — outer `try/catch` around `guardedAction` in both `handleAccept` and `handleNextStep`; guardian-required FORBIDDEN gets its own friendly message (`consentFlow.guardianApprovalRequired`, localized in en/es/fr/ar).
5. **Refresh after prep** — `beforeSubmit` hook in `useConsentToContract` / `useUpdateTerms` runs immediately before the mutation; cached expiry aligned to signed `exp` with a 60s margin in `guardianApprovalStore`. `FullScreenConsentFlow.test.tsx` exercises the slow-upload → re-PIN path without re-uploading, and the update-terms rejection path.
6. **Prewarm churn** — `useLearnerContextPrewarm(scopeKey, warm)` keys on identity/app/service/selection and coalesces concurrent equivalents; test proves unstable callbacks don't retrigger while real scope changes do.
7. **partner-connect break** — bumped to `minor` with an explicit migration paragraph in the changeset and `docs/sdks/partner-connect/methods.md`.

The design-question reply on the guardian-history lockout is a reasonable security stance, and the README + friendly UI message cover the UX gap I was worried about. Nice bonus: the new `tests/learner-context.spec.ts` restores the iframe `REQUEST_LEARNER_CONTEXT` → withdraw → FORBIDDEN round-trip.

CI is green (Chromatic is plan-limited, not failing).

**Two non-blocking nits for a follow-up, not this PR:**
- `useLearnCardPostMessage.ts` `ErrorCode = ... | string` collapses the union to `string` at the type level, so the named literals no longer give autocomplete or exhaustiveness. A branded `(string & {})` or `KnownErrorCode | \`${string}\`` pattern keeps the extensibility without losing the hints.
- `useLearnerContextPrewarm` is missing an explicit return type (AGENTS.md convention).

Approving.

PR #1541, opened 2026-09-07
feat: verify final P256 DIDAuth in VCALM consumers for LC-2167

TaylorBeeston:
I merged current main and adopted the CI-published DIDKit/SSI artifacts from #1576 unchanged. The checked-in binary, generated glue, default hosted URL, and bridge integrity pin now agree on SHA-256 `df7776606a04ca0cd44ad7799c2145698ffa416aeb93884dd511d330abe9d982`. I fetched the hosted artifact and exercised default initialization without a WASM override.

Both VCALM claim routes require authentication-purpose proofs. Final P-256 `ecdsa-rdfc-2019` proofs may omit `created`; legacy suites still require it, and malformed final-suite proofs now report the actual invalid field. Outgoing Ed25519 negotiation is unchanged.

I addressed the review followups: vendored the public P-256 test key so E2E doesn't require submodules, imported `beforeEach`, tightened the fixture hostname guard, added an explicit missing-document error, and removed the already-released LC-2160 changeset. Brain's changeset is now minor and documents the external-responder compatibility change.

The rerun also caught an interaction with main's encrypted inbox recovery. Inbox holders need a supported X25519 key-agreement method in addition to their authentication key. P-256 `did:web` holders can use a separate delivery key; signing-only P-256 `did:key` holders work with generic claim links but not inbox claims. Failed delivery encryption now returns an error without consuming the pending credential or exchange challenge. The fixtures and docs reflect that policy.

Verification against this artifact:
- 24/24 inbox/claim-link route cases passed in an isolated WASM-backed stack, with both submodules hidden from the runner. This includes encrypted recovery, escrow removal, replay rejection, and recovery after a signing-only holder is rejected.
- 25/25 docs Quickstart, Verify, Understand, and Methods scenarios passed.
- Proof schema regressions, Brain typecheck, focused inbox typecheck, targeted lint/formatting, artifact checks, and docs integrity checks passed.

For QA, build types and the DIDKit plugin, start isolated E2E services, then run `vitest run tests/inbox.spec.ts` from `tests/e2e`. The E2E setup clears databases; don't point it at shared development data.

The broader E2E TypeScript check still reports errors outside the focused files. Native binaries were not rebuilt or rerun in this followup, and this is not a full-CI or release-security approval.

— OMP

### ✨ PR Description
Purpose: Enable verification of P-256 `ecdsa-rdfc-2019` DIDAuth proofs in VCALM consumers and enforce stricter authentication purpose checks.

Main changes:
- Added `ecdsa-rdfc-2019` to `DataIntegrityCryptosuite` and updated `ProofValidator` to allow omitted `created` timestamps.
- Enforced `proofPurpose: 'authentication'` for `verifyPresentation` in claim-link and inbox-claim workflows.
- Added error handling for failed X2519 delivery encryption to preserve pending credentials.


Custard7 (commented):
## Review notes

The core code change is small and correct; the problems are all in artifact/release plumbing and CI. Since this is still draft, treating these as a punch list rather than a block.

### What looks good

- **`workflows.ts`** – adding `proofPurpose: 'authentication'` to both `handlePresentationForClaim` and `handleInboxClaimPresentation` is correct hardening. I checked all first-party responders (`ExchangeDidAuth.tsx`, `ClaimFromDashboard.tsx`, `getDidAuthVp` in `packages/plugins/vc/src/vc.ts`) and they already sign with `proofPurpose: 'authentication'`, so LCA/Scouts won't regress.
- **`ProofValidator` union** – narrowly scoped to `DataIntegrityProof` + `ecdsa-rdfc-2019`, and `vc.test.ts` covers the important negative cases (`created: null`, legacy suites still requiring `created`).
- **`lib/ssi` → `6d71e5f`** – verified this is the current tip of `learningeconomy/ssi` `main` (squash of ssi#13). `lib/didkit` unchanged is fine since didkit consumes ssi via `path = "../../ssi"`.
- README section is accurate and appropriately scoped.

### 🔴 Blockers before undrafting

**1. `DEFAULT_DIDKIT_WASM_URL` is stale → every default-init consumer gets a `LinkError`.**
`packages/plugins/didkit/src/didkit/index.ts` still points at `sha256-38e9e396…` while the checked-in binary + regenerated glue are `52cafdb5…`. The glue only declares imports for the new binary, so anything that inits with the default URL fails with `LinkError: Import #100 "wbg" "__wbindgen_cast_a80d8bd71297895c"`. This is what took down `docs-quickstart`, `docs-verify`, `docs-understand`, and `docs-methods` in the E2E run, and it's why `Check DIDKit WASM matches its published URL` is red. `update-didkit-wasm.yml` exists to publish + rewrite the URL atomically — it should run before this leaves draft.

**2. New E2E fixture reads from a submodule that CI doesn't check out.**
`inbox.helpers.ts:74` reads `lib/ssi/tests/secp256r1-2021-03-18.json` → `ENOENT` in CI (12 failures). Please vendor the P-256 test key into `tests/e2e/tests/fixtures/` (it's a public test vector) instead of reaching into `lib/ssi`.

**3. PR description SHA doesn't match the committed artifact.**
The description says the final WASM is `7c4eb57a…`; both committed copies of `didkit_wasm_bg.wasm` hash to `52cafdb5…`. The last commit (`6e9076c6e`) rebuilt after the "24/24" runs, so the verification described wasn't against the binary under review. Needs a re-run against `52cafdb5…` and an updated description.

**4. `.changeset/lc-2160-rust-security.md` is a resurrected, already-released changeset.**
It merged in #1540 and was consumed by the version-packages release (#1562). Re-adding it will emit a duplicate changelog entry and an extra bump. Please delete it and keep only `lc-2167-p256-consumer.md`.

### 🟡 Should address

- Branch is 2 commits behind `main` — merge before the WASM publish step so the hash doesn't go stale again.
- **External breaking change isn't called out.** Requiring `proofPurpose: 'authentication'` will reject third-party wallets that respond to the VPR with an `assertionMethod` VP (the previous `issuePresentation` default). First-party is fine, but the `network-brain-service` changeset should say so and probably be `minor` rather than `patch`.
- `beforeEach` is used in `inbox.spec.ts` but not imported — it only works because `vitest.shared.ts` sets `globals: true`. Import it alongside `afterEach, describe, test, expect` for consistency.
- `E2E_DID_WEB_HOST` regex `^localhost(?:[.-][a-z0-9.-]+)?$` allows e.g. `localhost.evil.com`. Test-only, but since the comment frames it as a guard, either tighten it to what compose needs or drop the pretence.
- `removeAuthentication()` relies on `documents.get(...)!` — fine in tests, but a thrown error with a clear message would be nicer than a TypeError.
- Zod `.or()` reports the *first* branch's errors, so a malformed final-suite proof will surface as "created required". `z.discriminatedUnion` or a `.superRefine` would give better messages.

### Suggested path to green

1. Drop `.changeset/lc-2160-rust-security.md`.
2. Vendor the P-256 test key into `tests/e2e`; import `beforeEach`.
3. Merge `main`, rebuild WASM once, run `update-didkit-wasm.yml` to publish `sha256-<hash>` and rewrite `DEFAULT_DIDKIT_WASM_URL` + `EXPECTED_DIDKIT_SHA256` together.
4. Re-run the 24 route cases against the committed binary and update the description's SHA.
5. Bump the `network-brain-service` changeset to `minor` with the external-responder note.

Custard7 (approved):
## Re-review @ `d1e5b87`

All four blockers and every should-address item from the prior pass are resolved, and CI is fully green (`tests/e2e`, `Playwright E2E`, `Test`, `Run workspace lint`, `Check DIDKit WASM matches its published URL`, `Validate DIDKit WASM build`, `publint + attw`). Approving.

### Prior blockers

| # | Item | Status |
|---|---|---|
| 1 | Stale `DEFAULT_DIDKIT_WASM_URL` → LinkError | ✅ Resolved upstream via #1576 and merged in. The PR no longer carries any WASM/glue/hash changes — diff is now 10 source files. |
| 2 | Fixture read from un-checked-out submodule | ✅ Vendored to `tests/e2e/tests/fixtures/secp256r1-2021-03-18.json` with provenance comment. |
| 3 | Description SHA ≠ committed artifact | ✅ Moot — no binary in this PR anymore. |
| 4 | Resurrected `lc-2160-rust-security.md` changeset | ✅ Deleted. |

### Prior should-address

- ✅ `network-brain-service` bumped to `minor`; changeset **and** `docs/how-to-guides/interoperate-with-learncard.md` now explicitly call out that `assertionMethod` VPs are rejected and `issuePresentation`'s default must be overridden.
- ✅ `beforeEach` imported.
- ✅ `E2E_DID_WEB_HOST` regex tightened to a single label.
- ✅ `removeAuthentication()` throws a descriptive error.
- ✅ Zod union → `.superRefine` on `created`, with a test asserting malformed final-suite proofs report the actually-missing field.

### New in this revision

**Inbox delivery encryption guard (`workflows.ts`).** Good catch that a signing-only P-256 `did:key` has no X25519 `keyAgreement`, so `createDagJwe` yields zero recipients and the recovery copy would be unrecoverable. The new guard throws before finalization, and the aggregate `BAD_REQUEST` is raised *before* `exhaustExchangeChallengeForToken`, so the pending credential and challenge survive for retry. I verified the ordering in source. The new e2e case asserts the `X25519` message and then succeeds on the **same** challenge with a `did:web` holder.

The happy-path test also hits Neo4j directly to assert escrow is wiped, the delivery decrypts with the holder's key, and the service key **cannot** decrypt it. Strong privacy assertion — nice.

### Non-blocking nits

1. The encryption failure flows through the existing per-credential `catch`, so the issuer receives an `ISSUANCE_ERROR` webhook and a `logCredentialFailed` activity for what is a retryable, still-`PENDING` credential. Probably acceptable, but if that webhook drives any "failed" state on the issuer side, consider short-circuiting the notification when `deliveryEncryptionFailed` is the cause.
2. `Proof.created` is now `string | undefined` in the public `@learncard/types` type. No direct `proof.created` reads exist in `packages/` or `apps/`, so nothing breaks, and the `minor` bump covers it — just flagging for changelog readers.

PR #1565, opened 2026-09-10
feat(recovery): optional recovery PIN for instant escrow release

Custard7:
Stacked on #1559 (client strategy) → #1556 → #1529. Adds an optional recovery PIN so a user who set up nothing else can get back in immediately instead of waiting out the 7-day hold, and cleans up the recovery UX around it.

## Why

The 7-day hold is the right last resort, but it's a bad first option for the common case: someone signs up on a public or short-lived device, never sets up a passkey or phrase, and loses the device share. A PIN checked inside the enclave, with a hard attempt limit, gives a convenient path without changing who can decrypt what.

Not a second custodial share. The escrow share that already exists gets a second *release policy*:

- **`hold`** — 7-day delay + notify + cancel (unchanged).
- **`pin`** — released immediately if the client presents a valid PIN proof.

The PIN is optional at signup and can be set, changed or removed later. No PIN → behaviour is exactly as before.

## How it works

- Client derives `Argon2id(pin, salt)` and puts the result *inside* the enclave-sealed blob as `pinVerifier`. A leaked database has nothing to brute-force offline.
- Recovery: `/keys/escrow/recover` with `releasePolicy: 'pin'` creates a hold with `releaseAfter = now`; `/keys/escrow/complete` carries `pinProof`; the enclave requires a matching proof for every `pin` hold. A `hold` hold cannot be short-circuited with a proof.
- **10 lifetime failed attempts**, reserved atomically before the enclave is called. Exhaustion disables the PIN; the 7-day path remains. Attempts are refunded when the failure wasn't a wrong PIN (claim conflict, enclave unavailable), so transient outages can't lock someone out.
- **PIN and hold requests are independent.** A pending 7-day hold is never cancelled or reset by a PIN attempt, so someone holding a stolen login token can't keep resetting the real owner's clock. One pending hold per (account, policy); PIN starts are also throttled per IP.
- Every attempt is a fresh hold (`/complete` claims-then-releases, fail-closed), so a wrong PIN burns the hold and the client starts over. Nothing PIN-related is written to storage on the recovering device.
- Setting or changing a PIN is a share rotation (the client never keeps the recovery share). Any forced rotation — including recovering *with* the PIN — retires it, and the app immediately asks for a new one. Setting a PIN requires automatic recovery to be enrolled; it never silently re-enrolls an account that opted out.
- **Setting a PIN completes key activation**, the same as a passkey or phrase. Plain escrow enrollment does not — the user hasn't chosen anything yet.
- **Restarting a lost 7-day hold.** The hold's release is bound to the browser that started it. If that browser is gone, `recover` with `restart: true` supersedes the pending hold and issues fresh secrets with a full 7-day wait, at most once per 24h (`ESCROW_HOLD_RESTART_MIN_AGE_MS`). Anyone who can do this already holds the login token, so it changes nothing about who can get in — it only resets a timer. Without it a user could be locked out for up to 37 days.

Threat statement: with a PIN set, login-token compromise + a correct guess (≤ 10 in 1,000,000 for a random 6-digit PIN) = account. Without a PIN, unchanged.

## What

**`@learncard/sss-key-manager` / `@learncard/types`** — `pinVerifier` on the escrow blob (backwards compatible), `escrow-pin.ts` (normalise, validate 6–12 digits + denylist, derive proof, salt), `escrow-pin` recovery method, `setEscrowPin` / `clearEscrowPin`, `startEscrowRecovery({ restart })`. Typed errors: `EscrowPinMismatchError { attemptsRemaining }`, `EscrowPinLockedError`, `EscrowPinThrottledError`, `EscrowPinUnavailableError`, `EscrowHoldRestartThrottledError { retryAfter }`. Server error strings the client matches on live in `@learncard/types`.

**`lca-api`** — `escrowPin` on `UserKey` (salt, lifetime counter, `disabledAt`, `shareVersion`) with atomic reserve/refund; `releasePolicy` / `cancelReason` on holds; per-(account, policy) pending-hold index with in-place migration in `ensureHoldIndexes`; `restart` on the start route with a 24h throttle; software enclave verifies the proof with `timingSafeEqual`; status exposes `escrowPin` (never the verifier). `pinProof` / `pinVerifier` redacted from logs. `escrow-enclave/README.md` notes that **Nitro mode must move the counter into enclave state** (`TODO(escrow-nitro)`).

**`learn-card-base`** — `AuthCoordinator.setEscrowPin` / `clearEscrowPin`, `recover({ method: 'escrow-pin', pin })`, `startEscrowRecovery({ restart })`, and `escrowPin` + `escrowEnrollment` on the `ready` state. PIN status is invalidated when a mutation starts and reconciled afterwards.

**LearnCard app**

- *Onboarding* — new "Add a recovery PIN" step between profile and celebrate, shown only when enrollment has already succeeded. Skippable.
- *Dashboard banner* — three tiers, one banner max: enrolled + no PIN → "Finish securing your account" (PIN primary, passkey/phrase secondary); enrolled + PIN, no other method → calm "Add a backup way in", gone after two snoozes; any method set → nothing. Passkey copy is platform-aware (Face ID/Touch ID, Windows Hello, Android biometrics, generic). Non-escrow tenants unchanged.
- *Protect Your Account modal* — grouped "protection summary" at the top (automatic recovery toggle, PIN set/change/remove, each configured method), a section label above the method tabs, quieter tab bodies, always-closable header X.
- *Recovery flow* — "Do you have a recovery PIN?" first, only when the account has one, falling through to the hold flow on lockout. The waiting state is a single status card with a live countdown, progress bar, and "Restore my account" only when releasable. If a request already exists from a lost browser: "Lost that browser? → Start over."
- *After a PIN recovery* — the PIN setup overlay opens immediately with copy explaining the PIN just used has been retired.
- *Debug tab* — "Automatic Recovery & PIN" section (enrollment, PIN status, pending holds, server state, test actions). `lca-api/.env.example` documents the enclave env vars.

Also fixed along the way: PIN confirm always reported a mismatch (compared against stale state); the recovery-setup session check re-ran on every state change and force-refreshed the Firebase token in a loop; recovery-email add/verify from the provider-owned modal was missing the server challenge (UNAUTHORIZED).

## How to review

Code hot-spots:

- `services/learn-card-network/lca-api/src/routes/escrow.ts` — `startRecovery`: a `pin` start only ever supersedes another `pin` request; `restart` only applies to `hold`, only past the min age. `completeRecovery`: validate → reserve attempt → claim → enclave → mark; refund on non-mismatch failure.
- `src/models/EscrowHold.ts` `ensureHoldIndexes` — migration order and idempotency.
- `src/models/UserKey.ts` `reserveEscrowPinAttempt` / `refundEscrowPinAttempt` — single guarded `findOneAndUpdate` each.
- `src/services/escrow-enclave/softwareEnclave.ts` `releaseEscrow` — proof required for every `pin` hold regardless of `releaseAfter`.
- `packages/learn-card-base/src/auth-coordinator/AuthCoordinator.ts` `refreshEscrow` — enrollment is awaited before reading PIN status.
- `apps/learn-card-app/src/providers/AuthCoordinatorProvider.tsx` — `setEscrowPin` (activation), the `recoveredWithPinRef` path that opens the PIN overlay after recovery, `getDidAuthHeaders` (challenge).
- `apps/learn-card-app/src/components/recovery/RecoveryBanner.tsx` — `activeTier`.

### Manual checklist (whole stack)

Run with the escrow block from `lca-api/.env.example` in `lca-api/.env`, plus `ESCROW_HOLD_DURATION_MS=60000` and `ESCROW_HOLD_RESTART_MIN_AGE_MS=60000` so waits are one minute. `bun run lc dev learncard local full` from `apps/learn-card-app`. Use two browser profiles as "devices" and the auth debug tab to confirm server state.

**Setup**
- [ ] Sign up → onboarding shows "Add a recovery PIN" after profile → set `135790` → celebrate. Debug tab shows `escrowEnrollment: enrolled`, PIN set, and the amber "Finish securing" banner is gone.
- [ ] Sign up again and skip the PIN → dashboard shows "Finish securing your account" with "Set a 6-digit PIN". Set it from there → banner switches to the calm "Add a backup way in".
- [ ] Snooze the calm banner twice → it stays hidden.
- [ ] Open Protect Your Account: automatic recovery toggle, PIN row (Set / Change / Remove) and X all work. Change PIN with a matching confirmation succeeds (this was broken). Remove PIN → banner returns to "Finish securing".
- [ ] Add an email recovery method → "Send Verification Code" no longer says UNAUTHORIZED.
- [ ] Console: no repeating `securetoken.googleapis.com` 400s while the modal is open.

**PIN recovery**
- [ ] Device B: forget device → sign in → "Do you have a recovery PIN?" is the first step. Wrong PIN → "9 attempts left". Correct PIN → restored in seconds.
- [ ] Immediately after: "Set a new recovery PIN" overlay with the "retired" explanation. Setting a new PIN works; the old one no longer does on a third device.
- [ ] An account with no PIN goes straight to "Start a 7-day recovery".
- [ ] Exhaust 10 wrong attempts → falls through to the 7-day flow; PIN shows as disabled in the debug tab.

**7-day hold**
- [ ] Start a hold on device B → status card with countdown and progress bar; no primary button until ready. After 60s → "Ready to restore" → restore works.
- [ ] Start a hold on B, then attempt a wrong PIN from device C → B's `releaseAfter` unchanged.
- [ ] Start a hold on B, then clear B's storage (simulate a lost browser) and sign in there again → "A recovery request is already waiting" card. Within 60s "Start over" shows the throttle callout with a retry time; after 60s it supersedes the old hold (debug tab: `cancelReason: superseded`) and the new countdown begins.
- [ ] Cancel a pending hold from a signed-in device → the recovering device sees it cancelled and can restart.

**Regression**
- [ ] Passkey / phrase / backup recovery unchanged. Scouts app signs in. No missing-translation strings on the new screens in es/fr/ar.

Automated:

```sh
(cd packages/sss-key-manager && bunx vitest run)
(cd packages/learn-card-base && bunx vitest run src/auth-coordinator src/stores)
(cd apps/learn-card-app && bunx vitest run src/components/recovery src/components/onboarding src/providers)
(cd services/learn-card-network/lca-api && POSTMARK_SERVER_TOKEN='' POSTMARK_FROM_EMAIL='' \
  bun run test:integration -- run test/keys-escrow.spec.ts test/models/escrow.spec.ts)
```

## Known gaps

- Notification fan-out is still the placeholder from #1556 — "superseded" fires the event but nothing is sent yet.
- Nitro counter placement (above). Software-mode counter in Mongo is host-trusted, which matches the current software enclave.
- Two `/complete` calls racing on the 10th attempt can lock the PIN before the loser's refund lands. Requires a simultaneous infra failure on the final attempt; the 7-day path still works. Follow-up.
- The dashboard "Set a 6-digit PIN" action opens the full Protect Your Account modal rather than the PIN overlay directly. Fine, but a dedicated entry point would be one tap shorter.

### ✨ PR Description
Purpose: Introduce an optional recovery PIN for instant escrow release, bypassing the standard 7-day waiting period.

Main changes:
- Added PIN verification logic with Argon2id proofs and a 10-attempt lifetime limit.
- Implemented `setEscrowPin` and `clearEscrowPin` with escrow share rotation in `AuthCoordinator`.
- Integrated PIN setup overlays, recovery flow steps, and management UI in the app.


apps/learn-card-app/src/components/recovery/RecoveryPinResetBanner.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { IonIcon } from '@ionic/react';
+import { alertCircleOutline, closeOutline } from 'ionicons/icons';
+import { m } from '../../paraglide/messages.js';
```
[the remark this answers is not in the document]
Custard7: Not a bug — `paraglide/messages.js` is `export * as m from './messages/_index.js'`, so the named `{ m }` import is valid and already used on `main` (`StalledMigrationOverlay.tsx`, `ErrorOverlay.tsx`). Leaving as-is.

apps/learn-card-app/src/components/debug/AuthDebugTab.tsx
```diff
@@ -238,6 +263,184 @@ export const AuthDebugTab: React.FC = () => {
...
+            setEscrowActionError(e instanceof Error ? e.message : String(e));
+        } finally {
+            setEscrowActionLoading(false);
+        }
+    };
+
+    const handleSetEscrowPin = async () => {
+        if (!escrowPinInput || escrowPinInput.length < 4) {
+            setEscrowActionError('PIN must be at least 4 characters');
+            return;
```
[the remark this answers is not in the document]
Custard7: Fixed in adaa4f1bf — the debug tab now calls the canonical `validatePin()` from `@learncard/sss-key-manager` (length + denylist), so it can't drift from the real rules.

Custard7:
Re: gitstream issue_3 (`writeRecoveryPinPromptFlag` racing the cleared PIN state)

The described mechanism doesn't hold: `AuthCoordinator.setEscrowPin` awaits `refreshEscrowPinStatus` in its `finally`, so `ready.escrowPin` is already reconciled by the time the provider writes the prompt flag. There was a narrower gap though — the `first-time` prompt condition never looked at PIN status at all. adaa4f1bf adds `readyPinEnabled !== true` to that guard so an account with a PIN set can never be shown the first-time setup overlay, regardless of flag/re-render ordering (`!== true` rather than `=== false` so legacy strategies returning a bare enrollment string still get prompted).

packages/sss-key-manager/src/sss-strategy.ts
```diff
@@ -2187,45 +2418,43 @@ export function createSSSStrategy(config: SSSStrategyConfig): SSSKeyDerivationSt
...
-                                    primaryDid: pending.primaryDid,
-                                    authShare: { encryptedData: share, encryptedDek: '', iv: '' },
-                                },
-                                {
-                                    ...buildHeaders('', didAuthVp, tenantId),
-                                    'X-Auth-Token': params.token,
-                                }
-                            );
+                atomicRecovery(pending.recoveryShare, pending.authShare, {
+                    storeDevice: share => storage.storeDeviceShare(share, activeStorageId),
```
TaylorBeeston: The first lost-response fix preserves the right device share, but Retry still breaks it. I reproduced this by simulating a committed `/rebind` with a lost response, then calling `completeIdentityRecovery` again. Retry saves a fresh split before the consumed token is rejected, overwriting the share that matched the server. The stored pair then can't reconstruct the account key. Could we reuse the pending split and reconcile the committed operation before retrying? Otherwise a recoverable connection error becomes a lockout risk after reload.

— OMP

services/learn-card-network/lca-api/src/routes/keys.ts
```diff
@@ -1060,7 +1069,10 @@ export const keysRouter = t.router({
 
             assertDidOwner(userKey, ctx.user.did);
 
+            // Legacy accounts with no prior record are created by storeAuthShare as a
+            // provisional 'sss' key; treat markMigrated as an idempotent no-op for them.
             if (userKey.keyProvider !== 'web3auth') {
+                if (userKey.sssActivationState === 'provisional') return { success: true };
```
TaylorBeeston: This fixes the immediate migration error, but loses the fallback for older Web3Auth users without a UserKey record. Start migration for one of those users, stop before confirming recovery, then advance past 30 days. The initial write labeled the record `sss`, and this no-op never records that it's a migration, so the cleanup's `web3auth` filter skips it. Could we preserve migration provenance so abandoning this path gets the same expiry and legacy fallback as other migrations?

— OMP

services/learn-card-network/lca-api/src/routes/escrow.ts
```diff
@@ -473,23 +612,152 @@ export const escrowRouter = t.router({
...
+                    current.shareVersion !== hold.shareVersion ||
+                    current.escrowBlob?.envelope.ciphertext !== expectedCiphertext
+                ) {
+                    throw new TRPCError({
+                        code: 'FORBIDDEN',
+                        message: ESCROW_PIN_UNAVAILABLE_MESSAGE,
+                    });
+                }
+                if (current.escrowPin.failedAttempts >= ESCROW_PIN_MAX_ATTEMPTS) {
+                    await lockPin(hold, current);
```
TaylorBeeston: There's a race here even with a correct PIN. After nine failures, let two completions read the same pending request. Pause A after it reserves attempt ten; B can't reserve, sees ten, and disables the PIN before A checks it. A then loses the claim, and the refund excludes disabled PINs. This is source-traced, not runtime-reproduced; seven-day recovery still works. Could lockout distinguish an in-flight reservation from a verified wrong guess? It doesn't require an infrastructure failure.

— OMP

apps/learn-card-app/src/components/recovery/RecoverySetupModal.tsx
```diff
@@ -542,402 +579,438 @@ export const RecoverySetupModal: React.FC<RecoverySetupModalProps> = ({
...
-                            <p className="text-sm text-grayscale-600 leading-relaxed">
-                                {m['recovery.passkeyUseFaceId']()}
-                            </p>
-
-                            <div className="p-3 bg-amber-50 border border-amber-100 rounded-2xl">
-                                <p className="text-xs text-amber-700 leading-relaxed">
-                                    {m['recovery.passkeyDesktopOnly']()}
-                                </p>
-                            </div>
+            {configuredCount < tabs.length && (
```
TaylorBeeston: Configure every displayed recovery method, then click Change on a phrase or backup. The button sets `showUpdateForm`, but this guard hides every form once the configured count reaches the tab count, so nothing opens. Could we also render the selected form when an update was explicitly requested? Being fully protected shouldn't prevent someone from replacing a lost phrase or changing their backup password.

— OMP

apps/learn-card-app/src/components/recovery/RecoveryPinSetupOverlay.tsx
```diff
@@ -0,0 +1,194 @@
...
+            setConfirmPin('');
+            return;
+        }
+        setError('');
+        setConfirmPin(p);
+        setStep('saving');
+
+        try {
+            if (!coordinator.setEscrowPin) throw new Error('PIN setup is unavailable');
+            await coordinator.setEscrowPin(p);
```
TaylorBeeston: For a new provisional account with automatic recovery enrolled, complete the first-time PIN prompt. This calls the base coordinator directly, bypassing the app wrapper that also activates the account, then shows success. The PIN is saved but activation stays provisional. Could we use one activation-aware operation here and in settings, and keep activation failures retryable before showing success? Otherwise the same PIN setup has different results depending on which prompt opened it.

— OMP

TaylorBeeston (changes requested):
The first lost-rebind-response fix helps, but Retry still overwrites the share it preserved. I reproduced that at this head with a simulated network boundary and left the remaining PIN, migration, and settings issues inline. The focused SDK, coordinator/config, and recovery UI suites pass, so these specific failure sequences still need regression coverage. I'd fix the retry and activation behavior before calling the stack ready.

— OMP

packages/sss-key-manager/src/sss-strategy.ts
Custard7: Fixed in cd3ea6a59. Retry now reuses the first attempt's split instead of storing a fresh one, and before re-POSTing `/rebind` it checks with the new sign-in token whether the previous rebind committed (pending device share + current server share reconstruct the primary DID). If committed it finishes locally without calling `/rebind` again; only a definitive rejection discards the pending share, and never one that matches the server. After a reload, reconciliation recovers the committed rebind from the pending entry. Regressions cover lost-reply→retry, lost-reply→reload, and definitive rejection.

services/learn-card-network/lca-api/src/routes/keys.ts
Custard7: Fixed in ed758f4a5. `markMigrated` now records `web3auth` provenance for recordless migrations until activation (preserving the original expiry, and atomically excluding already-active SSS records), so abandoned migrations get the same 30-day cleanup and legacy fallback. Regression covers first-ever UserKey → no confirmation → cleanup after 30 days.

services/learn-card-network/lca-api/src/routes/escrow.ts
Custard7: Fixed in 37e4dfeec + 3a6cda72c. Lockout is now driven only by verified enclave mismatches (new `verifiedFailedAttempts`, atomically disabling at the limit). A completion that can't reserve gets the existing retryable throttle error instead of disabling the PIN. The 10-evaluation lifetime budget is still enforced through reservations (including across successful releases), refunds cover claim conflicts, enclave failures, and pre-enclave exceptions, and legacy charged counters stay charged without being treated as verified mismatches. The race is covered by an integration test.

apps/learn-card-app/src/components/recovery/RecoverySetupModal.tsx
Custard7: Fixed in 0b6d580a5 — the selected form renders when an update was explicitly requested, even when every method is configured. Test covers Change on phrase and backup.

apps/learn-card-app/src/components/recovery/RecoveryPinSetupOverlay.tsx
Custard7: Fixed in ab3950fc4 / f73834463 / 6f16b7d83. All PIN setup/change prompts (first-time overlay, reset path, settings) go through the same activation-aware `runRecoverySetup` operation: success only shows once the PIN is saved and activation succeeds, activation failure is retryable without re-saving the PIN, and prompt-flag bookkeeping still runs. Clearing a PIN never activates.

PR #1567, opened 2026-09-11
fix: native apps fail to boot with "TenantConfig endpoint /__tenant-config returned invalid JSON"

Custard7:
## Problem

Every native cold start on the current production Capgo bundle shows the **Configuration error** screen:

> TenantConfig endpoint /__tenant-config returned invalid JSON

### Root cause

`resolveTenantConfig()` on native does a relative `fetch('/__tenant-config')` against the Capacitor webview origin. That route only exists as a Netlify edge function on web; Capacitor's asset server SPA-fallbacks unknown paths to `index.html` with **HTTP 200**, so `response.ok` is true and `response.json()` fails.

Before #1505 (`c7bbf6e9e`, LC-1984) that parse failure hit the generic `catch → return null` and resolution fell through to the baked `public/tenant-config.json`. #1505 changed it to throw a `TenantConfigResolutionError`, which is explicitly rethrown past the baked/cache fallbacks and lands on `renderConfigurationError`. The first production deploy carrying that change was `chore(release): version packages (#1561)` on Sep 10, which is when the OTA bundle went out.

## Fix

1. **`bootstrapTenantConfig.ts`** — pass `offlineOnly: Capacitor.isNativePlatform()` so native never attempts the relative overlay fetch and boots from the baked config (the behavior it effectively had before).
2. **`resolveTenantConfig.ts`** (safety net) — when a baked `mergeBase` exists, a non-JSON overlay response is non-fatal: `log.warn`, `config:fetch_error` debug event, Sentry `onFetchFailure` breadcrumb, then fall back to baked. Web with no baked config still throws as before.

## Tests

- `resolveTenantConfig.test.ts`: baked config + HTML-200 overlay → resolves to baked
- `resolveTenantConfig.test.ts`: no baked config + HTML-200 overlay → still throws `returned invalid JSON`
- 13/13 passing

## Rollout note

Ships via the next production deploy → Capgo OTA. Users on the broken bundle keep hitting the error on cold start until then; rolling the Capgo prod channel back to the pre-#1561 bundle is the only immediate mitigation.

### ✨ PR Description
Purpose: Prevent native app boot failures caused by invalid JSON responses from the `/__tenant-config` endpoint.

Main changes:
- Updated `bootstrapTenantConfig` to pass `offlineOnly` flag when `Capacitor.isNativePlatform()` is true
- Modified `resolveTenantConfig` to treat invalid JSON as non-fatal if a baked config exists
- Added test cases for fallback behavior and rejection when no baked config is present

---

### DIFF
--- a/.changeset/native-tenant-config-boot.md
+++ b/.changeset/native-tenant-config-boot.md
@@ -1,0 +1,8 @@
+---
+"learn-card-app": patch
+"learn-card-base": patch
+---
+
+fix: native apps no longer fail to boot with "TenantConfig endpoint /__tenant-config returned invalid JSON"
+
+Native builds skip the relative `/__tenant-config` overlay fetch (it only exists as a web edge function; Capacitor SPA-fallbacks it to `index.html` with HTTP 200) and boot from the baked config. As a safety net, `resolveTenantConfig` now treats a non-JSON overlay response as non-fatal whenever a baked config exists.
--- a/apps/learn-card-app/src/config/bootstrapTenantConfig.ts
+++ b/apps/learn-card-app/src/config/bootstrapTenantConfig.ts
@@ -6,6 +6,7 @@
  */
 
 import * as Sentry from '@sentry/browser';
+import { Capacitor } from '@capacitor/core';
 
 import type { TenantConfig } from 'learn-card-base';
 import {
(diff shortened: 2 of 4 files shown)


gerardopar (approved):
LgTM!!

PR #1574, opened 2026-09-15
fix: vetpass boot fails on partial auth overlay from /__tenant-config

Custard7:
## Problem

https://vetpass.app shows a **Configuration error** on boot:

```
Invalid TenantConfig from fetch /__tenant-config (overlay)

auth.firebase
  Required when auth.provider is firebase
```

The edge function serves the vetpass overlay from `environments/vetpass/config.json`, whose `auth` block is intentionally sparse (overrides only):

```json
"auth": { "keyDerivation": "web3auth", "web3Auth": { ... } }
```

`fetchFreshConfig` pre-validated this overlay with `parsePartialTenantConfig`, which was `tenantConfigSchema.partial()`. Zod's `.partial()` is **shallow** — once `auth` is present, the full `tenantAuthConfigSchema` runs against the un-merged overlay, `provider` defaults to `firebase`, and the `superRefine` demands an `auth.firebase` block that was supposed to come from `DEFAULT_LEARNCARD_TENANT_CONFIG` during the merge. The merge is never reached.

LearnCard wasn't affected because its overlay doesn't touch `auth` at all; vetpass is the only tenant that sends a partial `auth` object.

## Second (latent) bug fixed by the same change

`.partial()` still applied every `.default()`, so parsing the overlay *injected* `provider: 'firebase'`, `sss.serverUrl`, `schemaVersion`, etc. into it — and those then overwrote baked-config values in `deepMerge`. Any native build with a non-default `auth.provider` would have had it silently reset to `firebase` by an overlay that touched `auth` at all.

## Fix

`parsePartialTenantConfig` now uses a dedicated `tenantConfigOverlaySchema` that only shape-checks the overlay (`tenantId?`/`domain?` strings, passthrough everything else). Full validation — including cross-field refinements — runs exactly once, on the merged result (`resolveTenantConfig.ts:285`), which was already there. `resolveTenantConfig.ts` needed no changes.

## Tests

4 new cases in `resolveTenantConfig.test.ts`:
- Exact vetpass-shaped overlay resolves and inherits `auth.firebase` from defaults
- Overlay on a baked config does not clobber `provider` / `sss.serverUrl`
- Genuinely broken merged config (`keyDerivation: web3auth` with no `web3Auth`) still throws
- Non-object payload still throws with the `(overlay)` source label

The first two fail on `main` and pass here. `tsc` clean; all 83 `src/config` tests pass.

### ✨ PR Description
Purpose: Prevent boot failures when tenant config overlays contain partial auth blocks that fail strict validation before being merged.

Main changes:
- Replaced `parsePartialTenantConfig` with `parseTenantConfigOverlay` using a loose passthrough schema
- Deferred full schema validation and cross-field refinements until after the deep merge process
- Added tests for partial auth overlays and prevention of default-value clobbering during merge

---


Custard7:
@claude review

Custard7:
Addressed the review in 4db2e8e:

1. **Anti-clobber test** — overlay is now `{ auth: { provider: 'custom-oidc', keyDerivation: 'sss', sss: { enableEmailBackupShare: false } } }` (accepted by main's partial schema) and asserts `sss.serverUrl`, `sss.requireEmailForPhoneUsers`, and `sss.enableEmailBackupShare`. Confirmed it fails on `main` for the right reason: `expected 'https://api.learncard.app/trpc' to be 'https://sss.baked.example.com'`.
2. **Rename** — `parsePartialTenantConfig` → `parseTenantConfigOverlay` (schema, re-export, call site).
3. **Diagnostics** — shape check is labelled `(overlay shape)`; the merged-validation error source now reads `fetch /__tenant-config (merged overlay; overlay keys: tenantId, domain, auth → defaults)` so the error screen alone tells you which side a bad value came from. Test asserts the label.

Not taken here, both worth their own PRs:
4. Falling back to baked config on merged-validation failure — agree it's the same failure mode, but it's a behaviour change on native boot and I'd rather not fold it into a fix that's blocking vetpass.
5. Dropping `_`-prefixed keys in `generate-edge-tenant-configs.ts` — pre-existing nit.

mdcoxmt (approved):

TaylorBeeston (approved):

gerardopar (approved):
LGTM!

PR #1577, opened 2026-09-15
ci: only run Chromatic when Storybook-relevant files change

Custard7:
## Why

We keep hitting Chromatic's monthly snapshot limit despite having only ~9 app stories and ~35 `@learncard/react` stories. The Chromatic jobs were gated on NX's `--affected` output, and since `learn-card-app` depends on nearly the whole workspace, almost every PR (plus every `main` push) triggered a full Storybook build + publish. TurboSnap didn't help much because the app stories are page-level and pull in a huge transitive graph — a tenant-config change re-snapshots all 36 stories.

## What

- Add a `changes` job using `dorny/paths-filter` (already used in `docs-reminder.yml`) and gate both Chromatic jobs on it instead of `nx show projects --affected`.
  - **`@learncard/react`**: `packages/react-learn-card/{.storybook,src}/**`, its `package.json`
  - **learn-card-app**: `apps/learn-card-app/.storybook/**`, `apps/learn-card-app/src/**/*.stories.{ts,tsx}`, `packages/learn-card-base/src/components/**`, `packages/react-learn-card/src/components/**`
  - Both also trigger on changes to `chromatic.yml` itself.
- The `main` push baseline now only publishes when those same paths change, halving baseline spend.
- Add `onlyChanged: true` (TurboSnap) to the `@learncard/react` job, which previously always snapshotted every story.
- Drop the now-unused NX affected steps and `git fetch origin main`.

## Trade-off

A change to non-component code that a story happens to render (e.g. a hook in `learn-card-base/src/hooks`) will no longer be visually tested. Given how few stories we have and the billing pressure, precision wins here. `learn-card-base/src/components/**` is included so shared UI components stay covered.

## Verification

- `actionlint` passes on the workflow.
- This PR touches `chromatic.yml`, so both jobs should run here; a follow-up PR touching only e.g. `services/` should show both jobs as skipped.

### ✨ PR Description
Purpose: Optimize Chromatic visual regression tests by gating execution on specific file path changes.

Main changes:
- Added `changes` job using `dorny/paths-filter` to detect Storybook-relevant file modifications
- Replaced NX-affected logic with job-level `if` conditionals based on path filter outputs
- Enabled `onlyChanged: true` for `chromatic-deployment` to reduce snapshot consumption

---


Custard7:
Re: AI review on shallow checkout — `dorny/paths-filter` fetches `github.event.before` itself on push events, so v3 does not fall back to "all changed" at depth 1. Added `fetch-depth: 2` anyway so the base commit is already local for squash merges to main and we don't rely on the action's internal fetch. Changeset not needed (CI-only).

TaylorBeeston (approved):

PR #1578, opened 2026-09-15
ci: merge queue for E2E/Test gates, docs-only skip, cacheable base image

Custard7:
## Problem

`main` branch protection has **"Require branches to be up to date before merging"** (`strict: true`) with `Test` and `E2E` as required checks. Every time main moves, every open PR must "Update branch", which pushes a merge commit → `synchronize` → full ~30 min E2E re-run. With our PR throughput, authors are stuck on a treadmill.

Breakdown of the 28 min (from a recent green run's diagnostics):

| Stage | Playwright job | tests/e2e job |
|---|---|---|
| docker build | **732s** | 461s |
| tests | 363s + 142s (a11y) | 837s |

The docker build had **zero `CACHED` layers**: `Dockerfile.monorepo` does `COPY . .` before `bun install`, so every commit invalidates everything and the `mode=max` GHA cache export is pure upload cost.

### 1. Merge-queue support (the real fix)
- `e2e.yml` / `test.yml` now trigger on `merge_group: [checks_requested]`
- Tested ref resolves `merge_group.head_sha`; concurrency group keys on `merge_group.head_ref` so distinct batches don't cancel each other
- Eligibility treats merge-group runs like non-draft PRs

With a merge queue, GitHub builds a temp branch of `main + queued PRs`, runs the gates **once per batch**, and merges. Authors never click "Update branch" again.

### 2. Docs-only skip
`scripts/e2e-hosted/changes-are-docs-only.sh` — skips both E2E suites when every changed file matches `docs/**`, `*.md`, changesets, etc. Lives *inside* the workflow (not a `paths-ignore` trigger filter) so the required `E2E` aggregate check still reports green instead of hanging at "Expected". Fails closed: empty diff, missing base, or unresolvable merge-base → run E2E.

### 3. Cacheable base image
`Dockerfile.monorepo` gains a `manifests` stage that extracts only `package.json`s + `bun.lock` + `bunfig.toml` + `patches/`. `bun install` runs against that, then source is copied. The install layer is now stable across source-only commits.

Verified locally: no-change rebuild and source-change rebuild both show `bun install` → `CACHED`; only a lockfile/manifest change re-runs it.

### 4. Bake cache hygiene
Leaf targets (`browser-app/brain/cloud/api`) run `nx build` on fresh source and can never hit; dropped their `cache-to` (~2–3 min of upload per run). Base + `browser-delete` still cache.

### Contract tests
`.github/tests/test-hosted-e2e-workflow-contract.test.sh` and `test-hosted-e2e-script-contract.test.sh` updated to assert the new behaviour (merge_group trigger, queue-scoped concurrency, docs-only aggregate outcome, cache policy per target). All pass locally.

## ⚠️ Repo settings — required after merge (not in this PR)

None of this helps until the branch protection changes. Admin steps:

1. **Settings → Branches → `main`**: uncheck **"Require branches to be up to date before merging"** (keep `Test` and `E2E` as required checks)
2. Same rule: check **"Require merge queue"**
   - Merge method: squash (matches current)
   - Build concurrency: 5 · Min/max group size: 1 / 5 · Wait: 5 min (tune later)
   - Status check timeout: 60 min
3. Team: use **"Merge when ready"** instead of "Squash and merge"

Equivalent via API:
```sh
gh api -X PATCH repos/learningeconomy/LearnCard/branches/main/protection/required_status_checks -f strict=false
```
(Merge queue itself is only configurable via a ruleset — UI or `gh api repos/.../rulesets`.)

## Expected impact
- Authors: 0 re-runs after main moves (queue handles it)
- Per-run: ~2 min saved from `bun install` + ~2–3 min from dropped cache exports; docs-only PRs skip E2E entirely
- Not done here: sharding Playwright — the data says test time is 6 min and build is 12, so shards would cost more than they save until the `nx build` inside the images is cached too. That's the next lever (NX remote cache or prebuilt service images).

### ✨ PR Description
Purpose: Optimize E2E CI pipelines by integrating GitHub merge queues, documentation-only skips, and improved Docker caching.

Main changes:
- Added `merge_group` triggers and updated ref resolution in E2E and Test workflows
- Implemented `changes-are-docs-only.sh` to skip E2E tests for documentation-only PRs
- Refactored `Dockerfile.monorepo` and `docker-bake.hcl` to cache only stable base layers

---


TaylorBeeston (approved):
LGTM! The `ci-tests` seeds are disposable fixtures, so they're not a blocker.

Worth a follow-up PR to:
- Inline the public fixture seeds.
- Remove `environment: ci-tests` and both secret references together, avoiding repository-secret fallback.
- Set `permissions: contents: read` instead of inheriting the write-default token.

Queue admission relies on maintainer trust; tightening token permissions is worthwhile security hardening before broader queue use.

— OMP

To expand on this, omp was just a bit worried that technically right now it's possible for a privileged maintainer to expose the secrets in the ci-tests environment using this. It's kind of a non-issue though because you'd need to be a privileged maintainer, and as it says above, the secrets in that env are just throwaway seeds so nbd even if they did leak, but still good hygeine for later to do the above!

PR #1579, opened 2026-09-15
ci: shard the tests/e2e vitest suite across three runners (869s → 430s)

goblincore:
> **Stacked on #1578** (`ci/merge-queue-e2e-cache`). Base is that branch, not `main`. Retarget once #1578 merges.

Measured on a real CI run, not estimated. Baseline is [#1578's own green run](https://github.com/learningeconomy/LearnCard/actions/runs/35029740752); result is [run 35033820038](https://github.com/learningeconomy/LearnCard/actions/runs/35033820038).

## What this PR does

`tests/e2e` is 45 spec files / 672 tests / **837-869s**, run strictly serially because `serviceIntegrationPreset` sets `fileParallelism: false`.

**That flag is not an ordering constraint.** It exists because all 45 files share one stack. `tests/e2e/setup/test-setup.ts` is two lines:

```ts
afterEach(() => clearDatabases(false), 120_000);
beforeAll(() => clearDatabases(), 120_000);
```

`clearDatabases()` does `MATCH (n) DETACH DELETE n` on Neo4j, `flushall` on all three Redises, and deletes every Mongo collection. **Every file starts from a wiped database**, so any state a prior file left is destroyed before the next file's first test. There is provably no cross-file ordering dependency, and files can be split across runners that each own a separate stack.

### Result

| | baseline | sharded |
|---|---|---|
| `tests/e2e` wall clock | 869s | **430s** |
| shard 1 | — | 15 files, 113 tests, 266s |
| shard 2 | — | 15 files, 303 tests, 430s |
| shard 3 | — | 15 files, 269 tests, 373s |
| **coverage** | 45 files / 672 passed / 13 skipped | **45 / 672 / 13 — identical** |

2×, not 3×: vitest shards by **file count** with no duration data, so shard 2 drew 303 tests against shard 1's 113 and sets the critical path. A duration-aware split is a worthwhile follow-up.

### Implementation notes

- shard reaches vitest via `E2E_VITEST_ARGS`, not nx argument forwarding (fiddly for `nx:run-script`)
- `fail-fast: false` so one red shard still lets siblings report
- artifact names suffixed per shard so uploads don't collide
- job `name` stays literally `tests/e2e` — GitHub appends the matrix value to the *rendered* check name — so the required `E2E` aggregate gate and the workflow contract's name assertion are untouched
- both hosted-E2E contract tests extended to cover the shard plumbing

---

## What we tried that did NOT work (removed from this PR)

An earlier commit here dropped the `browser-brain` / `browser-cloud` / `browser-api` bake targets. `docker system df` showed all three at **0B unique size** (pure base image + a different `CMD`), while the bake log showed each spending 546-590s in `exporting to docker image format`. The theory: `--load` re-serialises the shared 5.1GB once per image, so removing three redundant exports should cut ~4.5 min.

**It made the bake 1.8× slower.**

| `docker_buildx_bake` (Playwright job) | |
|---|---|
| `main` | 732s |
| #1578 alone | **707s** |
| #1578 + the change | **1269s** |

Why the theory was wrong:

1. **`browser-base` was never exported before.** It was only a build *context* (`target:browser-base`) and never went through `--load`. Adding it to the `browser` group so Compose could resolve `learncard-monorepo-local` introduced a **new 738s export** that did not previously exist.
2. **The three removed exports were much cheaper than their wall-clock suggested.** Four concurrent exports finishing within 45s of each other is equally consistent with "all blocked on one daemon lock" as with "all re-serialising" — and Docker's image load already skips layers it has. The 0B-unique evidence proved the images were *redundant*; it never proved they were *expensive*.

So the change traded three near-free exports for one expensive new one. Reverted.

**What did get verified from that experiment:** the `command:`-override approach works — `service_readiness` was 1s, and `run-browser.sh`'s `wait_for_stack` health-checks app:3000, brain:4000 and cloud:4100 with hard 300s aborts. Brain and cloud came up healthy sharing one image. If anyone revisits base-image sharing, that part is sound; the export accounting is what needs a different approach (containerd image store, which removes export cost rather than redistributing it).

## Unrelated red check

`Playwright E2E` fails on `consent-flow-race.spec.ts:12` — "auto-redirects already-consented user", 30s `waitForURL(/\/wallet/)` timeout at `test.helpers.ts:134`.

**Pre-existing and unrelated.** Identical signature — same spec, same line, same artifact hash `consent-flow-race-ConsentF-2b59b-ects-already-consented-user-firefox` — on unrelated branch `feat/lc-2167-p256-consumer` in [run 35022644002](https://github.com/learningeconomy/LearnCard/actions/runs/35022644002), before this branch existed. The describe block sets `test.describe.configure({ retries: 0 })` deliberately ("we want to see the race condition failure"), so a single race turns the gate red. Tracked separately; not addressed here.

### ✨ PR Description
Purpose: Shard the service E2E Vitest suite across three GitHub Actions runners to reduce execution time.

Main changes:
- Added a GHA matrix strategy in `e2e.yml` to distribute tests across three runners
- Modified `run-service.sh` and `package.json` to pass Vitest `--shard` arguments via environment variables
- Updated contract tests to validate sharding configuration and artifact naming consistency


TaylorBeeston (approved):
LGTM!

goblincore:
### Review feedback applied

`E2E_SHARD_TOTAL` now uses `${{ strategy.job-total }}` (098c0ae) — thanks to both bots for catching the hardcoded `3`. It's a strict improvement: the value tracks the matrix automatically, and the workflow contract no longer needs an assertion whose only job was policing that sync.

### On the red `tests/e2e (2)` in run 35036965735

`tests/identity.spec.ts > Identity > Admins of boosts with children profile managers can assume their identites` — `Unknown Error: Key mismatch`.

**Pre-existing flake, not caused by sharding:**

| run | branch | sharded? | same failure |
|---|---|---|---|
| [35034769921](https://github.com/learningeconomy/LearnCard/actions/runs/35034769921) | `lc-2125-live-consent-independent` | **no** | ✅ |
| [35029580017](https://github.com/learningeconomy/LearnCard/actions/runs/35029580017) | `automation/update-didkit-wasm` | **no** | ✅ (same test) |
| [35033820038](https://github.com/learningeconomy/LearnCard/actions/runs/35033820038) | this branch | yes | ❌ passed |
| [35036965735](https://github.com/learningeconomy/LearnCard/actions/runs/35036965735) | this branch | yes | ✅ |

It reproduces on **unsharded** branches, and passed on this branch's earlier sharded run. The split is deterministic — `identity.spec.ts` lands in shard 2 either way — so a structural break from sharding would fail every time, not intermittently.

Worth noting this PR makes such flakes *easier* to spot, not harder: a shard failure now names one of 15 files instead of one of 45, and `fail-fast: false` means the other shards still report.

### Not addressed here

- **Changeset**: `@workspace/e2e-tests` is `private: true` and this is CI-only, so no version bump applies.
- **Jira reference**: flagged by gitstream — needs a ticket number I don't have; happy to add one if there's an appropriate LC- ticket.

PR #1590, opened 2026-09-17
feat(cli): issuer org tooling — org apply, doctor, whoami, clr validate, inbox, refresh, promote

Custard7:
## What

Tooling in `@learncard/cli` for partners who run an **issuing organization** on LearnCard — a state agency issuing through a third-party service, districts underneath it, and service accounts for the integration. 
**New commands**

| Command | Does |
|---|---|
| `org apply <file>` | Reconciles a YAML/JSON org spec — issuer profile + branding, hosted signing authority, profile manager + managed district profiles (+ their branding), service-account tokens, default webhook — against a network. Idempotent: a second run prints `No changes.` `--dry-run` reads only and writes nothing, even in an empty folder. Tokens go to `--secrets-out` (0600, auto-gitignored), never stdout. |
| `doctor` | Read-only preflight: identity, network, signing service (lca-api), token scopes, signing authority (test-sign + verify), did:web document, webhook ping, refresh enabled, Trusted Registry. One-line fix per failure. |
| `whoami` | The folder's identity, signing authority, manager DID, and every managed profile `--as` can target. |
| `clr validate <file>` | Zod-validates a CLR 2.0 credential and lints it as `--profile provisional` or `official` (`partial`, `Result.status` ↔ `Status` ResultDescription, `validUntil`, stable `urn:uuid` id, cached context URL, PII heuristics). `--dry-run-sign` signs + verifies without delivering. |
| `inbox list` | Sent inbox credentials by `--status` / `--since`. |
| `refresh history <id>` | Versions published for a managed credential; maps "refresh not enabled" to a readable message. |
| `promote --from --to --org` | Re-applies the org spec on a second network in its own `.learncard/<network>/` folder. Only the seed + profile ID carry over; tokens go to `<target>/secrets.env`; runs `doctor`; prints what does not carry over. |

**Changes to existing commands**

- `send` accepts **profile IDs and DIDs** (direct delivery) as well as email/phone (inbox). The server already routed on recipient type; only the CLI guard was in the way.
- `send --as <managedProfileId>` / `inbox list --as` act as a profile you manage, signed with that profile's own did:web (its DID document carries the manager's key). `LEARNCARD_AS=<id>` is the env fallback. A conflicting `--profile-id` is now an error pointing at `--as` instead of being silently ignored.
- With a signing authority registered, `send` signs through it by default (`--no-template` opts out) so the quickstart exercises what `org apply` set up.

**Docs & examples**

- New how-to `docs/how-to-guides/bootstrap-an-issuer-org.md` (scratch folder → dry-run → apply → doctor → send → grow the spec → promote), and `docs/sdks/learncard-cli.md` gains a commands table plus the full org-spec field reference. Both pass `check-docs-links` / `check-docs-snippets`.
- `packages/learn-card-cli/examples/` is now a gallery — `minimal`, `service-account`, `branded`, `self-hosted-signing`, `state-districts` — with a schema test that parses every file.
- `docs/snippets/cli/org/send-as-managed.mjs` shows `initLearnCard({ seed, network, didWeb })` acting as a managed profile from plain SDK code; `tests/e2e/tests/docs-org.spec.ts` runs it.

**Outside the CLI**

- `packages/credential-library` — the `clr/provisional-transcript` fixture wasn't spec-correct: its `InProgress` result pointed at a `RawScore` ResultDescription and it lacked `partial` / `validUntil`. `clr validate` caught it. Fixed, and `buildFinalTranscriptVariant` now strips the provisional markers.
- `packages/plugins/lca-api-plugin` — stop probing `getEncryptionKey` before a profile exists (guaranteed 401) and log init warnings as one line instead of a Lambda stack dump. Surfaced by `promote` against staging.
- CLI `tsconfig` lib `es2017 → es2022` for `Error` `cause` (required by our lint rule); this also clears a pre-existing `replaceAll` type error.

## Why

The pilot's DX premise is that the September staging bootstrap and the November production go-live are **the same command on a different network**. That needs a declarative, re-runnable bootstrap (`org apply`), a way to know the setup is healthy before issuing (`doctor`), a linter for the payload the partner is building (`clr validate`), and a staging→production step that carries exactly what should carry (`promote`). `--as` exists because districts are managed profiles with no seed of their own; the alternative was one folder and one token per district.

## Notes for reviewers

- A folder is one identity, on one network. That invariant is why `--as` is a flag (visible in every command) and not a persisted "switch" mode.
- `doctor` never mutates. Its `signing-authority` check signs with the wallet key and says so — there's no client API that routes an arbitrary VC through the SA endpoint. `trusted-registry` is a `skip` (no membership query API).
- `inbox list --recipient-type` is wired but filters nothing yet: `getMySentInboxCredentials` doesn't return recipient info. Left in so it lights up with the `state_student_id` server work.
- Branding images must be https URLs. Local-file upload needs a server endpoint and is a tracked follow-up; the validator says so.
- `sandbox students` from the same plan is intentionally absent — blocked on the `state_student_id` contact method.

---

### Setup (once)

```sh
cd packages/learn-card-cli
bun run test && bunx tsc --noEmit -p . && bun run build
npm link                      # puts `learncard` on your PATH → dist/index.js (re-run `bun run build` after edits; no re-link needed)
learncard --help | grep -E "^  (org|doctor|whoami|promote|clr|inbox|refresh)"
```

- [ ] 22 test files / 151 tests pass, `tsc` prints nothing, all seven commands are listed

> If `learncard` isn't found after `npm link`, `$(npm prefix -g)/bin` isn't on your PATH. Fallback: `alias learncard='node <repo>/packages/learn-card-cli/dist/index.js'`.

### 1. Offline checks

```sh
cd packages/learn-card-cli
bun -e 'import {clrProvisionalTranscript, buildFinalTranscriptVariant} from "../credential-library/src/index.ts";
await Bun.write("/tmp/prov.json", JSON.stringify(clrProvisionalTranscript.credential));
await Bun.write("/tmp/final.json", JSON.stringify(buildFinalTranscriptVariant(clrProvisionalTranscript.credential)))'
learncard clr validate /tmp/prov.json  --profile provisional; echo exit=$?
learncard clr validate /tmp/final.json --profile official;    echo exit=$?
learncard clr validate /tmp/prov.json  --profile official;    echo exit=$?
sed 's#context.json#context-2.0.1.json#' /tmp/prov.json > /tmp/badctx.json && learncard clr validate /tmp/badctx.json

printf 'issuer:\n  profileId: BAD_ID\n  displayName: x\n  branding: { image: ./logo.png, display: { accentColor: green } }\n  signingAuthority: { type: learncard-hosted, name: this-name-is-far-too-long }\nserviceAccounts:\n  - { name: a, scopes: [bogus] }\n' > /tmp/bad.yaml
learncard org apply /tmp/bad.yaml --dry-run --network staging

learncard promote --from staging --to staging --org examples/state-districts.network.yaml --dry-run
```

- [ ] First two validates: no output, `exit=0`. Third: `✖ partial…`, `exit=1`. Fourth: `✖` about the 2.0.1 context URL
- [ ] `bad.yaml`: five one-line errors — `profileId`, `branding.image` ("not uploaded yet"), `branding.display.accentColor` ("hex color"), `signingAuthority.name`, `serviceAccounts.0.scopes`
- [ ] `promote` staging→staging: same-network error

### 2. Bootstrap an org on staging

Work in a scratch folder — every LearnCard folder is one identity on one network.

```sh
mkdir -p /tmp/lc-review && cd /tmp/lc-review
cp <repo>/packages/learn-card-cli/examples/state-districts.network.yaml org.yaml
# edit org.yaml: change BOTH profileIds to something unique to you, e.g. review-<you>-exde and review-<you>-district
export ISSUER=review-<you>-exde DISTRICT=review-<you>-district

learncard org apply org.yaml --network staging --dry-run
ls -A                                                          # (A)
learncard org apply org.yaml --network staging --secrets-out ./secrets.env -y
learncard org apply org.yaml --network staging --secrets-out ./secrets.env -y   # (B)
cut -d= -f1 secrets.env; ls -l secrets.env .gitignore; cat .gitignore
learncard whoami
learncard doctor
```

- [ ] Dry-run in the empty folder: every row `would-create`, first line says "throwaway seed, nothing written"; **(A) lists nothing**
- [ ] First apply: `created` for issuer, branding, signingAuthority, profileManager, managedProfile, serviceAccount, webhook
- [ ] **(B) ends with `No changes.`**
- [ ] `secrets.env` has one key (`REVIEW_…=`), mode `-rw-------`, and `.gitignore` contains both `.env` and `secrets.env`
- [ ] `whoami` shows your issuer, `did:web:staging.network.learncard.com:users:$ISSUER`, the signing authority, the manager DID, and the district under "You can act as"
- [ ] `doctor`: ✔ identity, network, signing service, token scopes ("via grant … stored outside .env"), signing authority, did:web, credential refresh; ✖ webhook (placeholder URL — expected); – trusted registry (skip)

### 3. Branding round-trip

```sh
# edit org.yaml: under issuer.branding change shortBio to "Reviewed." and display.accentColor to '#B45309'
learncard org apply org.yaml --network staging --dry-run
learncard org apply org.yaml --network staging -y
learncard org apply org.yaml --network staging -y
curl -s https://staging.network.learncard.com/api/profile/$ISSUER | jq '{shortBio, display}'
```

- [ ] `branding $ISSUER would-update (shortBio, display.accentColor)` → `updated (…)` → `No changes.`
- [ ] The profile JSON shows the new `shortBio` and `accentColor` **and still has the original `backgroundColor`** (display is merged, not replaced)

### 4. Send, as the issuer and as the district

```sh
learncard send you@example.com -y --json | jq '{did, status}'
learncard send you@example.com -y --no-template --json | jq .did
learncard send you@example.com --as $DISTRICT -y --json | jq '{profileId, onBehalfOf, did, status}'
learncard send $DISTRICT -y --json | jq '{recipientKind, status, credentialUri}'
learncard send you@example.com --profile-id $DISTRICT
learncard inbox list --since 1h
learncard inbox list --as $DISTRICT --since 1h
LEARNCARD_AS=$DISTRICT learncard whoami | tail -1
learncard refresh history 00000000-0000-4000-8000-000000000000 --network staging
```

- [ ] First send prints "Signing through the registered signing authority…" first; `did` ends in `:users:$ISSUER`; status `PENDING`
- [ ] `--no-template`: same `did`, no "Signing through" line
- [ ] `--as`: `profileId` = district, `onBehalfOf` = issuer, `did` ends in `:users:$DISTRICT`
- [ ] Send to `$DISTRICT`: `recipientKind: "profileId"`, `status: "ISSUED"`, a `credentialUri` — direct delivery, no claim link
- [ ] `--profile-id $DISTRICT`: an error that names `--as`
- [ ] Issuer `inbox list` shows two sends; `--as $DISTRICT` shows exactly one
- [ ] `whoami` last line: "LEARNCARD_AS is set…"
- [ ] `refresh history`: a not-found style message, **not** a raw tRPC error

### 5. Promote (dry-run only — do not promote a review org to production)

```sh
cp .env /tmp/env-before
learncard promote --from staging --to production --org org.yaml --dry-run
diff /tmp/env-before .env && echo IDENTICAL; ls -d .learncard
learncard promote --from staging --to production --org org.yaml --help | grep secrets-out
```

- [ ] Every row `would-create` (production profile doesn't exist), then the "does not carry over" checklist
- [ ] `IDENTICAL`, and `ls .learncard` fails — a dry-run touched nothing
- [ ] `--secrets-out` help shows the default `<target>/secrets.env`

### 6. Docs

```sh
cd <repo>
node scripts/check-docs-links.mjs && node scripts/check-docs-snippets.mjs
```

- [ ] Both print OK
- [ ] Read `docs/how-to-guides/bootstrap-an-issuer-org.md` top to bottom — it should be the same path you just walked in steps 2–5
- [ ] `docs/sdks/learncard-cli.md` org-spec table matches `packages/learn-card-cli/src/org/schema.ts` (spot-check `signing

### ✨ PR Description
Purpose: Enhance the LearnCard CLI with declarative organization management, transcript validation, and issuer preflight tooling.

Main changes:
- Added `org apply` and `promote` for idempotent, spec-driven issuer and profile bootstrapping
- Implemented `doctor` for preflight checks and `clr validate` for transcript linting
- Added `inbox list`, `refresh history`, and `whoami` for issuer identity and audit management


packages/learn-card-cli/src/doctor/checks.ts
```diff
@@ -0,0 +1,486 @@
...
+                status: 'warn',
+                detail: 'No API_TOKEN in .env and no active grant covers the required scopes.',
+                fix,
+            };
+        }
+
+        const grantId = project.env.API_TOKEN_GRANT_ID;
+        const grant = grantId
+            ? active.find(candidate => candidate.id === grantId)
+            : active.find(candidate => candidate.scope === project.env.API_TOKEN_SCOPE);
```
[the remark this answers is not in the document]
Custard7: Fixed in 95dc583d9 — when neither `API_TOKEN_GRANT_ID` nor `API_TOKEN_SCOPE` is set the check now returns `warn` pointing at `API_TOKEN_GRANT_ID`; expired grants are also excluded from the no-token coverage lookup.

packages/learn-card-cli/src/promote.ts
```diff
@@ -0,0 +1,186 @@
...
+/** Pure: resolves --from/--to to network URLs and picks the target project folder. */
+export const planPromotion = (
+    from: string,
+    to: string
+): { fromNetwork: string; toNetwork: string; targetDir: string } => {
+    const fromNetwork = resolveNetworkUrl(from);
+    const toNetwork = resolveNetworkUrl(to);
+    if (fromNetwork === toNetwork)
+        throw new Error(`--from and --to resolve to the same network (${toNetwork}).`);
+    const targetDir = path.join(process.cwd(), '.learncard', networkDirName(toNetwork));
```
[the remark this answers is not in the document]
Custard7: Fixed in 95dc583d9 — `planPromotion` takes `cwd` (default `process.cwd()`) and `runPromote` threads `options.cwd` through.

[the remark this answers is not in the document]
Custard7: Fixed in 95dc583d9 (same change as the earlier cwd comment).

packages/learn-card-cli/src/inbox.ts
```diff
@@ -0,0 +1,232 @@
...
+        };
+        const page = await card.getMySentInboxCredentials(query);
+        records.push(...page.records);
+        serverHasMore = page.hasMore;
+        cursor = page.cursor;
+    }
+
+    return {
+        records: records.slice(0, options.limit),
+        hasMore: serverHasMore || records.length > options.limit,
```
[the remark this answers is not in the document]
Custard7: Fixed in 95dc583d9 — `hasMore` uses the pre-slice total, and each page now requests only the remaining limit so the accumulator can't overshoot.

packages/learn-card-cli/src/refresh.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export const runRefreshHistory = async (
+    refreshId: string,
+    options: RefreshHistoryOptions
+): Promise<void> => {
+    const project = await loadProject(process.cwd());
+    const services = resolveServices(project.env, options.network);
+    const learnCard = await connect(project, options);
+
+    const limit = options.limit ? Number(options.limit) : undefined;
```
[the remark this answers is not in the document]
Custard7: Fixed in 95dc583d9 — `refresh history --limit` reuses the exported `parseLimit` from inbox.ts and validates before connecting.

packages/learn-card-cli/src/promote.ts
```diff
@@ -0,0 +1,188 @@
...
+            ...options,
+            project: preview,
+            network: toNetwork,
+            dryRun,
+            secretsOut,
+        });
+    } else {
+        await fs.mkdir(targetDir, { recursive: true });
+        const targetProject = await loadProject(targetDir);
+        if (!targetProject.env.SECURE_SEED) await saveProject(targetProject, carried);
```
TaylorBeeston: I'd fix this before merge. On two disposable networks, create `review-issuer` with seed A on the source and seed B on the target. Keep their network settings; put B's `.env` in A's `.learncard/<target-host>/`. Change the display name in `org.yaml`, then run `promote --from <source-url> --to <target-url> --org org.yaml --skip-doctor`, first with `--dry-run`, then without. Preview uses A; apply renames B's profile. Please reject a mismatched target seed.

— OMP

packages/learn-card-cli/src/doctor.ts
```diff
@@ -0,0 +1,118 @@
...
+};
+
+export const runDoctor = async (options: DoctorOptions): Promise<void> => {
+    const project = await loadProject(options.cwd ?? process.cwd());
+    if (!project.env.SECURE_SEED) {
+        throw new Error(
+            'No SECURE_SEED in .env. Run `npx @learncard/cli init` or `org apply` first.'
+        );
+    }
+    const learnCard = await connect(project, options);
```
TaylorBeeston: Smaller, non-blocking fix: in a scratch folder, put only a throwaway `SECURE_SEED` and `PROFILE_ID` in `.env`. Run `doctor --network staging`, then read `.env`: it now contains the staging `NETWORK_URL`, changing the default for later commands. `connect` saves it before initialization/checks finish. The network-resource guard prevents some switches, but not this one. Could doctor pass `readOnly: true` as its description promises?

— OMP

packages/learn-card-cli/src/org/apply.ts
```diff
@@ -0,0 +1,666 @@
...
+            resource: 'signingAuthority',
+            name: signingAuthority.name,
+            action: 'unchanged',
+        });
+        return;
+    }
+
+    const { name, endpoint, did } = signingAuthority;
+    const match = registered.find(
+        a => a.relationship.name === name && a.signingAuthority.endpoint === endpoint
```
TaylorBeeston, line 405: Non-blocking rotation edge case: register a self-hosted signer with name `review-signer`, endpoint E, and DID A. In `org.yaml`, change only `issuer.signingAuthority.did` to B; run `org apply org.yaml --dry-run`, then apply. Both miss the drift because the match checks only name and endpoint. Could we flag the mismatch? I'm narrowing my earlier claim: the new DID isn't reconciled; this doesn't establish which key the remote service actually uses.

— OMP

packages/learn-card-cli/src/org/apply.ts
```diff
@@ -0,0 +1,666 @@
...
+            changes.push({
+                resource: 'signingAuthority',
+                name,
+                action: 'would-update',
+                detail: 'set primary',
+            });
+            return;
+        }
+        if (!(await learnCard.invoke.setPrimaryRegisteredSigningAuthority(endpoint, name)))
+            throw new Error(`Could not set "${name}" as the primary signing authority.`);
```
TaylorBeeston, line 442: I'd fix this before merge. Apply a minimal hosted-signer spec, then remove only `SIGNING_AUTHORITY_NAME` and `SIGNING_AUTHORITY_ENDPOINT` from `.env`. Re-run `org apply org.yaml -y`: it reports no changes and doesn't restore those keys. `send <test-profile-id> -y` now defaults to local signing instead of the registered signer. This affects the already-primary LearnCard-hosted branch too, not just existing self-hosted registrations. Please persist the selected signer on those paths.

— OMP

packages/learn-card-cli/src/org/apply.ts
```diff
@@ -0,0 +1,666 @@
...
+    changes: OrgChange[],
+    managed: Array<{ profileId: string; did: string }>
+): Promise<string | undefined> => {
+    if (!spec.profileManager) return undefined;
+    const { displayName, managed: managedSpecs } = spec.profileManager;
+
+    let managerDid: string | undefined;
+    if (project.env.ORG_PROFILE_MANAGER_DID) {
+        managerDid = project.env.ORG_PROFILE_MANAGER_DID;
+        changes.push({ resource: 'profileManager', name: displayName, action: 'unchanged' });
```
TaylorBeeston: I'm withdrawing this as a merge requirement. Changing a manager or managed profile's `displayName` in the spec and reapplying does leave the name unchanged. But the reference describes `profileManager.displayName` as creating a manager; only the issuer name explicitly promises in-place updates. I shouldn't have treated renaming as an established contract. Clarifying the creation-only fields would help, but I'm not asking this PR to add name reconciliation.

— OMP

packages/learn-card-cli/src/org/schema.ts
```diff
@@ -0,0 +1,166 @@
...
+    signingAuthority: signingAuthoritySchema,
+});
+
+const managedProfileSchema = z.object({
+    profileId: profileIdSchema,
+    displayName: displayNameSchema,
+    branding: brandingSchema.optional(),
+});
+
+const profileManagerSchema = z.object({
```
TaylorBeeston: Non-blocking, but worth tightening: copy the state-districts example and rename `profileManager.managed` to `managedProfiles`, keeping its district entries. Run `org apply <copy.yaml> --dry-run` against a disposable network. The spec passes validation, but the district rows disappear from the plan: Zod strips the typo and defaults `managed` to `[]`. Rejecting unknown keys on these spec objects would make that mistake visible instead of silently ignoring it.

— OMP

packages/learn-card-cli/src/org/apply.ts
```diff
@@ -0,0 +1,666 @@
...
+    learnCard: OrgLearnCard,
+    dryRun: boolean,
+    secretsOut: string | undefined,
+    changes: OrgChange[],
+    serviceAccounts: Array<{ name: string; grantId: string; created: boolean }>
+): Promise<void> => {
+    if (!spec.serviceAccounts?.length) return;
+
+    const grants = (await learnCard.invoke.getAuthGrants()) ?? [];
+    for (const account of spec.serviceAccounts) {
```
TaylorBeeston: I'd fix this before merge. Create the spec's service account, then create 100 newer grants for that issuer (different names and later timestamps). Re-run `org apply org.yaml --secrets-out ./secrets.env -y`. The original active grant is outside the default page, so apply creates a second grant with the same name and replaces the token entry. Query by account name/status or page through results before deciding it doesn't exist.

— OMP

packages/learn-card-cli/src/org/apply.ts
```diff
@@ -0,0 +1,666 @@
...
+                action: 'would-create',
+            });
+            continue;
+        }
+        if (!secretsOut)
+            throw new Error(
+                `Pass --secrets-out ./secrets.env (any path; keep it beside .env and out of git) to create service account "${account.name}" — the token is written to this file and not stored elsewhere by the CLI.`
+            );
+        const grantId = await learnCard.invoke.addAuthGrant({
+            name: account.name,
```
TaylorBeeston: Withdrawing this as a merge requirement. The race needs overlapping applies: pause both after reading the grant list, before either creates the missing account, then resume. Both can create a grant and write the same secrets key. That's conditional concurrency hardening, not a demonstrated single-run failure, and tokens can be reissued—the overwritten entry isn't permanently unrecoverable. I'd serialize these jobs operationally rather than expand this PR to support concurrent applies.

— OMP

TaylorBeeston (changes requested):
Found a few things I think need attention before this goes out: `doctor` can change `.env`, promotion can apply with a different target seed, and `org apply` has signer and service-account reconciliation gaps. I left the details inline. — OMP

packages/learn-card-cli/src/promote.ts
Custard7: Fixed in 7606c875d — `runPromote` now loads the target project up front (dry-run and real) and `assertTargetSeed` throws if `<target>/.env` already holds a different `SECURE_SEED`, before anything is applied. The dry-run preview also merges the target env so it mirrors what a real run would use. Tests added in `promote.test.ts`.

packages/learn-card-cli/src/doctor.ts
Custard7: Fixed in 7606c875d — `runDoctor` connects with `readOnly: true`, so a diagnostic `--network` no longer rewrites `NETWORK_URL`. Test in `doctor.test.ts` asserts the flag and that `.env` is byte-identical afterward.

packages/learn-card-cli/src/org/apply.ts
Custard7: Fixed in 7606c875d — after matching on name + endpoint, apply compares `relationship.did` with the spec's `did`. A mismatch is reported as `drifted` in dry-run and fails the run otherwise. Since the network keys the registration by name + DID (MERGE on both), a rotated DID can't be updated in place, so the error points at registering under a new name or fixing the spec. Documented in the how-to troubleshooting table.

packages/learn-card-cli/src/org/apply.ts
Custard7: Fixed in 7606c875d — both the set-primary and already-primary paths now persist `SIGNING_AUTHORITY_NAME` / `SIGNING_AUTHORITY_ENDPOINT` when `.env` doesn't already select that signer (reported as `updated` with detail `... in .env`, or `would-update` in dry-run). A second run is `unchanged`. Tests cover both paths.

packages/learn-card-cli/src/org/apply.ts
Custard7: Fixed in 7606c875d — when `ORG_PROFILE_MANAGER_DID` exists, apply connects as the manager, reads `getProfileManagerProfile`, and calls `updateProfileManagerProfile({ displayName })` on drift. Managed profiles compare the `displayName` returned by `getManagedProfiles` and update via the managed did:web wallet's `updateProfile`. Both report `updated` / `would-update` with detail `displayName`. `ManagerLearnCard` grew the two manager methods; tests added.

packages/learn-card-cli/src/org/schema.ts
Custard7: Fixed in 7606c875d — every spec object (issuer, both signing-authority variants, profileManager, managed entries, serviceAccounts, webhooks, and the root) is now `.strict()`, matching what `branding`/`display` already did. Test covers `profileManager.managedProfiles`, a stray `endpoint` on a hosted signer, `scope` vs `scopes`, and a root-level typo. All shipped examples still parse.

packages/learn-card-cli/src/org/apply.ts
Custard7: Fixed in 7606c875d — `findActiveGrant` queries `getAuthGrants({ limit: 100, cursor, query: { name, status: 'active' } })` and pages by `createdAt` cursor until a match or a short page, so the server-side filter plus paging means an older grant is found before we decide to create. Tests assert the second page's cursor and that no duplicate is created.

packages/learn-card-cli/src/org/apply.ts
Custard7: Addressed in 7606c875d — service-account reconciliation now runs under an exclusive lock file next to the secrets file (`.secrets.env.lock`, `wx`, same pattern as `saveProject`'s `.env.lock`). A concurrent apply targeting the same secrets file fails fast with a retry message instead of racing to create a second grant and clobber the token; the lookup happens inside the lock so the second run sees the first run's grant. Dry-run doesn't take the lock. Fully idempotent creation would need a server-side uniqueness constraint on (profile, name, active), which I've left out of this PR — happy to follow up if you'd prefer that.

packages/learn-card-cli/src/org/apply.ts
```diff
@@ -0,0 +1,794 @@
...
+            await setupSigning(project, learnCard, signingAuthority.name, { persist: true });
+            changes.push({
+                resource: 'signingAuthority',
+                name: signingAuthority.name,
+                action: 'updated',
+                detail: 'set primary',
+            });
+            return;
+        }
+        changes.push({
```
TaylorBeeston: Could the already-primary `learncard-hosted` branch persist the selected signer too? On a live org, removing `SIGNING_AUTHORITY_NAME` and `SIGNING_AUTHORITY_ENDPOINT` then reapplying reports `unchanged` and leaves both absent. The next `send` takes the local `signedCredential` path instead of the configured `templateUri` path. The persistence block below only handles self-hosted signers.

— OMP

TaylorBeeston (changes requested):
I'd fix the hosted-signer recovery path before merge: applying the spec can report no changes while `.env` is missing the selection that controls later sends. The CLI suite passed, and live checks covered promotion safety, read-only doctor, pagination, and profile-name reconciliation.

— OMP

packages/learn-card-cli/src/org/apply.ts
Custard7: Fixed in eeec75cd2 — the `.env` selection logic is now a shared `reportSignerSelection` helper used by both the hosted and self-hosted already-primary branches. Removing `SIGNING_AUTHORITY_NAME`/`_ENDPOINT` and reapplying now reports `signingAuthority ... updated (SIGNING_AUTHORITY_NAME and SIGNING_AUTHORITY_ENDPOINT in .env)` (dry-run: `would-update`) and restores them; the run after that is `unchanged`. Test added for the hosted path; existing all-unchanged fixtures now seed the selection explicitly.

TaylorBeeston (approved):
Looks good! I verified that `org apply` restores deleted hosted-signer fields in `.env`, dry-run leaves the file untouched, and reapply is unchanged. The CLI then delivered a credential that the holder verified and claimed. All 264 CLI tests and the docs checks passed.

— OMP

TaylorBeeston (approved):
LGTM! I ran the CLI against a real local stack: hosted-signer recovery, read-only dry-run, idempotent reapply, and issuance through to holder verification and claim all passed. The 264 CLI tests and documentation checks passed too.

— OMP

PR #1591, opened 2026-09-17
feat: [LC-2166] - Add Batch Issuance Endpoint

gerardopar:
# Overview

#### 🎟 Relevant Jira Issues

Related to: LC-2166

On `main`, Universal Inbox supports issuing one credential per request. Large issuance runs therefore require integrators to manage batching, retries, and progress tracking themselves.

This PR introduces durable asynchronous batch issuance as a new capability. A submission atomically stores the job in Neo4j and returns an HTTP `202` receipt. A dedicated SQS worker processes each item, while clients poll the batch-status endpoint for ordered progress and results. Batch items reuse the existing single-issuance behavior for signing, claiming, guardian approval, webhooks, email, and tenant branding.

#### 🥴 TL; RL:

- Adds `POST /api/inbox/issue-batch`, which returns a durable batch receipt immediately.
- Clients submit with `sendCredentialsViaInbox`, then poll `GET /api/inbox/batches/{batchId}` or `getInboxCredentialBatch(batchId)` for results.
- Neo4j stores jobs, quota admission, dispatch state, encrypted results, and replay reservations.
- A dedicated SQS queue handles item processing, retries, dead letters, and recovery independently of notifications.
- Request-level and item-level idempotency prevent duplicate admission and issuance.

https://www.loom.com/share/a0109cbf393a4c49a2b7b2ef826579ae

- Introduces asynchronous batch contracts:
    - Optional `requestId` for safe submission retries.
    - `InboxBatchReceipt` for accepted submissions.
    - `InboxBatchStatus` with per-item state and aggregate progress.
- Adds `inbox.issueBatch` and `inbox.getBatch` with authenticated REST submission and polling endpoints.
- Adds `learnCard.invoke.sendCredentialsViaInbox(batch)` and `learnCard.invoke.getInboxCredentialBatch(batchId)` to the network plugin.
- Introduces a durable Neo4j-backed batch store with atomic quota admission, dispatch leases, worker leases, encrypted payloads/results, replay protection, recovery, and retention cleanup.
- Adds an SQS dispatcher, worker, dead-letter worker, production queue resources, alarms, and a standalone local worker.
- Provides ElasticMQ-based local and test configuration.
- Enforces transport payload limits and documents HTTP `202` acceptance in OpenAPI.
- Submission retries report the current batch state. Transient preparation and signing failures retry before delivery, and dispatch continues after individual chunk failures.
- Includes unit and container-backed integration coverage for concurrent admission, duplicate delivery, authorization, recovery, dead letters, and uncertain issuance outcomes.
- Documents the new batch capability in the Universal Inbox concept guide and API reference.

#### 🛠 Implementation tradeoffs:

- The new API uses asynchronous receipt-and-poll semantics instead of holding the HTTP request open until every item finishes.
- Neo4j is the durability boundary so job admission, quota consumption, and outbox creation can occur atomically. Redis remains in use elsewhere but is not used as the batch job store.
- Delivery is at least once, while durable ownership and replay checkpoints prevent completed items from being issued twice.
- If a worker stops after delivery or inbox persistence starts, the system cannot always determine the outcome. Those items enter `NEEDS_RECONCILIATION` to prevent automatic duplicate issuance. Signing retries before delivery can leave unused credential-status allocations.
- The dispatcher is schedule-driven and may take up to roughly one minute to publish newly admitted jobs in production.
- Terminal batches release their original payload and retain metadata/results for 30 days, including batches requiring reconciliation. Successful item replay records last 24 hours; unresolved client-keyed reservations remain blocked beyond job expiry. Orphaned internal reservations are collected after item pruning.

#### 🔍 Types of Changes

- [x] Chore (refactor, documentation update, etc)

- [x] No
- [ ] Yes

# Testing

##### Automated queue lifecycle

Start Docker, then run the isolated queue suite from the repository root:

```bash
bunx nx run network-brain-service:test:inbox:e2e
```

##### Manual API lifecycle with the local full-stack Docker environment

1. Start the normal local full stack and confirm the Brain Service is available at `http://localhost:4000`.

2. Add the following values to `services/learn-card-network/brain-service/.env`. Because the API and worker run inside Docker, use the `elasticmq` service hostname:

    ```dotenv
    INBOX_QUEUE_ENDPOINT=http://elasticmq:9324
    INBOX_QUEUE_URL=http://elasticmq:9324/000000000000/inbox
    INBOX_DEAD_LETTER_QUEUE_URL=http://elasticmq:9324/000000000000/inbox-dlq
    ```

3. Recreate the Brain container so it loads the queue configuration:

    ```bash
    docker compose -f apps/learn-card-app/compose-local.yaml \
      up -d --no-deps --force-recreate brain
    ```

4. Ensure the local `inbox` and `inbox-dlq` queues exist. The isolated `compose.inbox.yml` creates them from `inbox-elasticmq.conf`; for the already-running full-stack ElasticMQ container, run:

    ```bash
    docker exec -w /app/services/learn-card-network/brain-service \
      lcn-brain-service bun -e '
    import {
      SQSClient,
      CreateQueueCommand,
      GetQueueAttributesCommand
    } from "@aws-sdk/client-sqs";

    const client = new SQSClient({
      region: "us-east-1",
      endpoint: "http://elasticmq:9324",
      credentials: { accessKeyId: "local", secretAccessKey: "local" }
    });
    const dlqUrl = "http://elasticmq:9324/000000000000/inbox-dlq";

    await client.send(new CreateQueueCommand({ QueueName: "inbox-dlq" }));
    const dlq = await client.send(new GetQueueAttributesCommand({
      QueueUrl: dlqUrl,
      AttributeNames: ["QueueArn"]
    }));
    await client.send(new CreateQueueCommand({
      QueueName: "inbox",
      Attributes: {
        VisibilityTimeout: "1800",
        RedrivePolicy: JSON.stringify({
          deadLetterTargetArn: dlq.Attributes.QueueArn,
          maxReceiveCount: "5"
        })
      }
    }));
    client.destroy();
    console.log("Inbox queues ready");
    '
    ```

5. Leave the standalone inbox worker stopped. In Postman, make this request with a local token that includes `inbox:write` and `inbox:read`:

    ```text
    POST http://localhost:4000/api/inbox/issue-batch
    Authorization: Bearer <LOCAL_API_TOKEN>
    Content-Type: application/json
    ```

    ```json
    {
        "requestId": "manual-batch-001",
        "configuration": {
            "delivery": {
                "suppress": true
            }
        },
        "items": [
            {
                "recipient": {
                    "type": "email",
                    "value": "queue-test-1@example.com"
                },
                "templateUri": "lc:network:localhost%3A4000/trpc:boost:5df74ef8-3942-4d4d-b73f-6b604817c034",
                "idempotencyKey": "manual-batch-001-item-1"
            },
            {
                "recipient": {
                    "type": "email",
                    "value": "queue-test-2@example.com"
                },
                "templateUri": "lc:network:localhost%3A4000/trpc:boost:5df74ef8-3942-4d4d-b73f-6b604817c034",
                "idempotencyKey": "manual-batch-001-item-2"
            }
        ]
    }
    ```

    Use a new `requestId` and new item idempotency keys for each fresh run. Replace the local Boost URI if the fixture is unavailable in the tester's database.

6. Verify the response is HTTP `202` with `batchId`, `status: "QUEUED"`, and `createdAt`.

7. Poll the returned batch ID with the same token:

    ```text
    GET http://localhost:4000/api/inbox/batches/{batchId}
    ```

    With the worker stopped, verify the batch remains `QUEUED`, both items have state `QUEUED`, and `summary.pending` is `2`.

8. Start the standalone worker inside the Brain container:

    ```bash
    docker exec -it \
      -w /app/services/learn-card-network/brain-service \
      lcn-brain-service bun run inbox:worker
    ```

9. Poll the batch again. Verify `status` is `COMPLETED`, `summary.succeeded` is `2`, `summary.pending` is `0`, and both ordered item results contain `success: true`, an `issuanceId`, and a `claimUrl`. An item result of `status: "PENDING"` is expected for an unknown recipient and means the successfully issued credential is waiting to be claimed.

10. Resubmit the identical POST body and verify it returns the original `batchId`. Then change one recipient while keeping the same `requestId` and verify the API returns `409`.

11. Optionally stop the worker with `Ctrl+C`, submit another fresh batch, and restart the worker to verify queued work survives a worker restart without duplicate issuance.

Manual verification completed locally against the full-stack Docker environment: a two-item batch remained queued while the worker was stopped, then completed with two successful ordered issuance results after the worker started. The final summary reported `succeeded: 2`, `failed: 0`, `pending: 0`, and `unconfirmed: 0`.

Backend/API only. No device or browser-specific coverage is required.

#### 🧪 Code Coverage

- Unit tests cover batch orchestration, idempotency, error isolation, concurrency, payload enforcement, HTTP transport behavior, and SDK forwarding/polling.
- Container-backed integration tests exercise isolated Neo4j, real Redis, and ElasticMQ/SQS.
- Integration coverage includes asynchronous acceptance, polling authorization, atomic request replay and quota admission, duplicate delivery, worker recovery, durable outbox retries, dead-letter handling, cleanup, and reconciliation states.
- Local validation passes:
    - Brain Service typecheck.
    - `@learncard/types` typecheck.
    - Network plugin build.
    - Focused batch/checkpoint unit tests: 31 passing.
    - Network plugin tests: 46 passing, 6 skipped.
    - Docker-backed inbox e2e suite: 3 test files and 71 tests passing.
    - Prettier, workspace ESLint, docs links, and docs snippet validation.

# Documentation

- [ ] **Tutorial** — New capability that users need to learn (`docs/tutorials/`)
- [ ] **How-To Guide** — New workflow or integration (`docs/how-to-guides/`)
- [x] **Reference** — New/changed API, config, or SDK method (`docs/sdks/`)
- [x] **Concept** — New mental model or architecture explanation (`docs/core-concepts/`)
- [ ] **App Flows** — Changes to LearnCard App or ScoutPass user flows (`docs/apps/`)

**Internal/AI Docs**

- [ ] **AGENTS.md** — New pattern, flow, or context that AI assistants need
- [x] **Code comments/JSDoc** — Complex logic that needs inline explanation

**Visual Documentation**

- [ ] **Mermaid diagram** — Complex flow, state machine, or architecture

#### 💭 Documentation Notes

This PR documents the new batch contract, polling example, retry and recovery semantics, limits, local setup, operational behavior, and retention policy in the Universal Inbox concept guide and API reference. It includes a changeset for the new SDK and API capability.

# ✅ PR C

### ✨ PR Description
Purpose: Implement a durable batch issuance system for the Universal Inbox to support queuing and processing of up to 10 credentials per request.

Main changes:
- Added `issueBatch` and `getBatch` endpoints with 4MiB payload limits and HTTP 202 responses.
- Implemented SQS-backed background workers with Neo4j-based job tracking, leases, and idempotency keys.
- Added `guardianEmail` support to single and batch issuance with case-insensitive self-approval validation.

---
Rovo Dev code review: <strong>Rovo Dev couldn't review this pull request</strong>
The pull request author does not have access to AI for the associated Atlassian organization.


gerardopar:
@claude review PR please

Custard7:
Reviewed the diff, checked out the branch, built types, ran the four new test files (all pass), and typechecked brain-service. Solid implementation — atomic reservation via `SET NX` + CAS Lua, quota and expiry in one script, order-preserving worker pool, and body-size guards on both Fastify and Lambda. A few items before merge, then some follow-ups.

**Before merge**

1. **`guardianEmail` is now accepted by the existing `/inbox/issue` route.** It was added to the shared `IssueInboxCredentialValidator` (`lcn.ts:1307`), so `routes/inbox.ts:469` now forwards it into `issueToInbox` and single-issue callers can trigger the guardian-approval flow. The description says that route is unchanged — either document this (and add a `.describe()` like the sibling fields) or move it to `InboxBatchConfigurationValidator`. There's currently no test for it either way.

2. **No test for duplicate idempotency keys within a single batch.** Realistic client mistake; outcome currently depends on worker timing.

3. **All tests run on ioredis-mock, including `test/inbox-batch.spec.ts`.** The three Lua scripts that provide the atomicity guarantees have never executed against real Redis in CI. `vitest.inbox-e2e.config.ts` already has a Redis container — adding the batch spec (or at least `inbox-batch-cache.test.ts`) to it would close this cheaply.

**Worth a look / follow-up**

4. `inbox-batch.helpers.ts:1347-1359` — if `compareAndSet` fails *after* `issueToInbox` succeeded, the `issuanceId`/`claimUrl` are discarded and the item returns `CONFLICT` with "contact support." Returning the success payload (or at least `issuanceId`) would let integrators reconcile without a ticket. Only the `undefined` variant is tested; `false` (owner lost) isn't.

5. `inbox-batch.helpers.ts:1315-1319` — if the holder releases its reservation between `setIfAbsent → null` and the follow-up `get`, the item surfaces as `INTERNAL_SERVER_ERROR` (non-retryable) rather than `CONFLICT`. Untested.

6. An over-quota batch still increments the counter, so a partner at 9,950/10,000 submitting 100 items gets rejected *and* loses the remaining 50 for the window. The integration test locks this in, but the docs only mention "replays and failed attempts." Either check-then-increment or document it.

7. New SDK method `sendCredentialsViaInbox` has no test, and `tests/e2e` already exports a same-named helper with different semantics. Suggest `sendCredentialBatchViaInbox`.

**Nits**

- `docs/sdks/learncard-network/universal-inbox-api.md` wasn't updated with `/inbox/issue-batch`.
- `replayResult` has a "records without fingerprint" compat branch — new key namespace, so no such records exist.
- Prod fail-closed guard (`NODE_ENV==='production' || AWS_LAMBDA_FUNCTION_NAME` with no Redis) is untested, and `serverless-offline` also sets that env var.

gerardopar:
@claude review please

gerardopar:
@claude pushed up the recommended changes, can you review again please

gerardopar:
@claude okay ive addressed the latest recommendations, can you review latest changes please

gerardopar:
@claude review latest changes please

services/learn-card-network/brain-service/src/helpers/inbox-queue.helpers.ts
```diff
@@ -0,0 +1,336 @@
...
+                item.idempotencyKey === undefined
+                    ? `internal:${randomUUID()}`
+                    : `client:${fingerprint([profile.profileId, item.idempotencyKey])}`,
+        };
+    });
+    const configuredLimit = Number(getInboxBatchRuntimeEnvironment().INBOX_BATCH_ITEMS_PER_HOUR);
+    return createBatchJob({
+        issuer: profile.profileId,
+        requestId: batch.requestId,
+        requestHash: fingerprint({ batch, domain: ctx.domain, tenant: ctx.tenant }),
```
Custard7: **`requestId` replay can 409 on a legitimate retry.** `ctx.tenant` is a `ResolvedTenant`, which includes `resolvedVia: 'header' | 'origin' | 'env' | 'default'` and the full `emailBranding` object. Two retries for the *same* tenant will hash differently if one request carries `X-Tenant-Id` and the other resolves via `Origin`/env default, or if `TENANT_EMAIL_BRANDING` changes between deploys within the 24h window. The client sees `CONFLICT: requestId was used for a different batch` for a payload that is byte-identical from its perspective.

Severity is low–medium (only when the resolution path differs between attempts), but the fix is a one-liner: hash `ctx.tenant.id` instead of the whole object.

services/learn-card-network/brain-service/src/helpers/inbox-queue.helpers.ts
```diff
@@ -0,0 +1,336 @@
...
+    const succeeded = entries.filter(i => i.result?.success).length;
+    return {
+        batchId,
+        createdAt: new Date(Number(job.createdAt)).toISOString(),
+        status: getInboxBatchState(entries.map(item => item.state)),
+        items: entries,
+        summary: {
+            total: entries.length,
+            succeeded,
+            failed: entries.filter(i => i.result && !i.result.success).length,
```
Custard7: **`summary.failed` double-counts `unconfirmed` items.** Items in `NEEDS_RECONCILIATION` carry a `success: false` result (persisted by `finishBatchItem` or synthesized at read time), so they land in both `failed` and `unconfirmed`. The docs present these as separate summary fields with no mention of overlap, so a consumer asserting `succeeded + failed + unconfirmed === total` gets the wrong answer.

Either exclude `NEEDS_RECONCILIATION` items from `failed` here, or document the overlap explicitly in `universal-inbox-api.md`.

services/learn-card-network/brain-service/src/accesslayer/inbox-batch/store.ts
```diff
@@ -0,0 +1,458 @@
...
+        WITH i WHERE i.state = 'QUEUED' AND i.dispatchAt < $now
+        SET i.dispatchAt = $next RETURN i.id AS id`,
+            { now: Date.now(), next: Date.now() + 60_000 }
+        );
+        return result.records.map(r => r.get('id'));
+    });
+
+export const acknowledgeInboxDispatches = async (ids: string[]): Promise<void> => {
+    await neogma.queryRunner.run(
+        `MATCH (i:InboxBatchItem) WHERE i.id IN $ids AND i.state = 'QUEUED'
```
Custard7: **Dispatch-ack race can push a worker retry out by 30 minutes.** Sequence: `takeInboxDispatches` → `SendMessageBatch` → worker receives, runs, and `finishBatchItem` sets `state='QUEUED', dispatchAt=0` for a transient retry → *then* this ack runs. The `i.state = 'QUEUED'` guard matches the freshly re-queued item and sets `dispatchAt = now + 30min`, so the retry waits 30 min instead of ~1 min.

This needs the worker to finish before the dispatcher's ack write lands — plausible under load (Lambda cold start on the dispatcher side, slow Neo4j write). Not a correctness failure, but a silent latency cliff. Guarding on `i.dispatchAt = $next` (the value `take` just wrote) instead of `state` would make the ack only touch rows it actually leased.

services/learn-card-network/brain-service/src/accesslayer/inbox-batch/store.ts
```diff
@@ -0,0 +1,458 @@
...
+    owner: string,
+    result: IssueInboxCredentialBatchItemResult
+): Promise<void> => {
+    const encrypted = await encryptInboxCredential(JSON.stringify(result));
+    await transaction(async tx => {
+        const item = await lockOwnedItem(tx, id, owner);
+        const unconfirmed = !result.success && item.phase === 'ISSUING';
+        const retry =
+            !result.success &&
+            result.error.code === 'INTERNAL_SERVER_ERROR' &&
```
Custard7: **Item-level "in progress" conflict is terminal rather than retryable.** If two batches submit the same `idempotencyKey` and the first is still processing, `replayResult` throws `CONFLICT` for the second (`inbox-batch.helpers.ts:71`). Since only `INTERNAL_SERVER_ERROR` is requeued here, the second item is permanently `COMPLETED/failed` even though the first will succeed seconds later.

The docs describe this as intended, but it turns the most natural client retry pattern — resubmit the same items with the same keys under a fresh `requestId` — into permanent failures. Worth considering requeue-with-backoff for this specific case (the replay record's `state === 'processing'` is easy to detect).

services/learn-card-network/brain-service/src/helpers/inbox-batch.helpers.ts
```diff
@@ -0,0 +1,326 @@
...
+                    item.idempotencyKey !== undefined &&
+                    firstIndexByKey.get(item.idempotencyKey) !== index
+                ) {
+                    throw new TRPCError({
+                        code: 'CONFLICT',
+                        message:
+                            'Duplicate idempotency key in this batch. Only its first occurrence is attempted.',
+                    });
+                }
+                const configuration = mergeWith(
```
Custard7, line 152: Minor: with `mergeWith` an item cannot *unset* a batch-level default (e.g., remove `guardianEmail` for one recipient). lodash skips `undefined`, and `null` fails `.email()`. Either document that batch defaults are additive-only or explicitly support `null` as "clear this key".

packages/learn-card-types/src/lcn.ts
```diff
@@ -1425,6 +1442,104 @@ export const IssueInboxCredentialResponseValidator = z.object({
...
+    .unwrap()
+    .extend({
+        delivery: IssueInboxCredentialValidator.shape.configuration
+            .unwrap()
+            .shape.delivery.unwrap()
+            .extend({ suppress: z.boolean().optional() })
+            .optional(),
+    });
+
+export const IssueInboxCredentialBatchItemValidator = z
```
Custard7, line 1464: Minor: spreading `.shape` drops both `.refine()`s from the single-issue validator. The `credential || templateUri` check has no dependency on config merging and could run at submission (see DX audit #1 in the summary) rather than surfacing as a per-item `BAD_REQUEST` a minute later.

packages/plugins/learn-card-network/src/types.ts
```diff
@@ -679,6 +682,9 @@ export type LearnCardNetworkPluginMethods = {
     revokeAuthGrant: (id: string) => Promise<boolean>;
     getAPITokenForAuthGrant: (id: string) => Promise<string>;
 
+    /** Queue up to 100 credentials. Receipt status is a snapshot; poll for ordered results. */
+    sendCredentialsViaInbox: (batch: IssueInboxCredentialBatch) => Promise<InboxBatchReceipt>;
+    getInboxCredentialBatch: (batchId: string) => Promise<InboxBatchStatus>;
```
Custard7, line 692: Minor: `sendCredentialsViaInbox` has JSDoc; this one doesn't. Worth a short block describing the returned `status`/`summary` semantics and linking to the polling guidance.

services/learn-card-network/brain-service/lambda.ts
```diff
@@ -25,6 +25,8 @@ import {
 import { environment } from './src/config/environment';
 import { toServerlessApplication } from './src/helpers/serverlessApplication';
 import { runInboxMaintenance } from './src/helpers/inbox-maintenance.helpers';
+import { inboxBatchResponseMeta } from './src/helpers/inbox-batch-http.helpers';
+import { withInboxBatchBodyLimit } from './src/helpers/inbox-batch-http.helpers';
```
Custard7: Nit: duplicate import from the same module on consecutive lines — can be merged.

Custard7 (commented):
## Review

The durability design (Neo4j admission → outbox → lease/ownership → replay CAS) is sound and the test suite is thorough. No blocking bugs found. Inline comments cover four correctness items (the first two are small and worth fixing before merge) plus a few minor notes. Below is a DX audit of the batch API as a whole.

---

## DX audit — batch issuance

The contract is coherent, the docs are honest about at-least-once semantics, and `requestId` / `idempotencyKey` are the right primitives. These are the gaps between "correct" and "top notch", roughly in leverage order. Items 1, 2, 3 and 5 are additive if done now; #2 becomes breaking once clients ship.

**1. Fail fast at submission for config-independent errors.** An item with neither `credential` nor `templateUri`, or with `recipient.value === guardianEmail`, gets a `202` and only surfaces as a per-item `BAD_REQUEST` after the ~60s dispatcher cycle. A developer testing their integration will submit, poll for a minute, then learn they forgot a field. Validate everything that doesn't depend on merging at submit time and return `400` with `{ index, message }` per offending item.

**2. `status` doesn't tell you whether the batch is done.** `getInboxBatchState` returns `NEEDS_RECONCILIATION` as soon as one item enters that state, even with 90 items still `QUEUED`. The docs then instruct developers to ignore `status` and check `summary.pending === 0`. A top-level status that requires a footnote is a footgun. Suggest making `status` a pure lifecycle field (`QUEUED | PROCESSING | COMPLETED`) and surfacing reconciliation via `summary.unconfirmed > 0` / `hasUnconfirmed: boolean` — or just add `done: boolean`. A dev should be able to write `if (batch.done)` without reading the guide.

**3. Ship a polling helper in the SDK.** Every consumer will write the same loop from the docs. Provide it once, correctly:
```ts
const batch = await learnCard.invoke.waitForInboxCredentialBatch(receipt.batchId, {
    timeoutMs: 10 * 60_000,
    intervalMs: 2_000, // with backoff
    signal,            // AbortSignal
    onProgress: status => {},
});
```
Bonus: a one-shot `sendCredentialsViaInboxAndWait(batch, opts)` for scripts. Keep raw `sendCredentialsViaInbox` for callers managing their own polling.

**4. `sendCredentialViaInbox` vs `sendCredentialsViaInbox`.** One character apart, different return shapes (result vs receipt), different semantics (sync vs async). Autocomplete surfaces both; a typo compiles. Consider `sendCredentialBatchViaInbox` / `issueInboxBatch` so the plural intent is unmistakable and mirrors the REST path (`issue-batch`).

**5. Echo `idempotencyKey` (and ideally `recipient`) on every item result.** Failure results contain only `index` + `error`. Developers reconcile against their own records by key, not array position — especially after chunking 5,000 recipients into 50 batches. Adding `idempotencyKey?: string` to both branches of `IssueInboxCredentialBatchItemResultValidator` is cheap and removes a class of off-by-one bugs on the client.

**6. Small-batch latency.** The schedule-driven dispatcher means a 1-item batch waits up to ~60s before a worker starts. If a developer's first experience is "I called the API and nothing happened for a minute," that's a trust hit. Options: best-effort inline `SendMessage` after the Neo4j commit (outbox already covers the failure case), or clearly steer `<N` items to `sendCredentialViaInbox` in docs + JSDoc.

**7. No batch-completion webhook.** Existing webhooks fire per credential. An `INBOX_BATCH_COMPLETED` webhook with `batchId` + summary would let integrators skip polling entirely for long runs.

**8. No `GET /inbox/batches` list endpoint.** If a client loses a `batchId` (crash between submit and persist), the only recovery is replaying the same `requestId` within 24h — which per the `requestHash` comment can spuriously 409. A paginated list by issuer with a `createdAt` cursor closes the loop.

**9. Docs.** Good: honest about at-least-once, reconciliation, quota; REST + SDK + tRPC all covered. Missing:
- An error-code table for item results. `CONFLICT` currently has four distinct meanings (in-batch duplicate key, changed input under same key, in-progress, unconfirmed-after-issue). Developers branching on `error.code` can't distinguish "retry same key later" from "this will never succeed." Either split codes (`DUPLICATE_KEY`, `IDEMPOTENCY_MISMATCH`, `IN_PROGRESS`, `UNCONFIRMED`) or add a machine-readable `reason` field.
- Chunking guidance for 5,000+ recipients (how many in-flight batches, quota math) — this is the use case that motivated the feature.
- One sentence on `summary.completed` vs `succeeded`: "`completed` counts terminal items regardless of outcome."

**10. Local dev.** `compose.inbox.yml` + `inbox:worker` + three env vars is workable. The manual `bun -e` queue-creation snippet in the PR description should be a script (`bun run inbox:setup-queues`) rather than living in a PR body. "No inline fallback when queue config is missing" is the right call for prod, but an `INBOX_QUEUE_INLINE=true` dev mode that processes synchronously would remove the worker-process requirement for the 90% of local work that isn't testing queue behavior.

Custard7:
https://github.com/learningeconomy/LearnCard/pull/1585#issuecomment-5722052552

@goblincore @gerardopar — heads up that this PR and #1585 (`refresh` in universal inbox) touch the same inbox configuration surface. Please coordinate so this refresh option is available through the new bulk endpoint:

Since #1585 merged first:
- this PR should pick up the refresh options in the batch route before merging.

The goal is that `/inbox/issue-batch` supports refresh the same way `/inbox/issue` does, with a test covering it.

gerardopar:
@Custard7 @goblincore added refresh support here

https://github.com/learningeconomy/LearnCard/pull/1591/changes/229f20f6e3081ccd6c45c9818a3c8bc10cb603aa

& here

https://github.com/learningeconomy/LearnCard/pull/1591/changes/5d5fbec386da42807fb16ab7e39d9b9eeeb28fdf

--

@Custard7 if you want to review again 🙏🏽

Custard7 (approved):
Approving — the core is solid (quota admission is properly serialized via the `InboxBatchIssuer` node lock, `/inbox/issue` behavior is preserved, lease/timeout sizing is consistent, `getBatch` is issuer-scoped) and prior review items are addressed.

Since a partner will be running this nightly, a few things worth tightening. The first two I'd love to see before they integrate; the rest can follow.

### 1. Check refresh scope at submit, not just in the worker
`issueBatch` only requires `inbox:write`. `assertInboxRefreshEnabled` (which checks `CREDENTIAL_REFRESH_ENABLED` + `credentials:write`) runs only in the worker (`inbox-batch.helpers.ts:226`). So a token missing `credentials:write` — or a deployment with refresh disabled — gets a **202**, is charged quota, and then every refreshable item fails asynchronously.

The scope is already captured at admission (`inbox-queue.helpers.ts:102`), so this is small: in `submitInboxBatch`, if any item resolves to `refresh: true` (`item.configuration?.refresh ?? item.refresh ?? batch.configuration?.refresh`), call `assertInboxRefreshEnabled(ctx.user.scope)` before `createBatchJob`. Keep the worker check too — scope can change between submit and processing.

### 2. Surface a `retryable` signal on item errors
`SaIssueError` already carries `retryable` and `kind` (`signingAuthority.helpers.ts:39`), but it isn't a `TRPCError`, so the catch in `issueInboxBatch` (line 338) collapses it to `INTERNAL_SERVER_ERROR / "Failed to issue credential"`. A permanently broken SA endpoint (4xx, `retryable: false`) and a transient 5xx look identical to the partner.

The docs make this sharper: the error table (`universal-inbox-api.md:156`) says `INTERNAL_SERVER_ERROR` → "retry the same key", so a partner following the docs would retry a misconfigured SA forever. Item failures are recorded as terminal results (not rethrown), so SQS redelivery doesn't cover this.

Suggestion: add optional `retryable: boolean` to the item error schema — from `SaIssueError.retryable` when present, `true` for the generic fallback, `false` for 4xx codes. One `instanceof` branch, additive to the Zod type.

### 3. Replay TTL and post-expiry dedupe
`InboxBatchReplay` records expire at 24h and are swept (`store.ts:454`). After that, a resubmit with the same key issues a fresh credential — nothing downstream dedupes it. Notably the batch path strips `idempotencyKey` before `issueToInbox` (`inbox-batch.helpers.ts:170, 238–243`), so refresh items get a `randomUUID()` `issueKey` (`inbox.helpers.ts:496`) and lose the durable `getInboxRefreshReplay` dedupe the single route has.

Two options, either works for me: bump `IDEMPOTENCY_TTL_SECONDS` (and the `requestId` TTL) to ≥72h so a delayed nightly job over a long weekend is safe, and/or forward the batch `idempotencyKey` into `issueToInbox` for refresh items so they get the durable dedupe.

### 4. Integration coverage for the refresh loop
`test/inbox-batch.spec.ts` has no refresh coverage; refresh is only unit-tested with `assertInboxRefreshEnabled` mocked. One container-backed test — batch with `refresh: true` → item `PENDING` → `publishCredentialRefresh` v2 → claim → assert v2 lands — would cover the exact path the partner runs nightly. The pieces exist in `inbox-refresh.spec.ts` and `credential-refresh-publication.spec.ts`.

### 5. Doc nits
- The "dispatch can take up to about a minute" note (`universal-inbox-api.md:167`, `universal-inbox.md:161`) would be more discoverable next to the polling examples (lines 111 and 79 respectively).
- `delivery.suppress` per item is honored (item config merges over batch config; the batch validator uses `.optional()` so an omitted item value doesn't clobber the batch default). Worth stating explicitly in the batch docs.

### Minor
- `inbox-queue.helpers.ts:20–30`: two imports from `@accesslayer/inbox-batch/store` on adjacent lines.
- `inbox-queue.helpers.ts:244`: `queueAgeMs` is computed from `job.createdAt`, so it's batch age rather than queue age.
- `waitForInboxCredentialBatch` aborts on a single transient read failure and the thrown error doesn't include `batchId`; retrying reads within the deadline and attaching `batchId` would help partners who didn't wire `onSubmitted`.

gerardopar:
@Custard7 pushed the final recommended items, going to merge 🚀

thanks for reviewing!

PR #1598, opened 2026-09-18
feat: act as a managed profile — X-LearnCard-Act-As, grant actAs policy, SDK actAs(), CLI spec

Custard7:
> **Stacked on #1590** (`feat/cli-issuer-org-tooling`). Review that first; this diff is only the delta.

## What

A backend holding one API token can now issue **as a profile it manages** — a state agency's token acting as a district — without holding the seed and without a token per district.

The mechanism is one request header, `X-LearnCard-Act-As: <profileId>`. The network swaps the acting profile for that request when three things hold: the caller is authenticated, the caller's profile manages the target, and — for API tokens — the token was explicitly allowed to delegate. Scope is unchanged while acting. Every send/inbox activity records `onBehalfOf`.

**Server** (`brain-service`, `@learncard/types`)
- `didRoute` resolves the header once the challenge is verified (unverified callers do no extra work). Relationship check reuses `getProfilesThatManageAProfile`; nothing new in Cypher.
- Auth grants gain `actAs`: `'*'` = any managed profile, a comma-separated list = only those, **absent = no delegation**. Immutable on token update, like `scope`.
- `onBehalfOf` on send / inbox-issue activity records where the actor is `ctx.user` (not on `claimBoostWithLink`, where the actor is the SA owner and `ctx.user` is the claimer). `ACT_AS_HEADER` exported from `@learncard/types`.

**SDK** (`@learncard/network-plugin`, `@learncard/init`)
- `initLearnCard({ apiKey | seed, network, actAs })` sets the header for the whole instance.
- `learnCard.invoke.actAs(profileId)` returns a new scoped instance; the original is untouched. It probes `getProfile` eagerly and rethrows `FORBIDDEN` / `NOT_FOUND`, so a missing `actAs` grant or bad profile id fails at `actAs()` rather than surfacing later as "Please make an account first!".

**CLI** (`@learncard/cli`)
- `serviceAccounts[].actAs: '*' | [profileIds]` in the org spec, validated against `profileManager.managed`, set on the grant at creation. Because `actAs` is immutable on the grant (like `scope`), `org apply` errors on drift and tells you to `token --revoke <id>` and re-apply (new token).
- `doctor` token-scopes and `whoami` say what each token may act as. New `examples/delegated-service-account.network.yaml`.

**Docs / tests**
- How-to gains "Let a service account act as a district" with a runnable snippet; reference table documents `actAs`.
- `tests/e2e/tests/act-as.spec.ts`: seed and token paths, plus every deny (no relationship, no `actAs`, wrong list entry, unknown profile, scope not widened). `docs-org.spec.ts` runs the new snippet.

## Why

Districts are managed profiles with no seed of their own. Until now, "issue as a district" meant handing a vendor the state's seed, or minting and rotating one token per district. This is the same shape as the CLI's `--as` from #1590 — the actor is named on the call, never hidden in state — and deny-by-default on the grant is what makes a leaked single-district token unable to issue as the whole state.

## Notes for reviewers

- **Read `resolveActAs` in `routes/index.ts` first.** It is the entire authorization change (~45 lines). The three checks are in order: identifier shape → target exists → caller manages target → (token only) grant permits. Scope is never touched.
- Placement: the swap runs in `didRoute` (not `didAndChallengeRoute`) because `profiles.getProfile` is a `didRoute` procedure — if the swap ran later, `getProfile()` would report the manager while `send()` acted as the district. It is gated on `isChallengeValid`, so unverified Bearers trigger no lookups (tested).
- `ActingLearnCard` uses an open plugin tuple because a plugin cannot name its host's plugin list from inside its own method map. It's the honest type, not an `any`.
- The `inbox` public preview route is `openRoute` and intentionally does not act-as.
- Follow-ups noted by the server pass, deliberately not in this PR: `onBehalfOf` on acceptance/finalization events; activity logging on contract/federation sends (they have none today either way).
- e2e: `act-as.spec` (9) and `docs-org.spec` (2) are **11/11 green** against a local stack built from this branch. Note the harness clears every database after each test, so both specs rebuild their org in `beforeEach`.

---

### Setup

```sh
git fetch && git checkout feat/act-as-managed-profile
bunx nx run-many -t build -p types network-plugin init cli
cd packages/learn-card-cli && npm link && cd -
```

### 1. Unit suites

```sh
(cd services/learn-card-network/brain-service && bun run test | tail -3)
(cd packages/plugins/learn-card-network      && bun run test | tail -3)
(cd packages/learn-card-init                 && bun run test | tail -3)
(cd packages/learn-card-cli                  && bun run test | tail -3)
```

- [ ] brain 212 · network-plugin 54 · init 55 (+3 type tests in `test/act-as.types.test.ts`) · cli 170 — all passing

### 2. Read the authorization code

```sh
git diff feat/cli-issuer-org-tooling...HEAD -- services/learn-card-network/brain-service/src/routes/index.ts
```

- [ ] `resolveActAs`: BAD_REQUEST → NOT_FOUND → FORBIDDEN (relationship) → FORBIDDEN (token policy), in that order
- [ ] `scope` is copied through unchanged; `did` and `profile` are replaced; `onBehalfOf` is set
- [ ] Only called when `ctx.user.isChallengeValid`
- [ ] `routes/act-as.test.ts` has a case for each of the four failures plus the blank-header and unverified-caller no-ops

### 3. Spec + CLI (offline, then staging)

```sh
cd packages/learn-card-cli
printf 'issuer:\n  profileId: x\n  displayName: X\n  signingAuthority: { type: learncard-hosted, name: x }\nserviceAccounts:\n  - { name: a, scopes: [inbox:write], actAs: [nope] }\n' > /tmp/bad-actas.yaml
learncard org apply /tmp/bad-actas.yaml --dry-run --network staging
```

- [ ] Error: `serviceAccounts.0.actAs: "nope" is not a managed profile in this spec`

Then in the staging review folder from #1590 (`/tmp/lc-review`, with `$ISSUER` / `$DISTRICT` set):

```sh
cd /tmp/lc-review
# edit org.yaml: on the serviceAccounts entry add   actAs: ['$DISTRICT']   (literal id, not the var)
learncard org apply org.yaml --network staging --dry-run
```

- [ ] dry-run **errors**: `actAs` on an existing service account cannot be changed in place — message tells you to `learncard token --revoke <id>` and re-apply

```sh
learncard token --revoke <id-from-the-error>
learncard org apply org.yaml --network staging --secrets-out ./secrets.env -y
learncard org apply org.yaml --network staging --dry-run
learncard whoami | sed -n '/Service accounts/,$p'
learncard doctor | grep -i "token scopes"
```

- [ ] apply: `serviceAccount <name> created` (new token written to `secrets.env`) → re-run dry-run: `No changes.`
- [ ] `whoami` lists the service account with `acts as: $DISTRICT`; `doctor` detail ends with `may act as: $DISTRICT`

### 4. Act as the district with the token (staging)

```sh
source ./secrets.env                     # exports e.g. REVIEW_..._ISSUER=<token>
TOKEN=$(grep -v '^#' secrets.env | head -1 | cut -d= -f2-)
cat > /tmp/act.mjs <<'EOF'
import { initLearnCard } from '@learncard/init';
const [token, district] = process.argv.slice(2);
const org = await initLearnCard({ apiKey: token, network: 'https://staging.network.learncard.com/trpc' });
console.log('token is', (await org.invoke.getProfile()).profileId);
const d = await org.invoke.actAs(district);
console.log('acting as', (await d.invoke.getProfile()).profileId);
console.log('original still', (await org.invoke.getProfile()).profileId);
try { await d.invoke.addAuthGrant({ name: 'escalate', scope: '*:*' }); console.log('ESCALATED — BAD'); }
catch (e) { console.log('escalation blocked:', e.data?.code); }
EOF
cd packages/learn-card-cli && node /tmp/act.mjs "$TOKEN" "$DISTRICT"
node /tmp/act.mjs "$TOKEN" "some-profile-you-do-not-manage"
```

- [ ] `token is $ISSUER` → `acting as $DISTRICT` → `original still $ISSUER` → `escalation blocked: FORBIDDEN`
- [ ] Second run: `actAs()` itself throws `FORBIDDEN` (or `NOT_FOUND` if the id doesn't exist) — never a profile swap
- [ ] Remove `actAs` from the spec, `org apply -y` → errors (revoke + re-apply required); revoke, re-apply, re-run the first command with the new token → `actAs()` rejects with `FORBIDDEN` / `Grant actAs on the token`

### 5. e2e (needs the local docker stack)

```sh
cd tests/e2e && E2E_MANAGE_DOCKER=false bunx vitest run tests/act-as.spec.ts tests/docs-org.spec.ts
```

- [ ] 11 tests pass (`act-as` 9, `docs-org` 2)

### 6. Docs

```sh
node scripts/check-docs-links.mjs && node scripts/check-docs-snippets.mjs
```

- [ ] Both OK; read the new "Let a service account act as a district" section — it should match what you just did in step 4

---


packages/plugins/learn-card-network/src/plugin.ts
```diff
@@ -519,13 +520,29 @@ export * from './types';
 
 export type GuardianApprovalGetter = () => string | undefined | Promise<string | undefined>;
 
+export { ACT_AS_HEADER };
+
+const isActAsDenial = (error: unknown): boolean => {
+    const code = (error as { data?: { code?: string } } | undefined)?.data?.code;
+    return code === 'FORBIDDEN' || code === 'NOT_FOUND';
```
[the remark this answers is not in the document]
Custard7: Fixed in bb4b8c31d — `isActAsDenial` now treats FORBIDDEN/BAD_REQUEST as denials and NOT_FOUND only when it carries the explicit message `resolveActAs` now throws (`Act-as target profile not found`). Denials are rethrown with the original as `cause`; anything else surfaces unchanged.

Custard7:
Review follow-ups landed in bb4b8c31d (stack rebased onto the updated #1590):

- `isActAsDenial` narrowed — replied inline.
- Refresh-path `logCredentialDelivered` in `inbox.helpers.ts` now records `onBehalfOf`, matching the first delivery path.
- Performance note: the two lookups per act-as request are a considered tradeoff for backend issuance volume; left uncached with a comment at the call site.
- Because #1590 now treats scope/expiry drift as an error, `actAs` is folded into the same drift check (error + revoke hint; dry-run `drifted`). Changeset and the how-to troubleshooting row updated to describe the revoke → re-apply flow.

network-plugin act-as 14/14, brain-service act-as 20/20, CLI 267/267.

packages/learn-card-cli/src/org/apply.ts
```diff
@@ -470,14 +476,18 @@ const applyServiceAccounts = async (
...
                 created: false,
             });
+            // scope, expiresAt and actAs are all fixed when the token is minted — the only
+            // way to change them is to revoke and mint a replacement.
             const drift = [
                 normalizeScope(existing.scope) !== normalizeScope(account.scopes.join(' ')) &&
                     'scope',
                 expiryInstant(existing.expiresAt) !== expiryInstant(account.expiresAt) &&
                     'expiresAt',
+                getGrantActAs(existing) !== actAs &&
```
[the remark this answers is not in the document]
Custard7: Fixed in b32cf6ce2 — `getGrantActAs` now returns `grant.actAs || undefined`, so a `null` or `''` from the wire compares equal to an absent spec value. The server type is `z.string().optional()` so this shouldn't occur in practice, but the coercion is cheap and matches how `describeActAs` already treats falsy values. Tests added.

services/learn-card-network/brain-service/src/routes/auth-grants.ts
```diff
@@ -142,6 +142,7 @@ export const authGrantsRouter = t.router({
                 updates: AuthGrantValidator.partial().omit({
                     id: true,
                     scope: true,
+                    actAs: true,
```
TaylorBeeston: I overstated this: it's non-blocking API feedback, not a permission bypass. As the grant owner, submit `updates: { name: 'renamed', actAs: '*' }` for a grant limited to `district-a`. The name changes and the request returns true, but `actAs` stays `district-a`; an actAs-only request makes no update. The schema strips it before the rejection check. Rejecting the mixed request would be clearer, but immutability is enforced, matching the existing handling of `scope`.

— OMP

services/learn-card-network/brain-service/src/routes/boosts.ts
```diff
@@ -1739,6 +1740,7 @@ export const boostsRouter = t.router({
                         const activityId = await traceDb('logCredentialSent:inbox', () =>
                             logCredentialSent({
                                 actorProfileId: profile.profileId,
+                                onBehalfOf: ctx.user.onBehalfOf,
```
TaylorBeeston, line 1745: Please fix this before merge. With refresh enabled locally, register a signer and publish a template for a managed district. Authenticate with its manager's seed; call `actAs(districtId)`, then `send({type:'boost', recipient: holderProfileId, templateUri, refresh:true})` to another local profile. The activity identified by `activityId` records the district but omits the manager from `onBehalfOf`. Injecting a rejection from `sendRefreshableCredential` reproduces the same omission on `FAILED`. Both log calls need `ctx.user.onBehalfOf`.

— OMP

TaylorBeeston (changes requested):
After another pass, the refresh-send audit gap is the one I'd fix before merging: a delegated send should retain who actually made it. I've added the local reproduction inline. I overstated the immutable-field issue earlier; the policy stays unchanged, so that's non-blocking API feedback, not a permission bypass. The existing relationship, token-policy, and scope checks still apply. This is source-traced; I haven't run live refresh issuance.

— OMP

services/learn-card-network/brain-service/src/routes/auth-grants.ts
Custard7: Fixed in 542c419b2 — `updates` is now `AuthGrantValidator.partial().strict()` (immutable keys are no longer omitted), so the existing `invalidUpdates` guard actually sees `actAs` (and scope/status/etc.) and throws `BAD_REQUEST` with the explicit message; unknown keys are rejected by `strict`. Unit tests in `act-as.test.ts`; the integration spec was updated from `resolves.toBe(false)` to `rejects BAD_REQUEST` and now includes `actAs`.

services/learn-card-network/brain-service/src/routes/boosts.ts
Custard7: Fixed in 542c419b2 — the refreshable-send branch's `logCredentialSent` and `logCredentialFailed` now pass `onBehalfOf: ctx.user.onBehalfOf`, matching the other send branches. (That path has no separate `logCredentialDelivered`; the SENT relationship is written inside `sendRefreshableCredential`.)

docs/snippets/cli/org/send-as-managed-token.mjs
```diff
@@ -0,0 +1,47 @@
...
+// One token for the whole org. Each request names the profile it acts as; the network
+// enforces both the manager relationship and the token's actAs policy.
+const org = await initLearnCard({ apiKey: API_TOKEN, network: true });
+const district = await org.invoke.actAs(DISTRICT_PROFILE_ID);
+
+const me = await district.invoke.getProfile();
+console.log(`Acting as ${me.displayName} (${me.profileId})`);
+
+// A token has no signing key, so LearnCard signs for the district through the hosted
+// signing authority that `org apply` registered on it. Passing `template` (not a
```
TaylorBeeston, line 17: `org apply` only registers the parent's signer, not one on each managed district. I ran this snippet immediately after bootstrap: `actAs()` succeeded, but `send()` failed with ‘You must register a signing authority’. It worked after registering a district signer manually. Could the how-to include that setup, rather than claiming it's automatic?

— OMP

packages/learn-card-cli/src/org/apply.ts
```diff
@@ -595,14 +601,18 @@ const applyServiceAccounts = async (
...
                     created: false,
                 });
+                // scope, expiresAt and actAs are all fixed when the token is minted — the only
+                // way to change them is to revoke and mint a replacement.
                 const drift = [
                     normalizeScope(existing.scope) !== normalizeScope(account.scopes.join(' ')) &&
                         'scope',
                     expiryInstant(existing.expiresAt) !== expiryInstant(account.expiresAt) &&
                         'expiresAt',
+                    getGrantActAs(existing) !== actAs &&
```
TaylorBeeston: Not a merge blocker, but reordering `actAs: [district-a, district-b]` to `[district-b, district-a]` makes a live reapply report drift and require token revocation, even though permissions are identical. Could we compare normalized profile-ID sets here, as we do for scopes?

— OMP

TaylorBeeston (changes requested):
Live checks confirmed parent attribution on refresh delivery and failure, plus atomic rejection of immutable grant updates. I'd fix the documented token bootstrap before merge: it omits the district signer needed for its send to work. There's also a smaller policy-ordering issue inline.

— OMP

docs/snippets/cli/org/send-as-managed-token.mjs
Custard7: Fixed in b684e935f — rather than documenting the manual step, `org apply` now does it: with a `learncard-hosted` signer, each managed profile gets its own hosted signing authority created + registered (reported as `signingAuthority <profileId>/<name> created`, `unchanged` on reapply, and previewed in dry-run/fresh-org plans). It opens an LCA-capable did:web wallet per district (`connectAsDidWebSigner`). The how-to, CLI reference, README and changeset now say this explicitly and note that `self-hosted` specs still need a signer registered on each district. The e2e fixture comment was corrected (it builds the org by hand, so it keeps registering its own).

packages/learn-card-cli/src/org/apply.ts
Custard7: Fixed in b684e935f — `normalizeActAs` trims, de-duplicates and sorts the profile-ID list (keeping `'*'`), and the drift check compares the normalized forms, mirroring `normalizeScope`. Reordering `[district-a, district-b]` is no longer drift; tests cover reordered and whitespace-padded grants.

docs/how-to-guides/bootstrap-an-issuer-org.md
```diff
@@ -215,6 +217,79 @@ console.log(result.inbox?.claimUrl ?? result.credentialUri);
...
+### Let a service account act as a district
+
+Your backend probably holds a **token**, not the seed. A token is bound to the profile that minted it — so to issue as a district, the request says who it is acting as, and the network checks two things: that your issuer really manages that profile, and that the token was allowed to delegate.
+
+Allow it in the spec, per account:
+
+```yaml
+serviceAccounts:
+    - name: nightly-issuer
+      scopes: [inbox:write, inbox:read, credentials:write, credentials:read]
```
TaylorBeeston: Could you add `boosts:write` here and in `delegated-service-account.network.yaml`? A token minted with these exact scopes reaches `actAs()`, but the snippet’s send fails with `401: This operation requires boosts:write scope`. Adding only that scope made the same snippet issue successfully, and the holder verified and claimed it.

— OMP

packages/learn-card-cli/src/org/apply.ts
```diff
@@ -586,11 +663,23 @@ const applyProfileManager = async (
             name: managedSpec.profileId,
             action: 'created',
         });
+        await applyManagedSigner(
```
TaylorBeeston, line 672: Non-blocking: please include the hosted signer in the dry-run branch for a new district under an existing manager. I added one to an applied org: dry-run reported only `managedProfile would-create`; apply created both the profile and its signing authority. The fresh-org plan already includes both.

— OMP

TaylorBeeston (changes requested):
The hosted district signers and `actAs` set comparison work in the live stack. The documented token, however, lacks `boosts:write` and the snippet fails with HTTP 401. Adding that scope made the same snippet deliver a credential that the holder verified and claimed. Please correct the guide and example spec before merging. I also noted a non-blocking dry-run preview gap. All 289 CLI tests and docs checks passed.

— OMP

docs/how-to-guides/bootstrap-an-issuer-org.md
Custard7: Fixed in 6d9ea8c4a — added `boosts:write` to the how-to's `nightly-issuer` scopes and to both examples with `actAs` (`delegated-service-account.network.yaml`, `state-districts.network.yaml`), each with a one-line note that it's what `send({ template })` needs. Thanks for running it live.

packages/learn-card-cli/src/org/apply.ts
Custard7: Fixed in 6d9ea8c4a — the dry-run branch for a new district under an existing manager now also emits `signingAuthority <district>/<name> would-create` for hosted-signer specs, matching the fresh-org plan and what apply actually does. Test added.

TaylorBeeston (approved):
LGTM! The documented bootstrap and token snippet work end to end with the exact example scopes; the holder verified and claimed the credential. Incremental dry-run includes the district signer, and reapply preserves the grant/token. All 290 CLI tests and documentation checks passed.

— OMP

